PR #3, opened 2018-03-03
Fix 'kink' in change in lift due to ground effect tables for C172x an…

seanmcleod:
…d B737


bcoconni:
Quoting @seanmcleod from the ML
> I glanced at a number of other models in the JSBSim distribution and a number of them didn’t have the ground effect modeled at all, and others like the F-15 and F-16 had curves with different data, i.e. they hadn’t been a simple cut-paste of the C172x or B737 model’s data for ground effect.

Thanks Sean.

bcoconni:
Just for the record, we are using the same correction than the [c172p team PR #922](https://github.com/c172p-team/c172p-detailed/pull/922)

agodemar:
Well done.

PR #6, opened 2018-03-12
Adds a means to convert units before exporting property values.

bcoconni:
The unit conversion is made by defining a new type `template` of `<function>` elements which uses a dummy property `#`. The function is evaluated after having replaced the dummy property by the property value and the result is exported.

The `template` function is defined once and can be applied on as many properties as needed.
In addition, the user can create as many unit conversion functions as needed. These functions can be grouped in a separate file that can be included each time it is needed in a sort of library.

The new syntax is
```xml
<output>
  <function name="convertRadToDeg" type="template">
    <product>
      <property> # </property>
      <value> 57.29577951 </value>
    </product>
  </function>
  <property caption="flight-path/gamma-deg" apply="convertRadToDeg">
    flight-path/gamma-rad
  </property>
</output>
```


bcoconni:
This also addresses the [feature request #95](https://sourceforge.net/p/jsbsim/feature-requests/95/) on SourceForge.

PR #16, opened 2018-04-30
Add support for optional path for included files in FGOutput

seanmcleod:
Currently used by FGScript so that that Output elements within a script that include a file, e.g. unitconversions.xml for template functions can store the include file in the same directory as the script file rather than being forced to store it in the aircraft's directory.


bcoconni:
OK, it's fine with me.
Thanks for your patch.

PR #17, opened 2018-05-01
Add a set of template functions for unit conversions to SI units

seanmcleod:
Standard set of unit conversion template functions available for use across scripts.


bcoconni:
Thanks I would suggest to modify a script to use the feature and your file as an example.

seanmcleod:
I've added some example usage to the 737_cruise.xml script demonstrating how to use the unit conversion template functions in both script event notify elements and within an output element.

bcoconni:
Great !
Thanks.

PR #23, opened 2018-05-08
Update cython bindings for python 3.6 and 2.7. Updated Tests.

ezietsman:


seanmcleod:
Thanks @ezietsman. Do these changes mean both python 2.x and 3.x are now supported or just python 3.x?

In terms of the AppVeyor build for Windows at the moment our AppVeyor build uses the Visual Studio project file and not CMake which means the python API wrapper isn't built and the python tests aren't run.

tests/TestInputSocket.py
```diff
@@ -137,14 +137,14 @@ def setUp(self):
...
                         msg="Not connected to the JSBSim server.\nGot message '%s' instead" % (out,))
 
         # Check that "help" returns the minimum set of commands that will be
         # tested
-        self.assertEqual(sorted(map(lambda x: string.strip(string.split(x, '{')[0]),
-                                    string.split(_tn.sendCommand("help"), '\n')[2:-2])),
-                         ['get', 'help', 'hold', 'info', 'iterate', 'quit', 'resume', 'set'])
+        self.assertEqual(sorted(map(lambda x: x.split('{').strip()[0]),
+                                    _tn.sendCommand("help").split('\n')[2:-2]),
+                                    ['get', 'help', 'hold', 'info', 'iterate', 'quit', 'resume', 'set'])
```
bcoconni: According to my tests, this code has been incorrectly modified. It should rather be
```python
       self.assertEqual(sorted(map(lambda x: x.split('{')[0].strip(),
                                   _tn.sendCommand("help").split('\n')[2:-2])),
                               ['get', 'help', 'hold', 'info', 'iterate', 'quit', 'resume', 'set'])
```
Otherwise the test fails.

tests/TestInputSocket.py
```diff
@@ -179,14 +179,14 @@ def test_input_socket(self):
...
-        msg = string.split(tn.sendCommand("info"), '\n')
-        self.assertEqual(string.strip(string.split(msg[2], ':')[1]),
-                         string.strip(root.attrib['name']))
-        self.assertEqual(string.strip(string.split(msg[1], ':')[1]),
-                         string.strip(root.attrib['version']))
+        msg = tn.sendCommand("info").split('\n')
+        self.assertEqual(msg[2].split(':')[1],
+                         root.attrib['name'].strip())
+        self.assertEqual(msg[1].split(':')[1],
+                         root.attrib['version'].strip())
```
bcoconni: According to my tests, this code has been incorrectly modified. It should rather be
```python
       self.assertEqual(msg[2].split(':')[1].strip(),
                        root.attrib['name'].strip())
       self.assertEqual(msg[1].split(':')[1].strip(),
                        root.attrib['version'].strip())
```
Otherwise the test fails.

bcoconni:
@seanmcleod I have tested successfully this pull request with Python 2.7 but I had to make some changes that I have documented above. I will merge this PR as soon as the requested changes will be reviewed and included by @ezietsman

ezietsman:
Yes, I disabled fpectl completely. That module is not available in python 3 so I'll have to look at that in some more detail.

The build worked in python 2.7 and 3.6, besides for that one test. I'll have a look and fix the issues.

seanmcleod:
@ezietsman are you sure that fpectl isn't available in python 3? This documentation for python 3 states that it isn't built by default and discourages the use of it, as opposed to it not being available at all.

https://docs.python.org/3/library/fpectl.html

And anyway aren't we building the fpectl module ourselves anyway via fpectlmodule.cpp in the tests directory?

ezietsman:
I'll have a look at that source code and try to make it work.

seanmcleod:
Ah, forgot about the Mac OSX issue with building fpectl. It may be as simple as adding some Mac OSX ifdef to call feraiseexcept instead feenableexcept etc.

ezietsman:
I've added the fpect module back in, it seems to compile now using clang on macos but only for python 2.7. The problem here is that the way modules are wrapped in python 3 has changed. So this part:

    PyMODINIT_FUNC initfpectl(void)
    {
      PyObject *m = Py_InitModule("fpectl", fpectl_methods);
      if (m == NULL)
        return;
      PyObject *d = PyModule_GetDict(m);
      fpe_error = PyErr_NewException((char*)"fpectl.FloatingPointError",
                                     PyExc_FloatingPointError, NULL);
      if (fpe_error != NULL)
        PyDict_SetItemString(d, "FloatingPointError", fpe_error);
    }

needs to change when it is compiled using python 3.0. The new way well work in 2.7 too. I'll investigate further.

ezietsman:
https://docs.python.org/3/howto/cporting.html I think explains it. I'll look at it later at home.

bcoconni:
@ezietsman thanks for your work, this is progressing very well. On my side I'm working on adding a small test to the `fpectl`module to make sure that it transforms a FP exception to a Python exception when it is turned on and that FP exceptions are ignored otherwise. That test on `fpectl` will be part of the test suite.

Also for the record, our version of `fpectl` is designed to halt the tests when a FP exception is raised. We won't try to recover from such an exception and continue execution. This should simplify the design as all we need is to make sure is that the Python interpreter is capable to quit safely. I'll document that as well.

bcoconni:
@ezietsman I saw that you modified the port number of the Telnet session from 1137 to 2222 in `TestInputSocket`. Was that the reason why the test was failing on Mac OSX ? Is the port number 1137 already used/reserved on this platform ?

kkz003:
I cannot build the above code 

> I've added the fpect module back in, it seems to compile now using clang on macos but only for python 2.7. The problem here is that the way modules are wrapped in python 3 has changed. So this part:
> 
> ```
> PyMODINIT_FUNC initfpectl(void)
> {
>   PyObject *m = Py_InitModule("fpectl", fpectl_methods);
>   if (m == NULL)
>     return;
>   PyObject *d = PyModule_GetDict(m);
>   fpe_error = PyErr_NewException((char*)"fpectl.FloatingPointError",
>                                  PyExc_FloatingPointError, NULL);
>   if (fpe_error != NULL)
>     PyDict_SetItemString(d, "FloatingPointError", fpe_error);
> }
> ```
> 
> needs to change when it is compiled using python 3.0. The new way well work in 2.7 too. I'll investigate further.

I cannot build this part on python3.7 and ubuntu 18.04.
I am very beginner. Please give me any advice to solve this issue and build jsbsm.

cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
/3rd-party/src/jsbsim/tests/fpectlmodule.cpp: In function ‘PyObject* initfpectl()’:
/3rd-party/src/jsbsim/tests/fpectlmodule.cpp:130:17: error: ‘Py_InitModule’ was not declared in this scope
   PyObject *m = Py_InitModule("fpectl", fpectl_methods);
                 ^~~~~~~~~~~~~
/3rd-party/src/jsbsim/tests/fpectlmodule.cpp:130:17: note: suggested alternative: ‘Py_Initialize’
   PyObject *m = Py_InitModule("fpectl", fpectl_methods);
                 ^~~~~~~~~~~~~
                 Py_Initialize
/3rd-party/src/jsbsim/tests/fpectlmodule.cpp:132:5: error: return-statement with no value, in function returning ‘PyObject* {aka _object*}’ [-fpermissive]
     return;
     ^~~~~~
error: command 'gcc' failed with exit status 1

bcoconni:
@kkz003 
Please open a new issue in https://github.com/JSBSim-Team/jsbsim/issues for that topic.

PR #33, opened 2018-05-21
Add density altitude calculation to FGStandardAtmosphere

seanmcleod:


seanmcleod:
@djlinse thanks for the comments. Yep the equation which I grabbed from a much earlier version of JSBSim is only valid up to 36089ft since it uses a single lapse rate. It seems to be the common/standard formula you come across on Wikipedia etc. for density altitude up to 36089ft assuming dry air.

But until we implement one that is valid at higher altitudes we should document it's current limitation. I'll take a look at the link you posted.

I'll also look into the code structure suggestions.

seanmcleod:
@bcoconni @djlinse on some further studying of the atmosphere code I noticed a potential issue with a difference between the geometric altitude (GeoMet) and the geopotential altitude (GeoPot). 

Take a look at the following code from the test snippet that @bcoconni submitted. 

```
jsbSim_density_alt = fdm['atmosphere/density-altitude']
# Check that when setting the aircraft at the altitude jsbSim_density_alt with delta-T == 0.0
# we get the same density than at the pressure altitude with delta-T != 0.0
density_ref = fdm['atmosphere/rho-slugs_ft3']

fdm['atmosphere/delta-T'] = 0.0
fdm['ic/h-agl-ft'] = jsbSim_density_alt
fdm.run_ic()

self.assertAlmostEqual(density_ref, fdm['atmosphere/rho-slugs_ft3'], delta=1e-5)
```

The altitude passed to `fdm['ic/h-agl-ft']` is assumed to be a GeoMet altitude and so the atmosphere code converts it to a GeoPot altitude before making use of the atmosphere tables etc.

And in the current implementation of the `DensityAltitude()` function we return a GeoPot altitude, which the test then passes back in via `fdm['ic/h-agl-ft'] = jsbSim_density_alt` as a GeoMet altitude.

If I return a GeoMet altitude from the `DensityAltitude()` function then the error differences improve by 1-2 orders of magnitude from 10,000ft upwards, no real difference for the sea-level test.

Original test results:

| Altitude | Delta-T | Density Diff (rho-slugs_ft3) |
| --- | --- | --- |
| 0 | 0 | 0 |
| 0 | -27 | -3.17e-5 |
| 0 | 27 | 3.45e-9 |
| 10000 | 0 | 0 |
| 10000 | -27 | -1.57e-7 |
| 10000 | 27 | -2.53e-7 |
| 20000 | 0 | 0 |
| 20000 | -27 | -7.04e-7 |
| 20000 | 27 | -7.62e-7|

Updated test results with `DensityAltitude()` returning a GeoMet altitude.

| Altitude | Delta-T | Density Diff (rho-slugs_ft3) |
| --- | --- | --- |
| 0 | 0 | 0 |
| 0 | -27 | -3.17e-5 |
| 0 | 27 | 2.31e-9 |
| 10000 | 0 | 0 |
| 10000 | -27 | 9.80e-9 |
| 10000 | 27 | 1.16e-8 |
| 20000 | 0 | 0 |
| 20000 | -27 | 1.59e-8 |
| 20000 | 27 | 1.62e-8|

PR #36, opened 2018-05-27
Remove the effective cos(alpha) scaling of CAS

seanmcleod:


bcoconni:
This PR seems to arrive a bit too early. I didn't understand that we came to a conclusion on the mailing list.

seanmcleod:
Alan and I are in agreement that there shouldn't be an effective cos(alpha) factor applied to CAS. Tony replied with the following after I sent the link to the NACA report on pitot testing at different AoAs.

"So, at this point, I’ll concede that the error due to design and/or installation varies too much and that the only reliable one is the sea level assumption built into the CAS equations."

So with this commit that's what we end up with, i.e. "the sea level assumption built into the CAS equations" via VcalibratedFromMach.

There is still debate on what sorts of error modeling and how to go about doing the error modeling can/should be applied, but in my view there is general agreement between Alan, myself and Tony that we can't use the current effective cos(alpha) error factor.

I'll look into uninitialized variables/members.

bcoconni:
> There is still debate on what sorts of error modeling and how to go about doing the error modeling can/should be applied,

This is what I meant: apart from the fact that the `cos(alpha)` correction must be dropped there is no clear vision about what should be done next. So this patch seems a bit premature to me. We might remove things that will be reintroduced later once we agreed on the path forward.

bcoconni:
For the record, there is some code for the Pitot tube angle in the following files:
- `src/FGFDMExec.cpp`
- `src/models/FGAircraft.h`
- `src/models/FGAircraft.cpp`
- `src/models/FGAuxiliary.h`
- `src/models/FGAuxiliary.cpp`
- `src/initialization/FGInitialization.cpp`

The patch should remove all that code otherwise there will be bits of code for the Pitot angle that will remain here and there in JSBSim code base.

seanmcleod:
You're right in terms of the remaining bits of code related to the Pitot angle. I've gone through the files and it looks like the remaining references to the Pitot angle all related to modifying CAS and so I've removed them and submitted an additional commit for this pull request.

src/models/FGAuxiliary.cpp
```diff
@@ -85,7 +85,6 @@ FGAuxiliary::FGAuxiliary(FGFDMExec* fdmex) : FGModel(fdmex)
   vAeroPQR.InitMatrix();
   vMachUVW.InitMatrix();
   vWindUVW.InitMatrix();
```
bcoconni: You should remove the member `vWindUVW` as well since it has been added specifically for the Pitot angle computations (commit 32cacb4).

tests/CMakeLists.txt
```diff
@@ -41,7 +41,6 @@ set(PYTHON_TESTS ResetOutputFiles
                  TestScriptOutput
                  CheckSimTimeReset
                  TestHoldDown
-                 TestPitotAngle
```
bcoconni: You should delete the test source code as well
```bash
$ git rm tests/TestPitotAngle.py
```

seanmcleod:
Added a new commit to remove vWindUVW and to remove TestPitotAngle.py

PR #38, opened 2018-05-29
Add test for testing the density altitude function

seanmcleod:
First test implementation. Let me know if there are better ways of performing this sort of test.

### DIFF
--- a/tests/TestDensityAltitude.py
+++ b/tests/TestDensityAltitude.py
@@ -1,4 +1,4 @@
-# TestWaypoint.py
+# TestDensityAltitude.py
 #
 # Test that density altitude works 
 #
--- a/tests/TestDensityAltitude.py
+++ b/tests/TestDensityAltitude.py
@@ -1,0 +1,65 @@
+# TestWaypoint.py
+#
+# Test that density altitude works 
+#
+# Copyright (c) 2018 Sean McLeod
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, see <http://www.gnu.org/licenses/>
+#
+
+from JSBSim_utils import JSBSimTestCase, CreateFDM, RunTest
+
+class TestDensityAltitude(JSBSimTestCase):
+
+    def test_densityaltitude(self):
+        fdm = CreateFDM(self.sandbox)
+        
+        script_path = self.sandbox.path_to_jsbsim_file('scripts', '737_cruise.xml')
+        fdm.load_script(script_path)
+
+        # Reference data (Pressure Altitude, Temp Delta (Rankine), Density Altitude)
+        reference_data = [
+            (0, 0, 0),
+            (0, -27, -1838.73),
+            (0, 27, 1724.16),
+            (10000, 0, 10000),
+            (10000, -27, 8839.83),
+            (10000, 27, 11113.20),
+            (20000, 0, 20000),
+            (20000, -27, 19508.06),
+            (20000, 27, 20493.17)
+            ]
+
+        # Run through refernce data and compare JSBSim calculated density altitude to expected
+        for data in reference_data:
+            pressure_alt = data[0]
+            delta_T = data[1]
+            density_alt = data[2]
+
+            fdm['ic/h-agl-ft'] = pressure_alt
+            fdm.run_ic()
+            fdm['atmosphere/delta-T'] = delta_T
+
+            jsbSim_density_alt = fdm['atmosphere/density-altitude']
+
+            self.assertAlmostEqual(density_alt, jsbSim_density_alt, delta=1)
+
+            print('Pressure Altitude: {0}   Delta-T: {1}   Density Altitude: {2}'.format(pressure_alt, delta_T, jsbSim_density_alt))
+
+        del fdm
+
+
+RunTest(TestDensityAltitude)
+
+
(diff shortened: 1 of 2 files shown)


tests/TestDensityAltitude.py
```diff
@@ -0,0 +1,65 @@
...
+            fdm.run_ic()
+            fdm['atmosphere/delta-T'] = delta_T
+
+            jsbSim_density_alt = fdm['atmosphere/density-altitude']
+
+            self.assertAlmostEqual(density_alt, jsbSim_density_alt, delta=1)
+
+            print('Pressure Altitude: {0}   Delta-T: {1}   Density Altitude: {2}'.format(pressure_alt, delta_T, jsbSim_density_alt))
+
+        del fdm
```
bcoconni: Just to make sure that the density altitude computation is consistent with the rest of the atmosphere code, I would add additional tests to check that we get the same density when setting the aircraft at the altitude `density_alt` with `delta-T=0` (which is the definition of the density-altitude if I remember correctly)

This would look like (**disclaimer:** *this is untested code off the top of my head* :-)
```python
  density_ref = fdm['atmosphere/density']

  fdm['ic/h-agl-ft'] = density_alt
  fdm.run_ic()
  fdm['atmosphere/delta-T] = 0

  self.assertAlmostEqual(density_ref, fdm['atmosphere/density'])
```

seanmcleod: You're effectively saying that the density altitude and pressure altitude should be equal assuming a standard atmosphere with no temperature delta? I've included that check in my original reference data, i.e. you'll see a delta-T of 0 for all 3 altitudes.

seanmcleod:
Thanks for the comments/suggestions. I've added a new commit implementing them. See my separate comment to your comment regarding an additional test for delta-T = 0.

tests/TestDensityAltitude.py
bcoconni: No this is not what I mean.

The assertion you wrote is testing the C++ method `FGStandardAtmosphere::GetDensityAltitude()`. My point is that we must also check that the density altitude you get from there is consistent with the density computations made in the C++ class `FGStandardAtmosphere` (*spoiler: it is not*)

I have fixed my code (and tested it this time), it should be
```python
        for pressure_alt, delta_T, density_alt in reference_data:

            fdm['ic/h-agl-ft'] = pressure_alt
            fdm['atmosphere/delta-T'] = delta_T
            fdm.run_ic()

            jsbSim_density_alt = fdm['atmosphere/density-altitude']

            self.assertAlmostEqual(density_alt, jsbSim_density_alt, delta=1e-7)

            # Check that when setting the aircraft at the altitude jsbSim_density_alt with delta-T == 0.0
            # we get the same density than at the pressure altitude with delta-T != 0.0
            density_ref = fdm['atmosphere/rho-slugs_ft3']

            fdm['ic/h-agl-ft'] = jsbSim_density_alt
            fdm['atmosphere/delta-T'] = 0.0
            fdm.run_ic()

            self.assertAlmostEqual(density_ref, fdm['atmosphere/rho-slugs_ft3'])
```
This fails with the following error
```
 FAIL: test_densityaltitude (__main__.TestDensityAltitude)
   File "TestDensityAltitude.py", line 61, in test_densityaltitude
     self.assertAlmostEqual(density_ref, fdm['atmosphere/rho-slugs_ft3'])
 AssertionError: 0.002507425951046395 != 0.0025391558128942003 within 7 place
```
which means that the formula you are using to compute the density altitude is not consistent with the other formulas in FGStandardAtmosphere that computes the density.

This must be fixed.

seanmcleod: Okay I follow.

Initial quick test to get a feel for the size of the error.

| Altitude | Delta-T | Density Diff (rho-slugs_ft3) |
| --- | --- | --- |
| 0 | 0 | 0 |
| 0 | -27 | -3.17e-5 |
| 0 | 27 | 3.45e-9 |
| 10000 | 0 | 0 |
| 10000 | -27 | -1.57e-7 |
| 10000 | 27 | -2.53e-7 |
| 20000 | 0 | 0 |
| 20000 | -27 | -7.04e-7 |
| 20000 | 27 | -7.62e-7|

On initial inspection I see there is a slight difference in the lapse rate, `0.0035661600044609404` versus `0.00356616` but changing it makes virtually no difference, none of the density errors change by an order of magnitude.

I'll keep looking.

seanmcleod: @bcoconni I haven't been able to improve on the density error differences I listed above.

bcoconni:
@djlinse That's good to see you back on board, Dennis. I am in agreement with your statement.

tests/TestDensityAltitude.py
seanmcleod: @bcoconni @djlinse on some further studying of the atmosphere code I noticed a potential issue with a difference between the geometric altitude (GeoMet) and the geopotential altitude (GeoPot). 

Take a look at the following code from the test snippet that @bcoconni submitted. 

```
jsbSim_density_alt = fdm['atmosphere/density-altitude']
# Check that when setting the aircraft at the altitude jsbSim_density_alt with delta-T == 0.0
# we get the same density than at the pressure altitude with delta-T != 0.0
density_ref = fdm['atmosphere/rho-slugs_ft3']

fdm['atmosphere/delta-T'] = 0.0
fdm['ic/h-agl-ft'] = jsbSim_density_alt
fdm.run_ic()

self.assertAlmostEqual(density_ref, fdm['atmosphere/rho-slugs_ft3'], delta=1e-5)
```

The altitude passed to `fdm['ic/h-agl-ft']` is assumed to be a GeoMet altitude and so the atmosphere code converts it to a GeoPot altitude before making use of the atmosphere tables etc.

And in the current implementation of the `DensityAltitude()` function we return a GeoPot altitude, which the test then passes back in via 'fdm['ic/h-agl-ft'] = jsbSim_density_alt' as a GeoMet altitude.

If I return a GeoMet altitude from the `DensityAltitude()` function then the error differences improve by 1-2 orders of magnitude from 10,000ft upwards, no real difference for the sea-level test.

| Altitude | Delta-T | Density Diff (rho-slugs_ft3) |
| --- | --- | --- |
| 0 | 0 | 0 |
| 0 | -27 | -3.17e-5 |
| 0 | 27 | 2.31e-9 |
| 10000 | 0 | 0 |
| 10000 | -27 | 9.80e-9 |
| 10000 | 27 | 1.16e-8 |
| 20000 | 0 | 0 |
| 20000 | -27 | 1.59e-8 |
| 20000 | 27 | 1.62e-8|

PR #41, opened 2018-05-31
make JSBSim build on FreeBSD (lsb_release not existing)

alios:
On FreeBSD` lsb_release` does not exist and thats why the `cmake` step fails.
This patch checks if we are on a FreeBSD systems and does not execute the linux distrib detection, but instead sets `LINUX_DISTRIB_NAME` to `FreeBSD`.

Now JSBSim Builds fine on FreeBSD.


bcoconni:
The `LINUX_DISTRIB_NAME` is really only used for RPM or Debian packaging so it may not be relevant for FreeBSD but since it does not hurt I merged your PR.

Thanks.

PR #42, opened 2018-06-01
Update appveyor to use correct tag for releases

seanmcleod:
Change tag name for releases to JSBSim-trusty-v2018a.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,8 +40,8 @@ skip_tags: true
 
 deploy:
 - provider: GitHub
-  tag: JSBSim-win64-v2018a
-  description: VS 2017 - Release - x64
+  tag: JSBSim-trusty-v2018a
+  description: Windows - Release - x64
   auth_token:
     secure: YWG8FRFWv/ylTuA1DIFNcN01uFGz7xKit8CayOMTE0orPKXhztwLmwm8MZ4IzFbr
   artifact: build\src\JSBSim.exe,build\utils\aeromatic++\aeromatic.exe


bcoconni:
@seanmcleod
"trusty" corresponds to the name of the Linux distribution [Ubuntu 14.04 "Trusty Tahr"](http://releases.ubuntu.com/14.04/).

@agodemar
I understood you meant to have separate releases for Windows and Linux ?

seanmcleod:
Oh. I assumed the idea was to have a single/the same tag used.

Actually I'm a bit confused. So at the moment on each commit that passes the CI or is it only when you guys perform a release/deploy that commit or the head of the git source is tagged with a fixed set of tags, i.e. 'JSBSim-win64-v2018a' and 'JSBSim-trusty-v2018a' each time?

Doesn't this mean we have multiple commits in the git repository with the same tag names? Or does using the same tag name replace/remove the older ones?

bcoconni:
@seanmcleod 

> Oh. I assumed the idea was to have a single/the same tag used.

Now that I spent quite some time to make Travis CI deploy binaries under a unique release name, I can tell that it is perfectly feasible to release the Linux binaries under the same release name than Windows binaries. Would we choose that option, I would suggest to rename the Windows release name because `JSBSim-win64-v2018` does sound a bit Windows focused :-)
> So at the moment on each commit that passes the CI or is it only when you guys perform a release/deploy that commit or the head of the git source is tagged with a fixed set of tags, i.e. 'JSBSim-win64-v2018a' and 'JSBSim-trusty-v2018a' each time?

Regarding the releases made by Travis CI: everything is automated now. This means that each time a commit is pushed on the branch `master`, JSBSim is automatically built and tested. And if the tests pass then the binaries resulting from the commit are automatically deployed on the JSBSim GitHub project. No human intervention in the process, that's the beauty of continuous integration :-)
> Doesn't this mean we have multiple commits in the git repository with the same tag names? Or does using the same tag name replace/remove the older ones?

Regarding Travis CI, I have set the deployment process to overwrite existing files. However as a safety net, I also instructed the CI process to tag the binaries file names with the build ID. So at the moment, the Linux binaries all have a unique file name and there is no risk of confusion. However there is the risk of cluttering the release if no one cleans up the old binaries from time to time. But even that clean up action can be automated if we would need to.

I think that @agodemar chose a different option where each new released binaries overwrite the existing ones.

This is also the reason why I pushed the commit 7116e1d so that we can check where an executable is coming from. Suppose a user downloads an executable and then reports an issue. If we cannot reproduce the bug, we would like to know which source code has been used to build the executable (i.e. which commit ID). Since commit 7116e1d, if the executable has been built by Travis CI or AppVeyor, then the command `JSBSim --version` will report a message like
```
JSBSim Version: 1.0.0 [build 125/commit deadbeef] Jun 3 2018 11:34:39

seanmcleod:
@bcoconni I see @agodemar merged in my commit to the master branch changing the tag name used by AppVeyor to `JSBSim-trusty-v2018a` (as I mentioned I was confused about the tag name, didn't realise it referred to a specific Linux version).

Looking at the Releases now on GitHub it looks like you've added some descriptive text explaining which assets are Windows binaries, which are Linux binaries etc. So if we're going to continue with this tag setup I guess we should change the tag name to something platform neutral, i.e. don't include `trusty` in the tag name.

agodemar:
I propose to freeze the release tag name as `JSBSim-v2018a` and stick to that until the rest of the year. Starting from Oct-2018 we could switch to `v2018b` and so on.

PR #43, opened 2018-06-03
Avoid checking psi/phi if theta == 90 (gimbal lock)

seanmcleod:
Otherwise TestInitialConditions.py can fail

### DIFF
--- a/tests/TestInitialConditions.py
+++ b/tests/TestInitialConditions.py
@@ -351,7 +351,8 @@ class TestInitialConditions(JSBSimTestCase):
     def PhiOrPsiAndGimbalLock(self, var, vars):
         # Look out for quaternion singularity/gimbal-lock when ic/theta == 90
         # and avoid checking psi and phi in this case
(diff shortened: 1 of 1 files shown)


tests/TestInitialConditions.py
```diff
@@ -342,4 +348,11 @@ def test_set_initial_geodetic_latitude(self):
         self.assertEqual(ref['Time'][0], 0.0)
         self.assertAlmostEqual(ref['Latitude Geodetic (deg)'][0], glat)
 
+    def PhiOrPsiAndGimbalLock(self, var, vars):
+        # Look out for quaternion singularity/gimbal-lock when ic/theta == 90
+        # and avoid checking psi and phi in this case
+        if (var['tag'] == 'psi' or var['tag'] == 'phi') and [x for x in vars if x['tag'] == 'theta' and x['value'] == 90]:
```
bcoconni: Regarding the test `x['value'] == 90` : checking the equality of 2 floating point values is evil :-)
Please do something as
```python
abs(x['value']-90) < 1E-7
```

seanmcleod: @bcoconni less evil now with the new commit ;-)

PR #50, opened 2018-06-05
[Jupyter Notebooks] More work on doc/basic example 01

agodemar:


seanmcleod:
@agodemar typically you should exclude the `.ipynb_checkpoints` directory from the source code repository, add it to .gitignore etc. and remove the directory and files in it from the git repository.

Issue #44, opened 2018-06-03
Improvements to the density altitude computation


bcoconni:
### Intro
In addition to the discussion that is on going on PR #38, @djlinse made some comments on the original PR #33. But since PR #33 is closed, I have created this new issue and copied Dennis' comments here so that we can keep track of the discussion.

Below are Dennis' comments (*I fixed the layout a bit. Especially the URL which was not interpreted as a link by GitHub*)

### Discussion
1. As far as I've been able to tell (and I may be wrong!), the equations are only valid until the boundary of the troposphere and tropopause, i.e. 11000km/36089 feet.  They appear to be Eqn 17, the "Simpler Method of Calculation..." for density altitude given in the extended discussion (with multiple examples/calculators) found here: https://wahiduddin.net/calc/density_altitude.htm

Note that the Standard Atmosphere is dry air, i.e. 0% humidity.

As `FGStandardAtmosphere` goes all of the way up to 86km, this addition should either complete the calculations correctly for all altitudes, or clearly note where they are valid and where they aren't valid.

2. Even if correct, I would suggest that the constants (518.xxx, etc.) be written as constants (with units as a comment), otherwise they are just magic numbers.  Preferably with a reference to textbook or maybe the Standard Atmosphere document itself.

3. Is PressureAltitude the correct altitude to use here?  Even if it is identical to altitude in the current code, that may not always be the case.

4. Code structure-wise, I wonder if a "better" way to implement the Density altitude would *not* be to override the GetDensityAltitude() method from FGAtmosphere, but rather use the already existing DensityAltitude variable, and have the base class Calculate() method call
```c++
DensityAltitude = CalculateDensityAltitude(altitude);
```
instead of just setting it to `DensityAltitude = altitude`.  The base class would then have a virtual method
```c++
virtual double CalculateDensityAltitude(const double altitude) {return altitude;}
```
And similarly for pressure altitude.

Sean's new function would then be CalculateDensityAltitude() with the altitude passed in ("solving" the pressure altitude question).  And the getter functionality would just be getting the current density altitude, as is typically expected for getter functions.

If you don't want to make this change, then at least set the DensityAltitude variable.

seanmcleod:
@djlinse thanks for the comments. Yep the equation which I grabbed from a much earlier version of JSBSim is only valid up to 36089ft since it uses a single lapse rate. It seems to be the common/standard formula you come across on Wikipedia etc. for density altitude up to 36089ft assuming dry air.

[...]

PR #56, opened 2018-06-07
Return geometric alt for density alt, plus code cleanup

seanmcleod:
Return the geometric altitude from the density altitude calculation.
Code structure changes as per @djlinse's suggestions.
Document the valid range for the density altitude calculation and only use it in this range (toposphere). 
Document and show calculation of 'magic' numbers.

```c++
// Earth radius in ft as defined for ISA 1976 
const double FGStandardAtmosphere::EarthRadius = 20855531.5;

// 518.67 / 0.00356616
// Standard sea level temp / toposphere lapse rate
DAToposphereFactor = -(*StdAtmosTemperatureTable)(1, 1) / LapseRateVector[0]; 
```

Replace 4 instances of:

```c++
// From
double GeoPotAlt = (altitude*20855531.5)/(20855531.5+altitude);
// To
double GeoPotAlt = GeopotentialAltitude(altitude);
```

### DIFF
--- a/src/models/atmosphere/FGStandardAtmosphere.cpp
+++ b/src/models/atmosphere/FGStandardAtmosphere.cpp
@@ -57,6 +57,9 @@ IDENT(IdHdr,ID_STANDARDATMOSPHERE);
 CLASS IMPLEMENTATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
+// Earth radius in ft as defined for ISA 1976
+const double FGStandardAtmosphere::EarthRadius = 20855531.5;
+
 FGStandardAtmosphere::FGStandardAtmosphere(FGFDMExec* fdmex) : FGAtmosphere(fdmex),
                                                                TemperatureBias(0.0),
                                                                TemperatureDeltaGradient(0.0)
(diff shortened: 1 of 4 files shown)


src/models/atmosphere/FGStandardAtmosphere.cpp
```diff
@@ -126,6 +130,15 @@ bool FGStandardAtmosphere::InitModel(void)
...
   CalculateLapseRates();
   CalculatePressureBreakpoints();
+
+  // Density altitude parameters
+  // Density altitude formula only valid up until top of the Troposhere
+  TroposphereMaxAltitude = GeometricAltitude((*StdAtmosTemperatureTable)(2, 0));
+  // Standard sea level temp / Troposphere lapse rate
+  DATroposphereFactor = -(*StdAtmosTemperatureTable)(1, 1) / LapseRateVector[0];
+  // Unitless exponent computed using SI values from LR/(g0M - L*R) 
+  DATroposphereExponent = 0.2349781324440659;
```
bcoconni: You made `DATroposphereFactor` depend on the value of `LapseRateVector[0]` but hardcoded the exponent `DATroposphereExponent` while its depends also of `LapseRateVector[0]` ?

src/models/atmosphere/FGStandardAtmosphere.cpp
```diff
@@ -452,20 +465,29 @@ void FGStandardAtmosphere::ResetSLPressure()
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+// Calculates the density altitude given any temperature or pressure bias.
+// Currently the formula used is only valid up until the top of the Troposphere,
+// 11.019km - 36151.8ft.
+// https://en.wikipedia.org/wiki/Density_altitude
+// https://wahiduddin.net/calc/density_altitude.htm
```
bcoconni: These comments should be moved to the header `FGStandardAtmosphere.h` so that they appear in the [project Doxygen documentation](https://jsbsim-team.github.io/jsbsim/classJSBSim_1_1FGStandardAtmosphere.html)

For Doxygen to parse correctly the comments, the syntax in `FGStandardAtmosphere.h` should be similar to
```c++
/** Calculates the density altitude given any temperature or pressure bias.
 Currently the formula used is only valid up until the top of the Troposphere,
 11.019km - 36151.8ft.
@see
 https://en.wikipedia.org/wiki/Density_altitude
 https://wahiduddin.net/calc/density_altitude.htm
*/
```

bcoconni:
I have merged the PR #38 to help review the code.
The comments I made for an additional test condition are still applicable so I moved them to the issue #44

src/models/atmosphere/FGStandardAtmosphere.cpp
seanmcleod: Yep, see the discussion I had with Dennis with regards to the exponent on issue [44](https://github.com/JSBSim-Team/jsbsim/issues/44)

PR #60, opened 2018-06-10
Caclulate pressure altitude for non standard ISA day

seanmcleod:
@bcoconni @djlinse I've added some code to calculate the pressure altitude for a non standard ISA day.

I've followed the existing pattern of no `const` and I've made both `CalculateDensityAltitude()` and `CalculatePressureAltitude()` protected as per @bcoconni's comment on the density altitude thread.

The exponent is calculated using the same `Mair` and `Rstar` members that are used when performing the existing pressure calculations which means there is a very close match, on the order of 1e-11, when the Python test case compares the pressures.


tests/TestPressureAltitude.py
```diff
@@ -0,0 +1,71 @@
...
+            fdm['atmosphere/delta-T'] = 0.0
+            fdm.run_ic()
+
+            self.assertAlmostEqual(pressure_ref, fdm['atmosphere/P-psf'])
+
+        del fdm
+
+
+RunTest(TestPressureAltitude)
+
```
bcoconni: Just a minor comment : `TestPressureAltitude.py` is exactly the same code than `TestDensityAltitude.py` with the exception of the `reference_data` array and the tested property (`atmosphere/pressure-altitde` vs `atmosphere/density-altitude`).

The common code of these 2 files should be factorized.

djlinse:
Oops!  I opened this, but didn't finish the comment earlier.  And now you've merged it...

Here's what I wrote:

On looking at some other references, and thinking about it, this check [altitude > Troposphere altitude] is not the "boundary" to check for.  Since pressure altitude comes from a pressure measurement, what you need to compare is the pressure from GetPressure[i.e. the non-standard value that is the "measurement"] to the pressure in the standard atmosphere at the troposphere boundary.  If the pressure is greater than the standard value, then you use the equation given as you are still in the troposphere.  If it is less than the standard value, i.e. above 11000km in the standard atmosphere, then you have to use equation 33(b) from the ISA document.

In fact, to make the function fully general up to 86km, you just have to switch between using 33(a) and 33(b) depending on if there is a non-zero lapse rate or not, applying the appropriate boundary conditions at each layer.  This boundary conditions are already computed.

bcoconni:
@djlinse 
Hmm, I have asked myself the same questions but did not come to the same conclusion. Actually it all boils down to what is modified in the atmosphere when we are deviating from the standard ISA conditions and what remains fixed/immutable ?

The implementation of `FGStandardAtmosphere` is built on the assumption that the **altitudes of transition** between the different layers of the atmosphere **are fixed**. This means in particular that the troposphere and tropopause thicknesses are constant.

Now if you assume that the transition between the troposphere and the tropopause is made when the pressure equals the standard pressure at 11 km then this means that the troposphere thickness will vary. I guess this is a reasonable assumption as well but it does not fit with `FGStandardAtmosphere`.

djlinse:
@bcoconni 
Think of it this way.  What is pressure altitude?  It is the equivalent altitude in the standard atmosphere that has the same static pressure as you measure outside the aircraft.  So the measurement that you have is Pstatic.  It can be affected by temperature, etc.  There is *no* inherent knowledge of geometric altitude.  In my university days, I would take that static pressure value to a table in the back of a book (Roskam in my case as I studied at Kansas where he taught), scan down the pressure column until I read that value, then scan across to the ISA_altitude corresponding to that pressure.  (Of course, interpolating as necessary between table values.)

Yes, the altitudes of transition in the standard atmosphere are fixed, but pressure altitude is a fictitious altitude.  Just the true static pressure looked up in a known table of (IAS_pressure,ISA_altitude)

In reality, altitude shouldn't be the argument to the CalculatePressureAltitude() function in the first place!  It should be the "measured" pressure.  Maybe that would make it clearer.  That is what the routine is really asking.  The same could/should be said for CalculateDensityAltitude().

We (especially me!) sometimes get confused because we have perfect knowledge in a simulation, so it is easy to use whatever value we want.  In this case, I think going back to what is being measure and presented helps out.  CalculatePressureAltitude() is nothing more than a "table-look-up" of the function ISA_altitude(measured_static_pressure).

djlinse:
Looking a bit more at the code, in FGAtmosphere, Calculate already calculates the Pressure and Density, so those don't need to be calculated.  Maybe the calls should be

  PressureAltitude = CalculatePressureAltitude(Pressure, altitude);
  DensityAltitude = CalculateDensityAltitude(Density, altitude);

where the second argument is only used as the fallback value if the actual value is not calculated from Density or Pressure, such as in the default FGAtmosphere() versions.

Of course, even passing the values of Pressure and Density is not required, as they are members, but it is clearer that the call depends on a pressure or density, and not just the altitude.

seanmcleod:
@djlinse yes for a non standard ISA day I'm calling `GetPressure(altitude)` and `GetTemperature(altitude)` again within the `CalculateDensityAltitude(altitude)` when they have already been called and calculated in `FGAtmosphere::Calculate(altitude)`. So we could definitely use the pre-calculated values, either passed in or via the member variables. The same applies for `CalculatePressureAltitude(altitude)`.

In fact looking at `FGAtmosphere::Calculate(altitude)` I noticed that my current implementation of `CalculateDensityAltitude` and `CalculatePressureAltitude` is a bit flawed in that it doesn't respect the `atmosphere\override\*` properties. And they would be if the density and pressure are passed in or accessed via the member variables.

I was originally using the passed in altiude in the standard ISA day case to return it immediately without applying the lookup formula, for performance and accuracy, but I see in a recent commit by @bcoconni that for the `CalculateDensityAltitude()` case he's removed that check stating that the formula is 'robust' enough after he updated some of the constants.

seanmcleod:
@djlinse going up through the thread, in terms of your post regarding the altitude boundary check at the time I added the current boundary check based on geometric altitude it was as a quick placeholder without too much thought based on your comment that the formula used was only valid in the troposphere. In the meantime I've simply been trying to get the formula accurate enough for the troposphere and working on code structure change suggestions ;-)

But the next target was to extend the `DensityAltitude` and `PressureAltitude` functions to work for altitudes above the troposphere. I would agree with your thinking with regards to how we go about checking the 'boundary condition' within functions like `CalculatePressureAltitude` etc. to determine which formula to use.

Here is a plot I generated via JSBSim plotting geometric altitude versus pressure for an ISA standard day and for one with a deltaT of +27F and -27F - [Altitude Pressure Plot](https://1drv.ms/u/s!AiGCTsLLA9i1gvorQ4-BB8noD3O3JA).

So at a geometric altitude of 36,000ft, i.e. within 100-200ft below the troposphere boundary we have roughly:

| Delta T | Pressure (psf) |
| --- | --- |
| -27 | 431 |
| 0 | 474 |
| 27 | 516 |

So on the non standard day of -27F deltaT when we convert the pressure of 431 psf to the equivalent pressure altitude for the standard ISA day (basically finding the same pressure on the blue line) we will need to use the formula for the lower stratosphere. But for the +27F deltaT day we will use the troposphere formula.

djlinse:
@seanmcleod I hadn't remembered that you need to check the pressure boundaries, not the altitude boundaries, until I examined some other code I have worked on.

On a -27F deltaT day, the steps would be
1. For the geometric altitude, calculate the non-standard atmosphere static pressure, i.e. start at the altitude on the left axis and go horizontally to the red line.
2. Go vertically at a constant pressure to the blue line.
3. Go horizontally back to the left axis to determine the pressure altitude.

For the full ISA, the blue line should/would "stop" at ~474 pdf (whatever the ISA boundary pressure is) and then follow a different curve based on equation 33(b) of the ISA up to the next boundary (~65Kft).  But the pressure altitude procedure is the same three steps: Calculate static pressure in current atmosphere, find where that pressure intersects the ISA pressure vs altitude plot, and report that altitude as the pressure altitude (taking into account geometric vs. geopotential).

seanmcleod:
@djlinse just to be clear the blue line etc. that I plotted wasn't based on me using the 33(a) equation from 0ft to 65,000ft. I queried JSBSim for pressure values and so internally JSBSim switched between using equation 33(a) and 33(b). I then simply plotted the returned pressure values from JSBSim.

```python
    def get_pressures(self, fdm, deltaT):
        pressures = []

        for geoalt in range(0, 65000, 1000):
            fdm['ic/h-sl-ft'] = geoalt
            fdm['atmosphere/delta-T'] = deltaT
            fdm.run_ic()
            pressure = fdm['atmosphere/P-psf']
            pressures.append(pressure)

        return pressures
```

I was expecting to see some slight 'kink' in the lines I plotted as the altitude switched from the troposphere to the lower stratosphere, but they look pretty smooth and continuous.

djlinse:
@seanmcleod OK, I didn't realize that about the lines.

Not necessarily surprising that there isn't a 'kink' going from a non-integer power to an exponential that have been designed to fit together.  You might see something made a finer grid around the transition region and plotted the derivatives.  I don't know.

bcoconni:
@djlinse 
> In my university days, I would take that static pressure value to a table in the back of a book (Roskam in my case as I studied at Kansas where he taught), scan down the pressure column until I read that value, then scan across to the ISA_altitude corresponding to that pressure. (Of course, interpolating as necessary between table values.)

I finally wrapped my head around this and I am now realizing that we could have coded `CalculatePressureDensity` as a simple bisection on the altitude until `GetStdPressure(altitude)` equals  the current pressure. Equally, `CalculateDensityAltitude` could be made by bisecting the altitude until the ratio `GetStdPressure(altitude)/(Reng*GetStdTemperature(altitude))` equals the current density. This would have mimicked the process you followed when using the table in Roskam's book.

Now we have chosen to compute the pressure altitude and the density altitude with closed form formulas which is fine but forces us to write code to manage the different atmosphere layers and the corner cases where we have to switch from one layer to the other.

bcoconni:
@djlinse 
> I was expecting to see some slight 'kink' in the lines I plotted as the altitude switched from the troposphere to the lower stratosphere, but they look pretty smooth and continuous.

You won't find any kink because the pressure differential is continuous by construction. See formula (5) in the ISA document: it is valid for all the layers below 86km and it defines `d lnP` as an algebraic expression of the temperature T,  the altitude Z and some constants. Since the temperature T is continuous so is the differential of P with respect to the altitude hence the absence of kink.

djlinse:
@bcoconni Yes, clearly a bisection or similar (i.e. faster) root-finding algorithm could be used. Given that the equations are simple enough to invert (at least for pressure), it seems like the closed form solution is much more efficient as you only have to evaluate one "complex" equation, versus evaluating the GetStdPressure() function many times as you bisect to an answer of sufficient accuracy.  To get near the closed-form accuracy would take *many* iteration.  And most the information is already pre-calculated (such as the pressure table breakpoints).  I tend to only think of bisection and the like when it is difficult-to-impossible to invert the function.

I'm also such an old guy that I keep worrying about using up unnecessary computer cycles even when I know that there are plenty to spare on modern computers.

seanmcleod:
@bcoconni regarding kinks. But there is a very definite kink in the temperature at the upper boundary of the troposphere and we switch from formula 33(a) to 33(b), so I was expecting possibly at least a very subtle kink at the boundary.

[Temperature Plot](https://1drv.ms/u/s!AiGCTsLLA9i1gvosflvuIsw7RmRlgA)

I also plotted the density from 0ft to 65,000ft (again getting the data from JSBSim) and there does appear to be a very subtle kink at the troposphere boundary.

[Density Plot](https://1drv.ms/u/s!AiGCTsLLA9i1gvot_9m4xj-ud9r1Lw)

djlinse:
@seanmcleod As @bcoconni said, the pressure has a continuous first derivative by its construction.  Eqn 5 in the ISA defines dP/dZ as a continuous function.  It is only in integrating to a closed form solution that it is split into 33(a) and 33(b) when dT/dZ is zero.  Since temperature has a discontinuous first derivative by definition for the ISA, the kinks are obvious.  With the continuous first derivative, any change in pressure vs altitude curve would be much more subtle (assuming the second derivative isn't continuous).  A quick numerical differentiation of a bunch of data points around the transition altitude might show you the "kink".  As for density, it is proportional to P / T, so it it is a continuously differentiable function divided by a continuous function, thus a visible change in slope at the altitude boundary.

Clearly kinks like this are a way-of-life in the simulation world, as any table look-up has them.  There are two places where I have run into issues with them over the years:

1. Cm(alpha).  It is tempting, at times, just to throw "anything" in this curve, but if there is a serious kink, then you're saying that dCm/d(alpha) is discontinuous.  As that is an important term for classic stability analysis, a discontinuity causes lots of problems...

2. Depending on the robustness of the optimizer used for trim, a discontinuity slope can cause it a lot of confusion as it attempts to numerically calculate a derivative (Jacobian).  I had a problem with a thrust table one time.  Took me a long time to realize why I could easily trim at two nearby points (say 245KCAS and 255KCAS), but not between them.

PR #61, opened 2018-06-12
Update pressure and density altitude calculations

seanmcleod:
To support the 'atmosphere/override/*' properties and to implement @djlinse's suggestion for passing in the already calculated pressure and density values rather than re-computing them.

Next step is to extend the calculations above the troposphere.

### DIFF
--- a/src/models/FGAtmosphere.h
+++ b/src/models/FGAtmosphere.h
@@ -232,9 +232,9 @@ protected:
   /// Calculate the atmosphere for the given altitude.
   void Calculate(double altitude);
 
-  virtual double CalculateDensityAltitude(const double altitude) { return altitude; }
+  virtual double CalculateDensityAltitude(double density, double geoalt) { return geoalt; }
 
-  virtual double CalculatePressureAltitude(const double altitude) { return altitude; }
+  virtual double CalculatePressureAltitude(double pressure, double geoalt) { return geoalt; }
 
   // Converts to Rankine from one of several unit systems.
   virtual double ConvertToRankine(double t, eTemperature unit) const;
--- a/src/models/FGAtmosphere.cpp
+++ b/src/models/FGAtmosphere.cpp
@@ -130,8 +130,8 @@ void FGAtmosphere::Calculate(double altitude)
     Density = node->GetDouble("atmosphere/override/density");
 
   Soundspeed  = sqrt(SHRatio*Reng*(Temperature));
-  PressureAltitude = CalculatePressureAltitude(altitude);
-  DensityAltitude = CalculateDensityAltitude(altitude);
+  PressureAltitude = CalculatePressureAltitude(Pressure, altitude);
+  DensityAltitude = CalculateDensityAltitude(Density, altitude);
 
   Viscosity = Beta * pow(Temperature, 1.5) / (SutherlandConstant + Temperature);
   KinematicViscosity = Viscosity / Density;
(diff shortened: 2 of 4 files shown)


djlinse:
Be daring and change

if (geoalt > TroposphereMaxAltitude)

to

if (pressure < ISA_tropo_boundary_pressure)

whatever that number is, or from the ISA breakpoints themselves.  [Sorry, answering quickly while on a work telecon...]

seanmcleod:
@djlinse don't worry it's coming in the next pull request as the next step ;-)

src/models/FGAtmosphere.h
```diff
@@ -232,9 +232,9 @@ class FGAtmosphere : public FGModel {
   /// Calculate the atmosphere for the given altitude.
   void Calculate(double altitude);
 
-  virtual double CalculateDensityAltitude(const double altitude) { return altitude; }
+  virtual double CalculateDensityAltitude(double density, double geoalt) { return geoalt; }
```
bcoconni: I don't find the name `geoalt` very clear because both geometric and geopotential start with "geo".
And you dropped the `const`-ness of the parameters ?

src/models/FGAtmosphere.h
```diff
@@ -232,9 +232,9 @@ class FGAtmosphere : public FGModel {
...
 
-  virtual double CalculatePressureAltitude(const double altitude) { return altitude; }
+  virtual double CalculatePressureAltitude(double pressure, double geoalt) { return geoalt; }
```
bcoconni: Same as for `CalculateDensityAltitude`:
* `geoalt` not clear
* The `const`-ness of the `double` parameters is dropped ?

seanmcleod:
@bcoconni  will update in a commit in the next pull request.

`geoalt`, not sure what I was thinking, I was so focused on geometric altitude that I forgot it's ambiguous with geopotential.

In terms of documentation I already have Doxygen documentation for the functions in `FGStandardAtmosphere.h`, e.g. although I just noticed in terms of the argument name I need to update it in the Doxygen documentation string.

```c++
  /** Calculates the density altitude given any temperature or pressure bias.
  Currently the formula used is only valid up until the top of the Troposphere,
  11.019km - 36151.8ft.
  @param altitude
  @see
  https://en.wikipedia.org/wiki/Density_altitude
  https://wahiduddin.net/calc/density_altitude.htm
  */
  virtual double CalculateDensityAltitude(double density, double geoalt);
```

I'll add a basic Doxygen comment for the base class versions.

In terms of the `const` argument I was going for consistency. The base class `FGAtmosphere` had them as `const` but the derived class `FGStandardAtmosphere` didn't.

bcoconni:
@seanmcleod 
> In terms of the `const` argument I was going for consistency. The base class `FGAtmosphere` had them as `const` but the derived class `FGStandardAtmosphere` didn't.

Ok. No problem. It was just that we had a debate where I understood that you preferred `const` parameters for these functions.

seanmcleod:
No @djlinse was the one who introduced the `const` parameter, and during one of my commits I removed the `const` on the parameter and marked the method `const`. Which then triggered the discussion.

I don't have a specific preference for `const` on simple value type arguments like `int`, `double` etc.

bcoconni:
> I don't have a specific preference for `const` on simple value type arguments like `int`, `double` etc.

OK. Sorry for the confusion.

PR #65, opened 2018-06-14
Atmosphere updates

seanmcleod:
@djlinse I know I promised to bite the bullet for the next pull request and remove the troposphere check based on geometric altitude and replace it with one based on pressure while adding code to calculate pressure and density altitude all the way to 86km.

However in writing that code I want to use the constants as specified in the ISA document, in particular the standard R* constant rather than the one we currently use. The other issue with not using the standard R* constant is that our current implementations of equations 33(a) and 33(b) in the ISA document isn't an exact match in terms of the constants used.

So what I've done with this pull request is to define the relevant atmosphere constants as per the ISA document but with conversions to British units. So we also end up with a standard R* constant.

I've then also updated our 33(a) and 33(b) equations to use all the constants as per the ISA document.


djlinse:
Looks reasonable to me.  I'd suggest documenting the R*, etc., values briefly in the header file.  Also the conversion constants in the code, even if only in a comment.

PR #68, opened 2018-06-18
Atmosphere use pre-calculated lapserates

seanmcleod:
As per @djlinse's suggestion we now use the pre-calculated set of lapse rates for each atmospheric layer rather then re-calculating it for each sim time step in both `CalculatePressureAltitude()` and `CalculateDensityAltitude()`.

And an additional commit to rename all the member variables that include their type in their name.

### DIFF
--- a/src/models/atmosphere/FGStandardAtmosphere.cpp
+++ b/src/models/atmosphere/FGStandardAtmosphere.cpp
@@ -97,7 +97,7 @@ FGStandardAtmosphere::FGStandardAtmosphere(FGFDMExec* fdmex)
                            << 278385.8268 << 336.5028  // 84.852      86.000
                            << 298556.4304 << 336.5028; //             91.000 - First layer in high altitude regime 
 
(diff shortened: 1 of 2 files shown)


djlinse:
@seanmcleod Looks good to me.

My only suggestion is that that pb isn't a good name here:

> double pb = StdDensityBreakpoints[b];

I'm not sure if you were trying for maybe "p as rho" or something, but pb looks too much like a pressure value to me.  I don't have a great suggestion, other than "something else".  :-/

seanmcleod:
Yep I was going for rho versus the upper case Pb for pressure. Given modern editors etc. I guess I could find a Unicode value for the actual rho symbol :wink:

djlinse:
I'm not sure if I'm ready to read code with full Unicode values...

But I know it wouldn't be a "new" thing.  I recall in gradual school in about 1987 or 88 seeing undergraduates use APL on IBM PCs.  I was happy with my recently acquired C knowledge.  It was cryptic enough.

seanmcleod:
Now we need editors to render MathML and compiler support :wink:

Issue #62, opened 2018-06-13
[Code Quality] Define and apply our own coding style criteria


agodemar:
We should define our set of coding style rules and apply them consistently throughout the sources. This implies a code review, better if with automatic tools.

Something along the lines of [this guide by Google](https://google.github.io/styleguide/cppguide.html) might work.

Ideas?

PR #72, opened 2018-06-20
New behaviour when sending fdm data to output socket. New attribute action="WAIT_SOCKET_REPLY" in <output/> tag

agodemar:
This is a new feature that I have been testing. It works now as expected on my fork and it's time to submit a PR.

The enhancement is about the way JSBSim behaves when used as a client application, connected with an external server application, and sending FDM data on a socket.

**The idea**
The idea comes from [this research article](http://www.mdpi.com/2226-4310/5/2/42) demonstrating how JSBSim can be used to run flight simulations in presence of a gust field pre-calculated with [OpenFOAM](https://openfoam.org/). This is what I do:
-  The chosen aircraft model is flown through a volumetric CFD mesh. Each node in the grid owns three cartesian components of the local wind.
- At each step of the FDM propagation, JSBSim sends the aircraft actual position (lat, lon, agl-altitude) to an external server application on an output TCP socket. 
- *JSBSim blocks time propagation and waits for server reply*.
- The server reads and transforms the received geographic position in a triplet of Cartesian coordinates then accesses the velocity field associated to the mesh and calculates the local wind velocity by interpolation.
- The server sends back on the same socket the wind velocity.
- JSBSim sets the new values of the properties: `atmosphere/wind-north-fps`, `atmosphere/wind-east-fps`, and `atmosphere/wind-down-fps`.
- The FDM propagation continues.

**The enhancement**
We have now a new attribute `action` when we want to configure the `<output/>` tag of `type="SOCKET"`. The new attribute can be set to the value `"WAIT_SOCKET_REPLY"` as follows:
```xml
<!-- socket_oi.xml -->
<output name="localhost" type="SOCKET" port="1138" rate="20" action="WAIT_SOCKET_REPLY" >
  <property> position/lat-gc-deg </property>
  <property> position/long-gc-deg </property>
  <property> position/h-agl-meters </property>
</output>
```
When the attribute `action` is not set, the behaviour of JSBSim is unchanged. When it is set as in the above example, the selected properties are sent with a 20 Hz frequency on the output socket *but*, at each occurrence, the FDM pauses and waits for a reply. *The reply is a string*.
The reply may contain one or more lines and must end with a newline character. Each line should contain only a `set` command. The following is an example of reply:
```
"set fcs/rudder-cmd-norm 1↵
set atmosphere/gust-east-fps 5.0↵
set atmosphere/gust-north-fps 12.0↵
set atmosphere/gust-down-fps 0.0↵"
```
Such a reply is taken care of by the `FGOutputSocket::Print()` function. First, It sends all outbound data on the socket then it calls the function `FGfdmSocket::Receive()` on the current socket object.
```c++
void FGOutputSocket::Print(void)
{
    ...
    socket->Send();
    if (IsWaitSocketReply()) {
        std::string data = socket->Receive(); // get socket transmission if present
        ParseReply(data);
    }
}
```
Finally, the received data string is parsed by the function `FGOutputSocket::ParseReply`. If the string contains well formed `set` commands they are applied and the FDM propagation continues.

The function `FGOutputSocket::ParseReply` is adapted by the code found in function `FGInputSocket::Read`. Note that the class `FGInputSocket` is designed to use JSBSim as a server application accepting commands from external client apps via socket. On the contrary, the present enhancements modifies JSBSim behaviour *when it is used as a client application*. Without this enhancement JSBSim as a client just spits out data on the output socket. With this enhancement JSBSim execution can be modified by an external server's reply, and the incoming `set` commands are generic and are up to the server logic.

**Tests**
The enhancement has been tested with a server application that you can find in [this repository](https://github.com/agodemar/WiReS). In particular, I use a server app that replies each time to JSBSim output with the above example string. This server is implemented in the [MiniWRServer.C](https://github.com/agodemar/WiReS/blob/master/WiReS/MiniWRServer.C) source.

The server is executed assigning a listening port number equal to `1138`. Then JSBSim is launched with the following command line:
```bash
JSBSim --script=scripts/C1723.xml --logdirectivefile=socket_oi.xml
```
And this is what we have:
```
... Loading bla bla bla ...

>>>>>> WAIT_SOCKET_REPLY 
    Properties logged:
      - lat-gc-deg
      - long-gc-deg
      - h-agl-meters
      - h-agl-ft
Creating TCP socket on port 1138
Successfully connected to socket for output ...
[agodemar] socket will be used also for input.
[FGfdmSocket::Receive] reading data on the client's socket...
bytes read = 127
>>> [FGOutputSocket::PrintHeaders] DATA RECEIVE >>>
set fcs/rudder-cmd-norm 1
set atmosphere/gust-east-fps 5.0
set atmosphere/gust-north-fps 12.0
set atmosphere/gust-down-fps 0.0

>>> [FGOutputSocket::PrintHeaders] no parsing.
[FGfdmSocket::Receive] reading data on the client's socket...
bytes read = 127
>>> [FGOutputSocket::Print] DATA RECEIVE >>>
set fcs/rudder-cmd-norm 1
set atmosphere/gust-east-fps 5.0
set atmosphere/gust-north-fps 12.0
set atmosphere/gust-down-fps 0.0

>>> [FGOutputSocket::ParseReply] parsing inbound data >>>
>>> [FGOutputSocket::ParseReply] line: set fcs/rudder-cmd-norm 1
[FGOutputSocket::ParseReply] set successful.
>>> [FGOutputSocket::ParseReply] line: set atmosphere/gust-east-fps 5.0
[FGOutputSocket::ParseReply] set successful.
>>> [FGOutputSocket::ParseReply] line: set atmosphere/gust-north-fps 12.0
[FGOutputSocket::ParseReply] set successful.
>>> [FGOutputSocket::ParseReply] line: set atmosphere/gust-down-fps 0.0
[FGOutputSocket::ParseReply] set successful.

...

[FGfdmSocket::Receive] reading data on the client's socket...
bytes read = 127
>>> [FGOutputSocket::Print] DATA RECEIVE >>>
set fcs/rudder-cmd-norm 1
set atmosphere/gust-east-fps 5.0
set atmosphere/gust-north-fps 12.0
set atmosphere/gust-down-fps 0.0

>>> [FGOutputSocket::ParseReply] parsing inbound data >>>
>>> [FGOutputSocket::ParseReply] line: set fcs/rudder-cmd-norm 1
[FGOutputSocket::ParseReply] set successful.
>>> [FGOutputSocket::ParseReply] line: set atmosphere/gust-east-fps 5.0
[FGOutputSocket::ParseReply] set successful.
>>> [FGOutputSocket::ParseReply] line: set atmosphere/gust-north-fps 12.0
[FGOutputSocket::ParseReply] set successful.
>>> [FGOutputSocket::ParseReply] line: set atmosphere/gust-down-fps 0.0
[FGOutputSocket::ParseReply] set successful.

...

```


agodemar:
`TestInputSocket` fails. Probably because I have changed `FGfdmSocket` constructor's signature from
```c++
FGfdmSocket::FGfdmSocket(const string& address, int port, int protocol)
```
to
```c++
FGfdmSocket::FGfdmSocket(const string& address, int port, int protocol, bool dosocketreceive)
```

agodemar:
@seanmcleod 

> Seems to be inconsistent indentation between the original code and the new code. Tabs instead of spaces? Or just larger number of spaces?

This brings up issue #62 on our coding style. We should have a section in our [Programmer's Manual](https://jsbsim-team.github.io/jsbsim-reference-manual/mypages/programmer-manual) where we define our coding rules.

bcoconni:
@agodemar 
> We should have a section in our Programmer's Manual where we define our coding rules.

The comment from @seanmcleod regarding indentation is not minor. The diffs of this PR are full of lines which are just indentation modifications and this makes your PR hard to review because we have to sort out the real code modifications from the spurious indent mods.

The convention for JSBSim is 2 spaces indentation and no tabs. It has been so for years. I think we can meet that simple convention without a coding style document, can't we ?

bcoconni:
As I said previously, the multitude of indentation modifications makes your PR very hard to review so I could only conduct a general review.

### General philosophy of the PR
First of all, I find the general idea of your PR very interesting and my suggestions are made in the intent of including your code to JSBSim. I would like to make sure that my text is not perceived negatively :anguished: 

Unless I am mistaken you are using the **output** socket to get **input** data. This does not comply with JSBSim general design where input and output are separated and are taking place at different phases of the global loop in `FGFDMExec`. This design is not accidental and there is a good reason why input has been split from output.

The problem is that your code can potentially modify data *before* other `<output>` tags are executed. This could cause such `<output>` elements to report modified data instead of the genuine results from the time step that has just been executed. Of course, this is undesirable.

Each `<output>` element must have the guarantee that it is accessing genuine & unmodified data and furthermore that the data they are exporting **does not depend on the order in which `<output>` elements are executed**. If my comprehension of your PR is correct, it is not meeting this requirement.

I think you can meet your goal by leaving `FGOutputSocket` untouched and adding the new feature WAITING_SOCKET_REPLY to `FGInputSocket` instead. I am hoping that this should save most of the code modifications to `FGfdmSocket` as well.

Such a design modification to your code means that you should have to add an `<output>` element to your model to export data to the OpenFOAM server **and** an `<input>` element to import data from the server. In such a case, the execution of JSBSim would be suspended while executing the `<input>` element.

### Duplication of existing code
Once again, I was unable to review the details of your PR so I might be wrong.

You wrote code to parse the `set` commands. Such code already exists in `FGInputSocket` from line 148 and on (see [the source code here](https://github.com/JSBSim-Team/jsbsim/blob/93f89276968a223aa0bc8e3e34fc9f32f70f1fd3/src/input_output/FGInputSocket.cpp#L148)). You might need to inherit `FGInputSocket` for your particular purpose, if so, try not to duplicate that code.

### Unit conversion
This one is quite minor but we recently added a feature (commits 474ca41 and cfd181b) to prevent the proliferation in the C++ code of properties that express data in different units. Your XML code for output should rather look like
```xml
<output name="localhost" type="SOCKET" port="1138" rate="20" action="WAIT_SOCKET_REPLY"  file="uniconversions.xml">
  <property> position/lat-gc-deg </property>
  <property> position/long-gc-deg </property>
  <property apply="convert-ft-To-m" caption="position/h-agl-meters"> position/h-agl-ft </property>
</output>
```
Note that in the XML code above,
* `<output>` is including the file `unitconversions.xml` that contains most of the unit conversions you need. The file is available in the directory `scripts`
* The `<property>` for the height AGL is using two parameters
  * `apply` to specify the conversion and,
  * `caption` to specify the name of the property after the conversion took place.

This should save all the modifications you made to the files `src/FGPropagate.h` and `src/FGPropagate.Cpp` and reduce the size of the PR consequently.

agodemar:
@bcoconni 

Thanks for your comments. They clarify the rationale behind the design of the two classes `FGOutputSocket` and `FGInputSocket`. 

Once again, I think that the knowledge we produce in our discussions, here or in other mailing lists, can be preserved if we make it clear and reachable through a usable documentation. I'm thinking to the casual users or to those who my want to add code and contribute: at the moment they would spend nights reading through the issues before grasping the sense of a design choice.

The same applies to the coding conventions. 
> The convention for JSBSim is 2 spaces indentation and no tabs. It has been so for years. I think we can meet that simple convention without a coding style document, can't we ?

Of course, we can. But I also kno that something that looks evident to the experienced programmer, e.g. using a two space indentation instead of tabs, might not be so to the novice. I think that's far better having it stated clearly than left unsaid. Once again, documentation implies inclusion.

Getting to the original motivation of my PR,

> I think you can meet your goal by leaving FGOutputSocket untouched and adding the new feature WAITING_SOCKET_REPLY to FGInputSocket instead. I am hoping that this should save most of the code modifications to FGfdmSocket as well.

> Such a design modification to your code means that you should have to add an <output> element to your model to export data to the OpenFOAM server and an <input> element to import data from the server. In such a case, the execution of JSBSim would be suspended while executing the <input> element.

practically, you mean that I could have accomplished my goal - i.e. having JSBSim that talks to the OpenFOAM server - *with no additional code* but only with `<output/>` and `<input/>` settings?

seanmcleod:
@agodemar I was also thinking along the lines of whether you couldn't rather put the socket read blocking logic in an input element socket. But I wasn't too sure how the initial potential handshaking would work so I hadn't mentioned it yet. But reading @bcoconni's comments and thinking about it a bit more this is my general thinking.

So as @bcoconni said leave the output socket code as is. So JSBSim will send the relevant aircraft state to your OpenFOAM server without blocking on each sim loop iteration.

But you specify a new **blocking** input socket so that on each sim loop iteration JSBSim will block while it waits for socket input, whereas at the moment the FGInputSocket calls `socket->Receive()` as a non-blocking call so if there is no data available it simply returns and the sim loop continues.

Now the issue I had with this is the potential initial handshaking. For example at startup JSBSim will call the blocking input socket and block until data arrives, but the OpenFOAM server needs an initial aircraft state (lat, lon, alt) before it can send any data which it only gets once the output elements are processed at the end of the sim loop, but now we're stuck waiting for input on the very first iteration of the sim loop. So the system is deadlocked.

But thinking about it a bit more what about the following. On the initial iteration of the sim loop the input socket doesn't have an incoming connection yet, it's still waiting for a connection. So it doesn't block in this case it simply returns.

Then at the end of the first iteration of the sim loop the output socket sends the initial aircraft state to the OpenFOAM server. The OpenFOAM server on receiving this initial incoming connection now knows the IP address of JSBSim and so it then creates a connection to the input socket (with some configured port number) that is waiting for connections. And it sends back it's first set of wind velocities.

So now on the next iteration of the sim loop the input socket will be connected and it will issue a blocking read to retrieve the OpenFOAM set commands for the wind velocities etc. and will process them and return allowing the sim loop to now continue in lock-step with the OpenFOAM server.

agodemar:
> But thinking about it a bit more what about the following. On the initial iteration of the sim loop the input socket doesn't have an incoming connection yet, it's still waiting for a connection. So it doesn't block in this case it simply returns.

> Then at the end of the first iteration of the sim loop the output socket sends the initial aircraft state to the OpenFOAM server. The OpenFOAM server on receiving this initial incoming connection now knows the IP address of JSBSim and so it then creates a connection to the input socket (with some configured port number) that is waiting for connections. And it sends back it's first set of wind velocities.

>So now on the next iteration of the sim loop the input socket will be connected and it will issue a blocking read to retrieve the OpenFOAM set commands for the wind velocities etc. and will process them and return allowing the sim loop to now continue in lock-step with the OpenFOAM server.

This seems a good approach to me. Comes out that the extra logic has to be managed on the server side, which is good because it doesn't involve modifications to JSBSim code. There's the need of a convention on the port of the input socket, e.g. the server has to assume that if inbound data come on port 1138 than JSBSim is listening on port, say, 1139.

seanmcleod:
You don't necessarily need a convention in terms of port numbers. If you have a config file for your external server, e.g. OpenFOAM then you could specify the required port numbers in the config file as long as they match the port number specified in the JSBSim input and output sockets.

seanmcleod:
Glancing at the socket code a bit it looks like currently for the input socket case the socket passed to the `accept()` API is set to non-blocking, so it won't block waiting for a connection. The socket passed to the `recv()` API is also set to non-blocking after a successful `accept()`.

For the OpenFOAM case/example it doesn't sound like it is critical that the very first sim loop iteration needs data from the OpenFOAM server. From what I remember from the paper a while ago the aircraft model is trimmed in air with no turbulence first anyway etc.

But there may be some cases where you are integrating JSBSim with an external server supplying data like wind velocities etc. where it is critical that you have the input data from the external server on the very first simulation loop iteration. To handle that case we could potentially have multiple flags for the input socket which specify blocking versus non-blocking for the `accept()` and the `recv()` calls independently.

So for the OpenFOAM example you would leave the `accept()` as non-blocking but set `recv()` blocking but for another server where the initial data is critical you would set both `accept()` and `recv()` to blocking. In which case JSBSim would start up but block until the external server connected and sent the first packet.

agodemar:
I am happy with this discussion so far. I close the pull request and consider the approach suggested by @bcoconni and @seanmcleod. I've opened a new issue #74 where we can continue on the same topic.
Thanks

Issue #74, opened 2018-06-22
Output data via socket to an external server and wait for reply on an input socket


agodemar:
Here I collect the main points of a previous discussion in a closed PR #72 .

**The idea**
The idea comes from [this research article](http://www.mdpi.com/2226-4310/5/2/42) demonstrating how JSBSim can be used to run flight simulations in presence of a gust field pre-calculated with [OpenFOAM](https://openfoam.org/). This is what I do:
-  The chosen aircraft model is flown through a volumetric CFD mesh. Each node in the grid owns three cartesian components of the local wind.
- At each step of the FDM propagation, JSBSim sends the aircraft actual position (lat, lon, agl-altitude) to an external server application on an output TCP socket. 
- *JSBSim blocks time propagation and waits for server reply*.
- The server reads and transforms the received geographic position in a triplet of Cartesian coordinates then accesses the velocity field associated to the mesh and calculates the local wind velocity by interpolation.
- The server sends back *on the same socket* (in that case) the wind velocity.
- JSBSim sets the new values of the properties: `atmosphere/wind-north-fps`, `atmosphere/wind-east-fps`, and `atmosphere/wind-down-fps`.
- The FDM propagation continues.

**Configuration**
The O/I can be configures via the `<output/>` and `<input/>` tags as follows:
```xml
<!-- socket_oi.xml -->
<output name="localhost" type="SOCKET" port="1138" rate="20" file="uniconversions.xml">
  <property> position/lat-gc-deg </property>
  <property> position/long-gc-deg </property>
  <property apply="convert-ft-To-m" caption="position/h-agl-meters"> position/h-agl-ft </property>
</output>
<input type="SOCKET" port="1139" />
```

**Strategy**

The server application (OpenFOAM server) should do most of the work. 

On the initial iteration of the sim loop the JSBSim input socket doesn't have an incoming connection yet, it's still waiting for a connection. So it doesn't block in this case it simply returns.

Then at the end of the first iteration of the sim loop the output socket sends the initial aircraft state to the OpenFOAM server. The OpenFOAM server on receiving this initial incoming connection now knows the IP address of JSBSim and so it then creates a connection to the input socket (with some configured port number, i.e. 1139) that is waiting for connections. And it sends back it's first set of wind velocities.

[...]

PR #81, opened 2018-06-30
Implementation of feature request #74

bcoconni:
This is a proposal for the new feature discussed in #74.


bcoconni:
@seanmcleod 
> What has been the general rule in terms of case sensitivity for attribute values?

I am not aware of any rule regarding case sensitivity but I see what you mean: I just made the attribute "action" case insensitive.

agodemar:
@bcoconni looks good to me. Thank you for the effort. I'll test this feature with my OpenFOAM server code later in the upcoming week.

PR #92, opened 2018-07-14
Filters: fix incorrect coefficient rates caused by execrate.

Zaretto:
The coefficients are initialized during construction - however at this point dT isn't correct - so we need to set these up again the first time they are used.

### DIFF
--- a/src/models/flight_control/FGFilter.cpp
+++ b/src/models/flight_control/FGFilter.cpp
@@ -190,6 +190,9 @@ bool FGFilter::Run(void)
 {
   if (Initialize) {
 
+    // need to do this now as dt will not be correct during initialization
+    CalculateDynamicFilters();
+
     PreviousOutput2 = PreviousInput2 = PreviousOutput1 = PreviousInput1 = Output = Input;
     Initialize = false;
 


bcoconni:
> The coefficients are initialized during construction - however at this point dT isn't correct

What do you mean by *dT isn't correct* ? 

`FGFilter::FGFilter` is called by `FGFCS::Load` which is called by `FGFDMExec::LoadModel` which is called by `FGScript::LoadScript` which is called after `dT` is set. So I am intrigued by your statement.

Could you please provide a test case where the initialization is failing ?

Zaretto:
dT hasn't been set according to execrate during the component creation. It sets dT in the base constructor using the FCS - which of course doesn't have the execrate because execrate sits on the channel.

image

Possibly a better fix would be to set dT in FGFCS::Load based on the channel rather than the FCS; I did consider doing this but decided against it because it means more changes to the code (as the components don't have the channel and would therefore need this to be passed into the constructors). I thought this was a higher risk. However ff you want me to do this I'm happy to have a look at doing it this way, or any other way that you prefer.

image

bcoconni:
@Zaretto 
> dT hasn't been set according to execrate during the component creation.

OK got it.

> It sets dT in the base constructor using the FCS - which of course doesn't have the execrate because execrate sits on the channel.

OK but the execrate is set before `FGFilter::FGFilter` is called so why don't you pass `Rate` to the constructor `FGFilter::FGFilter` ?

Zaretto:
After some though I have an alternative solution that is more elegant; to set the rate variable in the system to the execrate in line 528 of  FGFCS::Load and set it back to the original value at the end of the while loop (line 590)

image

image

and remove the changes in FGFilter.cpp

Zaretto:
I've just tested your patch and it works fine thanks. Do you want me to commit this to my branch - or would you prefer just to do it yourself?

abassign:
Hi, I read the thread of discussion, I find it very interesting and I hope that the change can be as soon as propagated in Flightgear that I think uses a recompiled version of JSBSim. The reason for a quick modification of this code to solve the bug I had also reported to Zaretto because I'm working on the FIAT G91-R1B which is replacing almost all the NASAL code with JSBSim code, and much of my code, for a greater efficiency, use Execrate> 1.

bcoconni:
@Zaretto 
> I've just tested your patch and it works fine thanks. Do you want me to commit this to my branch - or would you prefer just to do it yourself?

I just pushed the patch to JSBSim (commit fc3b870)

bcoconni:
And it is now pushed to FlightGear as well.

PR #93, opened 2018-07-15
Improved configuration error detection in FGSwitch w.r.t. late bound …

andgi:
Currently FGSwitch does not evaluate output values until they are selected and skips evaluation of conditions after the one selected. Hence, some configuration faults might not be detected until far into a run. This changes makes sure that all conditions and output values can be evaluated during the first time step so that problems do not get hidden until later times. This makes FGSwitch behave more like the other components w.r.t. late bound properties.

If there are no objections I'll add this to master tomorrow.
(Don't merge this request unless it is a fast-forward merge, thank you. I prefer a linear history.)

### DIFF
--- a/aircraft/p51d/Systems/electrical.xml
+++ b/aircraft/p51d/Systems/electrical.xml
@@ -10,6 +10,7 @@
     <property value="0">systems/right-flor-gain</property>
 
     <!-- Additional properties which are defined separately in FlightGear -->
+    <property value="0">/controls/gear/gear-safe-light-check</property>
     <property value="0">/controls/gear/gear-unsafe-light-check</property>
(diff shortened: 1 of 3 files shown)


bcoconni:
@andgi 
Your patch breaks 4 tests

11 - TestInitialConditions (Failed)
12 - CheckScripts (Failed)
13 - CheckAircrafts (Failed)
27 - TestKinematic (Failed)

Could you please fix that before pushing your patch to master ?

Thanks.

andgi:
@bcoconni 
I don't have python set up here but I'll look at the tests. E.g. the p51d was missing a declaration of an interface property. Let's see if it changes anything.

bcoconni:
@andgi 
> Let's see if it changes anything.

It does. All tests pass now.

PR #98, opened 2018-07-16
Run Python tests on AppVeyor

seanmcleod:
@bcoconni @agodemar now that all the Python tests run under Windows I'm enabling them via `ctest` on AppVeyor. Don't merge this pull request until I've confirmed that the tests are running successfully on AppVeyor.


seanmcleod:
@bcoconni @agodemar you can go ahead and merge this now. The Python tests all run successfully on AppVeyor.

Fyi, enabling `-V verbose` output with `ctest` in order to see why the tests were failing results in a build run that hits the 60min mark without finishing all the tests which is the time limit we have for AppVeyor builds.

Running without the verbose option results in a build and test run of about 5 mins.

bcoconni:
@seanmcleod 
Good job !

PR #100, opened 2018-07-19
added conversion M/S to KTS

WillemEerland:
I wanted to initialise the speed in "M/S" and found out I couldn't because there was no conversion to KTS. There is now.

### DIFF
--- a/src/input_output/FGXMLElement.cpp
+++ b/src/input_output/FGXMLElement.cpp
@@ -124,6 +124,7 @@ Element::Element(const string& nm)
     convert["KTS"]["FT/SEC"] = 1.68781;
     convert["FT/SEC"]["KTS"] = 1.0/convert["KTS"]["FT/SEC"];
     convert["M/S"]["FT/S"] = 3.2808399;
+    convert["M/S"]["KTS"] = convert["M/S"]["FT/S"]/convert["KTS"]["FT/SEC"];
     convert["M/SEC"]["FT/SEC"] = 3.2808399;
     convert["FT/S"]["M/S"] = 1.0/convert["M/S"]["FT/S"];
     convert["M/SEC"]["FT/SEC"] = 3.2808399;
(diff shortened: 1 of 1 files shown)


bcoconni:
@andgi 
The abbreviation for the unit of time *second* in the SI units is *"s"* and *m/s* is a SI unit.

Indeed historically the abbreviation "SEC" has been added to JSBSim (to be used with British/Imperial units ?) However *"m/sec"* has always looked weird to me but if you feel that it is a miss, feel free to push a commit.

andgi:
@bcoconni Well, yes, I grew up with SI so that is no news to me :)

Ok, so you propose that we use "S" with SI units and "SEC" with British/Imperial units? There is as far as I know no such consistency currently. "SEC" is used in almost all units in FGXMLElement, I think "S" only occurs in "M/S".

bcoconni:
@andgi 
> Well, yes, I grew up with SI so that is no news to me :)

Sorry there was no offense intended.

>  so you propose that we use "S" with SI units and "SEC" with British/Imperial units?

No I don't. All I am saying is that *I guess* that "SEC" was introduced by people using British units. I am not proposing to modify anything. All I meant was that I can live without "M/SEC" but if you feel that it should be added, go ahead !

> As far as I see "SEC" already occurs in SI-based units RAD/SEC, N/M/SEC, N/M2/SEC2, M/SEC, M4*SEC/KG, N/SEC in FGXMLElement.

There is plenty to say regarding unit naming in JSBSim:
* Square is not used consistently : sometimes it is materialized by the number 2 appended to the unit (see your examples above), sometimes by "sq" prepended to the unit such as `metrics/Sw-sqft`
* @djlinse also filled a [bug issue on SF](https://sourceforge.net/p/jsbsim/bugs/73/) a long time ago about SFC units
* The same abbreviation "LBS" is used for pounds (weight) and pounds-force (force) so is LBS/(HR*LBS) homogeneous to HR^-1 ?
* etc.

But well, we have to live with our legacy :smiley:

andgi:
@bcoconni No offense taken. :) And I know we have to live with our legacy. It is just that now between "M/S" and "M/SEC" which both were there already, "M/S" can be used where JSBSim needs to convert it to knot while "M/SEC" cannot.

PR #104, opened 2018-07-24
Allow any trim type to be specified in initialization file

seanmcleod:
Currently the `<trim>` element in an initialization file can only be used to specify `0 - no trim` or `1 - ground trim`. This pull requests adds support for all the currently supported trim types to be specified and maintains backwards compatibility.

`- trim (0 for no trim, 1 for ground trim, 'Longitudinal', 'Full', 'Ground', 'Pullup', 'Custom', 'Turn') `

### DIFF
--- a/src/JSBSim.cpp
+++ b/src/JSBSim.cpp
@@ -440,7 +440,8 @@ int real_main(int argc, char* argv[])
 
     if (!FDMExec->GetPropertyManager()->GetNode(CommandLineProperties[i])) {
       cerr << endl << "  No property by the name " << CommandLineProperties[i] << endl;
-      goto quit;
+      delete FDMExec;
+      exit(-1);
     } else {
       FDMExec->SetPropertyValue(CommandLineProperties[i], CommandLinePropertyValues[i]);
     }
@@ -550,9 +551,6 @@ int real_main(int argc, char* argv[])
     }
 
   }
-
-  
-quit:
 
   // PRINT ENDING CLOCK TIME
   time(&tod);
--- a/src/JSBSim.cpp
+++ b/src/JSBSim.cpp
@@ -454,10 +454,20 @@ int real_main(int argc, char* argv[])
   // Dump the simulation state (position, orientation, etc.)
   FDMExec->GetPropagate()->DumpState();
   
-  if (FDMExec->GetIC()->NeedTrim()) {
-    trimmer = new JSBSim::FGTrim( FDMExec );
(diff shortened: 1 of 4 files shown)


src/initialization/FGInitialCondition.cpp
```diff
@@ -979,6 +980,27 @@ bool FGInitialCondition::LoadLatitude(Element* position_el)
 
 //******************************************************************************
 
+void FGInitialCondition::SetTrimRequest(std::string& trim)
```
bcoconni: I think you need to use `const` for this code to compile under Travis CI
```c++
void FGInitialCondition::SetTrimRequest(const std::string& trim)
```

src/JSBSim.cpp
```diff
@@ -454,10 +454,20 @@ int real_main(int argc, char* argv[])
...
-  if (FDMExec->GetIC()->NeedTrim()) {
-    trimmer = new JSBSim::FGTrim( FDMExec );
+  // Perform trim if requested via the initialization file
+  JSBSim::TrimMode icTrimRequested = (JSBSim::TrimMode)FDMExec->GetIC()->TrimRequested();
+  if (icTrimRequested != JSBSim::TrimMode::tNone) {
+    trimmer = new JSBSim::FGTrim( FDMExec, icTrimRequested );
     try {
+      // If PullUp requested then we need to pass in target Nlf
+      if (icTrimRequested == JSBSim::TrimMode::tPullup)
+        trimmer->SetTargetNlf(FDMExec->GetIC()->GetTargetNlfIC());
```
bcoconni: Why isn't that code in lines 462-464 included in `FGTrim::FGTrim` ?
I am pretty sure scripts will ignore the Nlf target otherwise.

seanmcleod:
I've fixed the gcc compiler error regarding the `string` argument and fixed the merge conflict. But I see now via Travis that there is another gcc only (compiles fine on AppVeyor) error now regarding a goto and local variable initialization etc. which I'll look into later today.

I'll also respond later to the question regarding `targetNlf`.

src/JSBSim.cpp
seanmcleod: I've moved the code to `FGTrim::FGTrim` which means the following will now also work which never worked previously in addition to pullup trim working from a `<trim>` element in the initialization file.

```xml
<!-- Pullup trim -->
<set name="simulation/do_simple_trim" value="4" />
```

Issue #63, opened 2018-06-14
Using the latest JSBSim in the ArduPilot SITL simulation


WillemEerland:
The documentation for setting up a SITL simulation using ArduPilot (ArduPlane) and JSBSim, found at 

http://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#setting-up-sitl-on-linux

points to the following version of JSBSim:

www.github.com/tridge/jsbsim (updated last Feb 2016) 

and only has minor modifications (includes a stepping function) to

www.github.com/arktools/jsbsim (updated last Dec 2012)

Any advice on how to use the latest JSBSim to use with ArduPilot? Any advice is appreciated.

edit: and a quick link to the file that "communicates" with JSBSim

https://github.com/ArduPilot/ardupilot/blob/master/libraries/SITL/SIM_JSBSim.cpp

seanmcleod:
Glancing through the `SIM_JSBSim.cpp` I don't see anything obvious that isn't supported by the current JSBSim code. It looks like it's pretty much using the FlightGear protocol via a local socket.

So unless someone else replies with specific experience using the latest JSBSim source with Ardupilot I'd suggest simply building the latest JSBSim source and trying it out and if something doesn't work then post what the issue is.

WillemEerland:
Thanks for your reply. At least now I know the functionality should work.

I've been making quite a bit of progress today, and will post more when I run into specific issues, as you suggested. I was merely hoping someone had already done this already and could save me the effort by sharing their experiences.

Once I've got it working I'll write up a tutorial to go in the documentation.

WillemEerland:
And fixed... The problem seems to be that the latest JSBSim only recognises "UDP" in the XML files, when it sees "udp" it switches to TCP.

Anyways, I modified the ArduPilot code slightly:
- removing the check for "(ArduPilot)" in the version
- changing "udp" to "UDP"

https://github.com/WillemEerland/ardupilot/tree/sitl

Perhaps we can assume that if someone writes "udp" in the XML, they want a UDP port, and not TCP?

seanmcleod:
Good to know that no changes were required in the latest version of JSBSim. We should make the string comparison for protocol type case-insensitive in JSBSim.

Does that mean the older fork of JSBSim that was used for SITL with Ardupilot was case-insentitive with regards to the protocol type?

WillemEerland:
Either that or the default changed from UDP to TCP

[...]

PR #109, opened 2018-08-10
Uniform reply behaviour of FGInputSocket::Read

agodemar:
This PR comes from issue #63, where part of the discussion concerns the behaviour of JSBSim when used as an input server responding synchronously to an external application connected via socket.

In that discussion I refer to an example of use where [a client app](https://github.com/agodemar/WiReS/blob/master/WiReS/MinimalClientServer.C) 
- sends first a `get` command to JSBSim,
- JSBSim replies sending a string back to the client with the value of the asked property,
- the client then sends a `set` command, which succeeds
- finally the socket is closed by the client and JSBSim continues its propagation algorithm until the final time.

All commands sent as string by the example client application to JSBSim through the input socket are digested by JSBSim's `FGInputSocket::Read`. This function parses the string checking the validity of the command, eg.:
- `get <property>`
- `set <property> <value>`
- `hold`
- `resume`
- `iterate <int>`
- `quit`
- `info`
- `help`

and finally calls `Reply` that sends back through the socket a string message to the client. 

In the current version of `FGInputSocket::Read`, for some commands like a successful `set <property>` JSBSim replies with an *empty string* followed by the prompt `JSBSim> `. No new line character. See [FGInputSocket.cpp](https://github.com/JSBSim-Team/jsbsim/blob/master/src/input_output/FGInputSocket.cpp), line 173. The same behaviour is implemented for `hold`, `resume`, `iterate`, and `quit`.

In the present PR I've changed the reply messages in case of successful `set`, `hold`, `resume`, `iterate`, and `quit` prepending to the `"JSBSim> "` prompt a confirmation message terminated by a new line character `\n`. For instance:
```
Set successful
JSBSim>
```
or
```
Holding
JSBSim>
```
or
```
Resuming
JSBSim>
```
and so on.

This harmless change

- gives a uniform reply behaviour, i.e. always a one-line message terminated by a `\n` plus the prompt string `JSBSim> `
- facilitates external client applications when parsing the replies, e.g. they could read from the socket until the `\n` character and throw away the remaining ones.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,8 @@ ipch/
 ## Jupyter
 #################
 .ipynb_checkpoints
+
+#################
+## Mac OSX
+#################
+.DS_Store
--- a/aircraft/p51d/Systems/electrical.xml
+++ b/aircraft/p51d/Systems/electrical.xml
@@ -10,6 +10,7 @@
     <property value="0">systems/right-flor-gain</property>
 
     <!-- Additional properties which are defined separately in FlightGear -->
+    <property value="0">/controls/gear/gear-safe-light-check</property>
     <property value="0">/controls/gear/gear-unsafe-light-check</property>
     <property value="0">/controls/lighting/left-flor-pos</property>
     <property value="0">/controls/lighting/right-flor-pos</property>
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,13 +16,13 @@ version: 1.0.{build}
 build_script:
 - set PATH=c:\python36-x64;c:\python36-x64\Scripts;%PATH%
 - ren c:\python27 python27-hide
-- pip install Cython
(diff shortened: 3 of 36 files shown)


seanmcleod:
Given the original interactive design with the `JSBSim>` prompt etc. I guess there shouldn't be any issue adding success messages as part of the reply.

For automated programs I guess currently they could read the reply looking for the `JSBSim>` prompt string as opposed to specifically looking for a `\n` terminated string?

If you merge this in make sure to squash all the commits into a single commit so that some of the earlier commits that temporarily added debugging code like the example below don't end up in the main JSBSim branch.

```
cout << "FGInputSocket::Read - Sim Frame - " << this->FDMExec->GetFrame() << " Data: " << data << endl; 
```

PR #115, opened 2018-09-08
Increase the precision of the constants used in CAS - Mach conversions.

seanmcleod:


seanmcleod:
Busy fixing the FGJSBBaseTest1 test failure.

Issue #31, opened 2018-05-20
`TestInputSocket` randomly fails


bcoconni:
There has been several occurrences of a failure of `TestInputSocket` during the builds made by Travis CI (builds [58.1](https://travis-ci.org/JSBSim-Team/jsbsim/jobs/380426980), [59.1](https://travis-ci.org/JSBSim-Team/jsbsim/jobs/380444668), [60.1](https://travis-ci.org/JSBSim-Team/jsbsim/jobs/380449707)  have failed but [62.1](https://travis-ci.org/JSBSim-Team/jsbsim/jobs/381169894) and [57.x](https://travis-ci.org/JSBSim-Team/jsbsim/builds/380382870) have succeeded).
* The source code tested by `TestInputSocket` has not been modified by the commits that triggered the builds mentioned above so these builds were expected to succeed.
* The failures only occurred when the test was run by Python 2.7 (the same tests succeeded when they were run by Python 3.6)

My experience while running `TestInputSocket` is that these failures are due to a race condition. The test is using 2 threads:
1. The first thread runs a `JSBSim` instance
2. The second thread runs a `telnet` instance

The test consists in checking that the 2 threads are capable of exchanging information via a socket.

Despite the usage of 2 threads, this tests is not really based on multithreading due to [Python GIL](https://wiki.python.org/moin/GlobalInterpreterLock). Despite the GIL, some thread synchronization incantations are needed to get the expected result. This synchronization seems to randomly fail with Python 2.7.

bcoconni:
I updated the Travis CI build process to upload the tests log to **gist** if one the tests fails (commits 041238e and e64975d). The files will be uploaded under my [GitHub account](https://gist.github.com/bcoconni).
Alternatively `gist` outputs the web address of the gist to which the logs have been copied.

bcoconni:
Woohoo ! The honeypot caught the evil bug :-) `TestInputSocket` failed on a build of my personal fork and the test log is available at [that gist](https://gist.github.com/679fa3780d8c6f0958ead8fc85c61438).

bcoconni:
Since [Python 2.7 has reached its official end of life since Jan 01, 2020](https://www.python.org/doc/sunset-python-2), and this bug is limited to Python 2.7 and we drop support of Python 2.7 (see issue #29) then the current issue is closed.

PR #121, opened 2018-09-27
FGTURBINE: StartN1 / StartN2 values patch

legoboyvdlp:
Hi,
I have a small patch allowing aircraft developers to change the value to which the starter rotates the fan before ignition starts, allowing them to change it from 5% / 25% giving more control over startup time on turbine engines. I will create another patch soon to add a similar n1spinup value between this value and idlen1.

I did not test it, as I am on Windows, and compilation is a bit more difficult... it _should_ work, but might not!
There is also no sanity check, so you could have some crazy number: e.g. it trying to spin up to 100% if you set startn1 to that in the xml. But I think JSBSIM is "garbage in, garbage out"? 😄  

 
Edit: I have now added a value n1start / n2start which allows the user to control the time between the values I added previously and idle n1 / n2.

So, overall I added:
 
  ```
<startnX> {number} </startnX> added
  <nXspinup> {number} </nXspinup> adjusted: time between 0% and startnX%
  <nXstart> {number} </nXstart> added, time between startnX% and idlenX%
```
where X is equal to 1 or 2.

I realise the similarity between startnX and nXstart might confuse some users, any suggestions as to what to change the names to?

Also, you might like to squash these commits - I am using the web interface until I can start compiling.

Thanks!

### DIFF
--- a/src/models/propulsion/FGTurbine.cpp
+++ b/src/models/propulsion/FGTurbine.cpp
@@ -70,7 +70,7 @@ FGTurbine::FGTurbine(FGFDMExec* exec, Element *el, int engine_number, struct Inp
   Augmented = AugMethod = Injected = 0;
   BypassRatio = BleedDemand = 0.0;
   IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = 0;
-  N1_spinup = 1.0; N2_spinup = 3.0;
+  N1_spinup = 1.0; N2_spinup = 3.0; StartN1 = 5.21; StartN2 = 25.18;
   InjectionTime = 30.0;
   InjectionTimer = InjWaterNorm = 0.0;
   EPR = 1.0;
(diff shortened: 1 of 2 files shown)


legoboyvdlp:
You might wish to squash those commits into one - sorry, I forgot to add both files when I first pushed, so I made a second commit later.

legoboyvdlp:
pushed second part, and first post is updated

legoboyvdlp:
@seanmcleod @bcoconni

bcoconni:
I'm mostly fine with your PR. However I would appreciate if you would name the variables `start_Nx` and `Nx_start` with something more explicit. Because honestly I am not sure how someone could tell the difference between the variables just on the basis of the order of the words `start` and `Nx`.

The same remark applies to the XML tags `<startnX>` and `<nXstart>`

legoboyvdlp:
@bcoconni if I may ask to confirm, what I do to test in FlightGear now I _finally_ have that compile setup after a week or more of trying is to replace fgturbine in flightgear's source directory and recompile, correct?

I might need to add some corrections, since the state logic seems to check N2 is greater than 15 which could theoretically break the engine in various cases.

Thanks,

legoboyvdlp:
@bcoconni
I wonder if you could test this pull request or tell me how to do it? 

Is passing the FDM files directly from Flight gear to standalone JSBSIM possible?

legoboyvdlp:
Hello,
Could anyone please check the failed build?

And this is ready for merging if you all are happy with it.

bcoconni:
@legoboyvdlp 
Sorry, I've been busy those last weeks.
> Is passing the FDM files directly from Flight gear to standalone JSBSIM possible?

It really depends on the model at stake. Some of FlightGear models use Nasal scripts and cannot be loaded with JSBSim standalone without prior clean up.

> Could anyone please check the failed build ?

This has been fixed: the test passes now. The problem was due to the known bug #31.

> And this is ready for merging if you all are happy with it.

Done.

legoboyvdlp:
Many thanks!

PR #129, opened 2018-11-01
PythonJSBSim link error with VS 2017

bcoconni:
Patch to fix the issue reported by Alan Teeder on the mailing list.
> I am getting the following error when tying to build the test suite on Windows with Cmake 12.3 ,  VS 2017  and Python 3.6
>
> Any ideas ?
>
> Alan
```bash 
1>------ Build started: Project: PythonJSBSim, Configuration: Release x64 ------
1>Compiling Cython CXX source for jsbsim...
1>Building Custom Rule C:/FlightGear/jsbsim/python/CMakeLists.txt
1>CMake does not need to re-run because C:/FlightGear/jsbsim/build64/python/CMakeFiles/generate.stamp is up-to-date.
1>Building Python modules...
1>running build_ext
1>building 'jsbsim' extension
1>CUSTOMBUILD : error : command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Tools\\MSVC\\14.15.26726\\bin\\HostX86\\x64\\link.exe' failed with exit status 1181
1>Done building project "PythonJSBSim.vcxproj" -- FAILED.
```


bcoconni:
The suggestion I made to Alan consisted in a manual modification to the build file. But the purpose is to keep a completely automated process so I am investigating how this can be achieved.

So far I am able to reproduce the error message reported to Alan. My problem is to extract the $<CONFIG> value.

seanmcleod:
Ah, I re-read your email to Alan and I see you had him hardcode the path to `src/release`. So it seems in the ninja case the .lib file is created in `src` but not in the msbuild case. So ideally we need a value in `setup.py.in` that works for both ninja and msbuild.

bcoconni:
> So ideally we need a value in `setup.py.in` that works for both ninja and msbuild.

Yes that's the plan.

PR #139, opened 2018-12-21
Add ability to control spindown rate factor

legoboyvdlp:
Allows user to adjust spindown time by using n1spindown and n2spindown, which will be used as such:

```
N1rate = N1 / n1spindown
N2rate = N2 / n1spindown
```

If this pull request is accepted, I would like this to be in 2019.1 with the other changes if at all possible?


bcoconni:
Committed. Thanks for the patch.
>  I would like this to be in 2019.1 with the other changes if at all possible?

I think so. The JSBSim code is pushed regularly to FlightGear. Drop us a note if that does not happen soon enough.

bcoconni:
JSBSim code has just been pushed to FlightGear. You should now be able to use your code.

legoboyvdlp:
great job, thanks!

Issue #128, opened 2018-10-31
Problem for use of the actuator in rotary displacement when it exceeds 360 ° ...


abassign:
The use of the actuator allows obtaining a very realistic result for linear or rotary displacement systems but limited within a certain range of values. While it does not work properly for multi-turn rotary systems. For example this wind speed display is made by simulating an electromechanical component. The effect between 0 and 9 is perfect, but between 9 and 0 or between 0 and 9 (backward rotation) the actuator finds its equilibrium state after a complete rotation. The problem is that Actuator does not provide the ability to manage these particular types of devices. The same problem is present for the lap-compass needle which, when it exceeds 360 °, makes a further rotation to find the equilibrium point.

This is an example:
https://youtu.be/SnTTpJdyFlw
```xml
<actuator name="digit3AW-inc">
     <input>systems/gauges/PHI/wind/digit3A</input>
     <lag>0.0</lag>
     <rate_limit>10</rate_limit>
     <bias>0.002</bias>
     <deadband_width>0.01</deadband_width>
     <hysteresis_width>0.01</hysteresis_width>
     <output>systems/gauges/PHI/wind/digit3AW</output>
</actuator>
```

I thought a lot about this problem, and I sensed that in fact the "actuator" function works very well for linear displacements or with values within a certain range, but it can not operate correctly on cyclic values, like those presented here. I solved the problem with a couple of complex fcs-functions, but I'm certainly not satisfied. I think it would be much better to insert a control that, in the case of actuators that have to control systems that perform an indeterminate number of turns, correct the discontinuity both forward and backward, avoiding the problem shown here in the movie.

My solution is in the list displayed: https://i.postimg.cc/RFyGmqYz/Screenshot-20181031-222326.png

That is, I transform the problem by constructing a sort of counter, then first calculating the variation and then adding the value currently used by the actuator function, if it exceeds the maximum value (in the example 9.9 for pisitive variation) I set the displayed value to zero. On the other hand, if the new value is less than zero, I set the displayed value to 9.9.

Obviously there are many different ways, but it would not be difficult to allow the actuator to manage this type of device.

I'd like to know if this change can be made.

Thank you

mike402:
My solution would be to filter sine and cosine separately, but it is way too bulky and perhaps has hidden gotchas.

For now I have to leave 360-degree needles in instruments and such unfiltered. Same problem with FlightGear propertyrule-based filtering.

abassign:
[...]

PR #154, opened 2019-01-11
FGLinearActuator

abassign:
Hi, I have made a linear actuator to convert a rotation to several revolutions in a linear displacement. I discussed this device [here](https://forum.flightgear.org/viewtopic.php?f=49&t=35221) in the "FlightGear ‹ Development ‹ Aircraft ‹ Flight dynamics model" and in [GIT this topic](https://github.com/JSBSim-Team/jsbsim/issues/128#issuecomment-452220985). I installed a demo version on my local PC and I made a [movie](https://youtu.be/9WE-ElasXrY) showing its features. I then discussed this object with @agodemar that advised me to proceed this way. For the work on the FIAT G91R1B sdi FGFS this type of actuator is very important and has the advantage of allowing very easily to obtain a correct rotation or movement of the instruments that perform multiple rotations, such as the compass, the radio-compass, the counters of miles with geneva gear etc. This object also introduces a set and reset command that makes it similar to a memory flip-flop with analogue output, an excellent system for creating a simple event counter. This object should not be confused with the currently present Actuator, as it is used before this device. The device has a system to discriminate the direction of rotation which can however be forced with an explicit direction of rotation. In this case, if the rotation direction does not coincide with the explicit direction, the device does not modify the output making it similar to an escapement. The instructions are in the file: models/flight_control/FGLinearActuator.h Obviously, as you have invited me to do @agodemar , I am willing to write a user manual with examples of this device and to maintain it. I will write a user manual with various examples of this device and to maintain it.

Thanks for your attention, best regards

### DIFF
--- a/src/models/flight_control/FGLinearActuator.cpp
+++ b/src/models/flight_control/FGLinearActuator.cpp
@@ -1,252 +1,286 @@
-/*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
- * 
- * Module:       FGLinearActuator.cpp
- * Author:       Adriano Bassignana
- * Date started: 2019-01-03
- * 
- * ------------- Copyright (C) 2000 -------------
- * 
- * This program is free software; you can redistribute it and/or modify it under
- * the terms of the GNU Lesser General Public License as published by the Free Software
- * Foundation; either version 2 of the License, or (at your option) any later
- * version.
- * 
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
- * FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
- * details.
- * 
- * You should have received a copy of the GNU Lesser General Public License along with
- * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
- * Place - Suite 330, Boston, MA  02111-1307, USA.
- * 
- * Further information about the GNU Lesser General Public License can also be found on
- * the world wide web at http://www.gnu.org.
- * 
- * FUNCTIONAL DESCRIPTION
- * --------------------------------------------------------------------------------
- * 
- * HISTORY
- * --------------------------------------------------------------------------------
- * 
- * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
- * COMMENTS, REFERENCES,  and NOTES
- * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
- * 
- * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
- * INCLUDES
- * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
-
-#include "FGLinearActuator.h"
-#include "input_output/FGXMLElement.h"
-#include <iostream>
-
-using namespace std;
-
-namespace JSBSim {
-    
-    /*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-     * CLASS IMPLEMENTATION
-     * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
-    
-    FGLinear_Actuator::FGLinear_Actuator(FGFCS* fcs, Element* element) : FGFCSComponent(fcs, element)
-    {
-        initialized = 1;
-        versus = 0.0;
-        isSetProperty = false;
-        isResetProperty = false;
-        direction = 0;
-        countSpin = 0;
-        input_prec = 0.0;
-        module = 1.0;
-        rate = 0.3;
-        minRate = -0.3;
-        maxRate = 0.3;
-        lag = 0.0;
-        lagPrevius = 0.0;
-        gain = 1.0;
+ /* %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+  * 
+  * Module:       FGLinearActuator.cpp
+  * Author:       Adriano Bassignana
+  * Date started: 2019-01-03
+  * 
+  * ------------- Copyright (C) 2000 -------------
+  * 
+  * This program is free software; you can redistribute it and/or modify it under
+  * the terms of the GNU Lesser General Public License as published by the Free Software
+  * Foundation; either version 2 of the License, or (at your option) any later
+  * version.
+  * 
+  * This program is distributed in the hope that it will be useful, but WITHOUT
+  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  * FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+  * details.
+  * 
+  * You should have received a copy of the GNU Lesser General Public License along with
+  * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+  * Place - Suite 330, Boston, MA  02111-1307, USA.
+  * 
+  * Further information about the GNU Lesser General Public License can also be found on
+  * the world wide web at http://www.gnu.org.
+  * 
+  * FUNCTIONAL DESCRIPTION
+  * --------------------------------------------------------------------------------
+  * 
+  * HISTORY
+  * --------------------------------------------------------------------------------
+  * 
+  * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+  * COMMENTS, REFERENCES,  and NOTES
+  * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+  * 
+  * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+  * INCLUDES
+  * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
+ 
+ #include "FGLinearActuator.h"
+ #include "input_output/FGXMLElement.h"
+ #include "math/FGParameterValue.h"
+ #include <iostream>
+ 
+ #ifndef DEBUG
+ #define DEBUG 0
+ #endif
+ 
+ using namespace std;
+ 
+ namespace JSBSim {
+     
+     /* %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+      * CLASS IMPLEMENTATION
+      * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
+     
+     FGLinear_Actuator::FGLinear_Actuator(FGFCS* fcs, Element* element) : FGFCSComponent(fcs, element)
+     {
+         
+         if (element->FindElement("set")) {
+             isSetProperty = true;
+             setProperty = PropertyManager->CreatePropertyObject<double>(element->FindElementValue("set"));
+         }
+         if (element->FindElement("reset")) {
+             isResetProperty = true;
+             resetProperty = PropertyManager->CreatePropertyObject<double>(element->FindElementValue("reset"));
+         }
+         
+         ptrVersus = nullptr;
+         if (element->FindElement("versus")) {
+             string property_string = element->FindElementValue("versus");
+             ptrVersus = new FGParameterValue(property_string, PropertyManager);
+             double setVersus = ptrVersus->GetValue();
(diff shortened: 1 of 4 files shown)


src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+            std::cerr << "FGLinear_Actuator Cicle_zero " << InputNodes[0]->GetNameWithSign() << std::endl;
+            return true;
+        }
+        
+        Output = 0.0;
+        
+        set = 1.0;
+        reset = 0.0;
+        
+        if (isSetProperty) set = setProperty;
```
bcoconni: You can save the use of the boolean `isSetProperty`. Just make sure that `setProperty` is initialized to `null` during construction.
```suggestion
        if (setProperty) set = setProperty;
```

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        if (isResetProperty) reset = resetProperty;
```
bcoconni: Same as above, you can remove the boolean `isResetProperty`
```suggestion
        if (resetProperty) reset = resetProperty;
```

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        // for this Lag filter
+        double input = Output;
+        
+        if (lag > 0.0) {
+            if (lag == lagPrevius) {
+                Output = ca * (input + previousLagInput) + previousLagOutput * cb;
+            } else {
+                double denom = 2.00 + dt*lag;
+                ca = dt * lag / denom;
+                cb = (2.00 - dt * lag) / denom;
```
bcoconni: Constants `ca` and `cb` should be initialized in the constructor.
The variable `dt` can be considered constant.

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+namespace JSBSim {
+    
+    /*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+     * CLASS IMPLEMENTATION
+     * %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
+    
+    FGLinear_Actuator::FGLinear_Actuator(FGFCS* fcs, Element* element) : FGFCSComponent(fcs, element)
+    {
+        initialized = 1;
```
bcoconni: Use a `bool` for the variable `initialized`

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        versus = 0.0;
+        isSetProperty = false;
+        isResetProperty = false;
```
bcoconni: Remove the 2 variables `isSetProperty` and `isResetProperty` we don't need them (see comments on lines 118 and 119)

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+    //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    
+    void FGLinear_Actuator::Lag(void)
+    {
+        // "Output" on the right side of the "=" is the current frame input
+        // for this Lag filter
+        double input = Output;
+        
+        if (lag > 0.0) {
+            if (lag == lagPrevius) {
```
bcoconni: `lag` is assigned once in the constructor and never modified again so this test is always true.

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        isResetProperty = false;
+        direction = 0;
+        countSpin = 0;
+        input_prec = 0.0;
+        module = 1.0;
+        rate = 0.3;
+        minRate = -0.3;
+        maxRate = 0.3;
+        lag = 0.0;
+        lagPrevius = 0.0;
```
bcoconni: Since `lag` is not modified by the code, why are using the variable `lagPrevius` ?
See also the comment on line 182.

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        gain = 1.0;
+        
+        if (element->FindElement("set")) {
+            setProperty = PropertyManager->CreatePropertyObject<double>(element->FindElementValue("set"));
+            isSetProperty = true;
```
bcoconni: Remove `isSetProperty`
```suggestion
```

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        }
+        if (element->FindElement("reset")) {
+            resetProperty = PropertyManager->CreatePropertyObject<double>(element->FindElementValue("reset"));
+            isResetProperty = true;
```
bcoconni: Remove `isResetProperty`
```suggestion
```

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        if (element->FindElement("versus")) versus = element->FindElementValueAsNumber("versus");
+        if (element->FindElement("lag")) lag = element->FindElementValueAsNumber("lag");
+        if (element->FindElement("module")) module = element->FindElementValueAsNumber("module");
+        if (element->FindElement("rate")) {
+            rate = element->FindElementValueAsNumber("rate");
+            if (rate <= 0 or rate > 1.0) rate = 0.5;
```
bcoconni: I think you should output a warning here. The `<rate>` value is modified and the user must be warned that the code has modified the value he/she submitted.

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+    
+    //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    
+    bool FGLinear_Actuator::Run(void )
+    {
+        Input = InputNodes[0]->getDoubleValue();
+        if (initialized) {
+            input_prec = Input;
+            initialized = 0;
+            std::cerr << "FGLinear_Actuator Cicle_zero " << InputNodes[0]->GetNameWithSign() << std::endl;
```
bcoconni: The stream `std::cerr` is for error messages. I think you should rather use `std::cout` here.
Also you can remove the `std::` namespace since you declared the namespace in `using namespace std;`
```suggestion
            cout << "FGLinear_Actuator Circle_zero " << InputNodes[0]->GetNameWithSign() << endl;
```

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+            return true;
```
bcoconni: Why are you returning at this point ?

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+        reset = 0.0;
+        
+        if (isSetProperty) set = setProperty;
+        if (isResetProperty) reset = resetProperty;
+            
+        if (set > 0 ) {
+            
+            double input_delta = Input - input_prec;
+            
+            if ((versus < 0.0 and input_delta > 0.0) or (versus  > 0.0 and input_delta < 0.0)) {
```
bcoconni: Since the point is that `versus` and `input_delta` should be opposite in sign you can just check their product
```suggestion
            if (versus * input_delta < 0.0) {
```

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+    
+    void FGLinear_Actuator::Lag(void)
+    {
+        // "Output" on the right side of the "=" is the current frame input
+        // for this Lag filter
+        double input = Output;
+        
+        if (lag > 0.0) {
+            if (lag == lagPrevius) {
+                Output = ca * (input + previousLagInput) + previousLagOutput * cb;
```
bcoconni: I am pretty sure that `previousLagInput` and `input_prec` are redundant and contain the same value at any time.

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+                    }
+                }
+                
+            }
+        } else {
+            Input = input_prec;
+        }
+        
+        if (reset > 0) {
+            Input = 0.0;
```
bcoconni: Should `input_prec` be reset to `0.0` as well ?

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+                        direction = -1;
+                        //cout << "FGLinear_Actuator::Run" << InputNodes[0]->GetNameWithSign() << " direction - " << direction << endl;
+                    }
+                }
+                
+                if (direction != 0) {
+                    if (input_delta >= (module * maxRate) or input_delta <= (module * minRate)) {
+                        if (direction > 0) {
+                            countSpin++;
+                            cout << "FGLinear_Actuator::Run" << InputNodes[0]->GetNameWithSign() << " countSpin++ " << countSpin << " maxRate " << maxRate << " minRate " << minRate << endl;
```
bcoconni: Don't issue a message at each time step since that will flood the terminal/console of the user.
```suggestion
```

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+                            direction = 0;
+                        } else if (direction < 0) {
+                            countSpin--;
+                            cout << "FGLinear_Actuator::Run" << InputNodes[0]->GetNameWithSign() << " countSpin-- " << countSpin << " maxRate " << maxRate << " minRate " << minRate << endl;
```
bcoconni: Same as above don't issue messages at each time step.
```suggestion
```

src/models/flight_control/FGLinearActuator.h
```diff
@@ -0,0 +1,229 @@
...
+        /// Destructor
+        ~FGLinear_Actuator();
+        
+        /// The execution method for this FCS component.
+        bool Run(void);
+        
+    private:
+        
+        simgear::PropertyObject<double> setProperty;
+        bool isSetProperty;
```
bcoconni: Remove `isSetProperty`
```suggestion
```

src/models/flight_control/FGLinearActuator.h
```diff
@@ -0,0 +1,229 @@
...
+        FGLinear_Actuator(FGFCS* fcs, Element* element);
+        /// Destructor
+        ~FGLinear_Actuator();
+        
+        /// The execution method for this FCS component.
+        bool Run(void);
+        
+    private:
+        
+        simgear::PropertyObject<double> setProperty;
```
bcoconni: You should rather use a property node here
```suggestion
        FGPropertyNode_ptr setProperty;
```

src/models/flight_control/FGLinearActuator.h
```diff
@@ -0,0 +1,229 @@
...
+        bool isSetProperty;
+        double set;
+        simgear::PropertyObject<double> resetProperty;
```
bcoconni: You should use a property node here
```suggestion
        FGPropertyNode_ptr resetProperty;
```

src/models/flight_control/FGLinearActuator.h
```diff
@@ -0,0 +1,229 @@
...
+        bool isResetProperty;
```
bcoconni: Remove `isResetProperty`
```suggestion
```

src/models/flight_control/FGLinearActuator.h
```diff
@@ -0,0 +1,229 @@
...
+        double reset;
+        
+        int initialized;
```
bcoconni: This variable should be boolean
```suggestion
        bool initialized;
```

src/models/flight_control/FGLinearActuator.h
```diff
@@ -0,0 +1,229 @@
...
+        int direction;
+        int countSpin;
+        
+        double lagProperty;
+        double lagPrevius;
```
bcoconni: Is not  `lagPrevius` redundant with `input_prec` ?

src/models/flight_control/FGLinearActuator.cpp
```diff
@@ -0,0 +1,252 @@
...
+            if (input_delta != 0.0) {
+                if (input_delta <= (module * maxRate) and input_delta >= (module * minRate)) {
+                    if (input_delta > 0.0) {
+                        direction = 1;
+                        //cout << "FGLinear_Actuator::Run" << InputNodes[0]->GetNameWithSign() << " direction + " << direction << endl;
+                    } else if (input_delta < 0.0) {
+                        direction = -1;
+                        //cout << "FGLinear_Actuator::Run" << InputNodes[0]->GetNameWithSign() << " direction - " << direction << endl;
+                    }
+                }
```
bcoconni: Since `input_delta` is non zero, `direction` can be initialized with the sign of `input_delta` instead of the block between lines 131-138 : 
```c++
        direction = sign(input_delta);
```
You might need to include `FGJSBBase.h` for the `sign` function to be recognized by the compiler.

bcoconni:
@abassign 
The build of the your code fails because you did not add the files `FGLinearActuator.h` and `FGLinearActuator.cpp` to `models/flight_control/CMakeLists.txt`.

Could you also please fix that ?

src/models/flight_control/FGLinearActuator.cpp
abassign: The question:
        if (isSetProperty) set = setProperty;
        if (setProperty) set = setProperty;
Currently setProperty (and resetProperty) are binary values, but they are likely to become analog values as they can be an easy way to trigger a fuzzy feature for this object. For this I used a Boolean variable to indicate to the system that the object operates in Set-Reset mode and eventually use the Set-Reset values as analog data to set the initial output value (resetProperty) or setProperty to store the value only when a certain threshold is exceeded. The isSetProperty (or isResetProperty) method makes these properties similar to a three-state logic in which the static configuration is provided with set and reset if these are connected (isSet ... true) or a logic without a set and reset if they are not connected (isSet or isReset false).

abassign: Your statement is excellent and I very willingly follow the advice. So your correction is fine for me and I will be careful to avoid inserting the non-operative code in pull request, even if I plan to use it in the future.

src/models/flight_control/FGLinearActuator.cpp
abassign: Ok

src/models/flight_control/FGLinearActuator.cpp
abassign: Ok

src/models/flight_control/FGLinearActuator.cpp
abassign: ok

src/models/flight_control/FGLinearActuator.cpp
abassign: ok

bcoconni:
[For the record] This PR has been dropped in favor PR #160.

PR #157, opened 2019-01-13
Fix for the Python wheel package installation failure.

bcoconni:
:warning: Work In Progress. DO NOT MERGE :warning: 
For some reason Windows seem unable to locate the package that we just installed.
```bash
pip install python\dist\JSBSim-1.0.0.dev1-cp37-cp37m-win_amd64.whl
Processing c:\projects\jsbsim\build\python\dist\jsbsim-1.0.0.dev1-cp37-cp37m-win_amd64.whl
Requirement already satisfied: cython>=0.25 in c:\python37-x64\lib\site-packages (from JSBSim==1.0.0.dev1) (0.29.2)
Requirement already satisfied: setuptools>=18.0 in c:\python37-x64\lib\site-packages (from JSBSim==1.0.0.dev1) (39.0.1)
Requirement already satisfied: numpy in c:\python37-x64\lib\site-packages (from JSBSim==1.0.0.dev1) (1.15.4)
Installing collected packages: JSBSim
Successfully installed JSBSim-1.0.0.dev1
python -c "import jsbsim;fdm=jsbsim.FGFDMExec('.', None)"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'jsbsim'
Command exited with code 1
```
The purpose of this PR is to fix the problem.


bcoconni:
> How come the wheel has a requirement on cython, that's only needed for building the python module right? I wouldn't expect a wheel user needing cython.
Same goes for setuptools.
And lastly numpy is only required for running our python tests right? So also not really a requirement of the wheel?

Indeed that is the whole point of the PR :smile: 

`setuptools` and `cython` are normally only used for the source distribution but here `setup.py` fails to include the module to the wheel package for a reason that I'm still looking for.

bcoconni:
> And lastly numpy is only required for running our python tests right? So also not really a requirement of the wheel?

Actually `numpy` is needed because the Python module transforms `FGColumnVector3` to `numpy.array` and `FGMatrix33` to `numpy.matrix`.

PR #158, opened 2019-01-19
Abassign patch 1

abassign:
The testing of the module did not give me any problems, it was compiled directly for the development version of FGFS. Compared to the previous version, it has been introduced the possibility to vary the limitation in the direction of rotation (versus) and has been optimized. A bias has been introduced (which can be fixed or linked to a parameter) useful for managing tools that use offset commands. Honestly the name bias I used it as typical of an electronic control system, but for a linear actuator I think it is better to use the name "offset". In this phase of development it seems useful to me that others give advice in order to avoid the use of names not coherent with the function. There is also a hysteresis parameter (currently fixed) which allows to obtain an operation in steps typical of the step-motor or the escapement of the clocks. Something of the kind is also present in the classic actuator (which is often placed after this type of actuator), but this value operates on the input data and not in the output data. The hysteresis module allows the module to operate only if an input is large enough to exceed the hysteresis value. The use of hysteresis can lead to a considerable reduction in workload due to this module. An example is the use of the module to manage the gyrocompasses that have microvariations that can be blocked by the hysteresis without ever losing the consistency of the system. I think that another code of GSBSim could be reviewed in this logic avoiding an excessive use of the Execrete parameter.
For the output there is always the lag and the clip, theoretically these two parameters could be not inserted and left to a possible subsequent actuator module, but it is often convenient to have it and more compact the code. The code currently contains the bebug prints that can be activated / deactivated during the compilation of the code itself, will be removed in the final version, however some of these prints take place only in the case of effective mofification of the output.

### DIFF
--- a/src/models/FGFCS.cpp
+++ b/src/models/FGFCS.cpp
@@ -57,6 +57,7 @@ INCLUDES
 #include "models/flight_control/FGWaypoint.h"
 #include "models/flight_control/FGAngles.h"
 #include "models/flight_control/FGDistributor.h"
+#include "models/flight_control/FGLinearActuator.h"
 
 #include "FGFCSChannel.h"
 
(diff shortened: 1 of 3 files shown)


bcoconni:
[For the record] This PR has been dropped by its author in favor of PR #160.

PR #165, opened 2019-02-06
Res. err. uninitialized variables for Lag opt.

abassign:
I detected an error when the lag is activated, it may happen that the value of the lag initially assumes an absurd value. The reason is that the previousLagInput and previousLagOutput variables were not initialized. This change, which I tested on FGFS aircraft G91R1B, corrects this error.

### DIFF
--- a/src/models/flight_control/FGLinearActuator.cpp
+++ b/src/models/flight_control/FGLinearActuator.cpp
@@ -56,10 +56,6 @@ namespace JSBSim {
     FGLinearActuator::FGLinearActuator(FGFCS* fcs, Element* element) : FGFCSComponent(fcs, element)
     {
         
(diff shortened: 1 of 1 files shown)


abassign:
The error is quite serious and so I hope it can be solved as soon as possible. I also ask if you can upgrade JSBSim for FGFS experimental version so that my other friends of the Discord group can experiment the FGFS G91R1B with the FGLinearActuator function. Thank you :)

bcoconni:
@abassign 
>  I also ask if you can upgrade JSBSim for FGFS experimental version so that my other friends of the Discord group can experiment the FGFS G91R1B with the FGLinearActuator function.

Done. The feature including the current PR is now available in FlightGear `next` branch and will be part of the 2018.4 release.

Issue #181, opened 2019-04-13
Regression: switch statement no longer allows use of <and> and <or>


Zaretto:
Probably as a result of the change to support nested tests in the switch statement the use of the and/or elements no longer works.

This example switch condition the test value="1" never becomes true even when all of the 
conditions are satisfied.

```
       <switch name="systems/cadc/manuv-active">
         <default value="0.0"/>

         <test value="0">
           <or>
             fcs/flap-cmd-norm gt 0.3714285714285714
             aero/alpha-deg lt systems/cadc/manuv-flapslat-retraction-aoa
             velocities/mach gt 
systems/cadc/manuv-flap-partially-extended-max-mn
             gear/gear-cmd-norm ne 0
           </or>
           systems/cadc/manuv-active ne 0
         </test>


         <test value="1">
           <or>
             <and>
               fcs/flap-cmd-norm le 0.3714285714285714
               aero/alpha-deg gt systems/cadc/manuv-flapslat-extension-aoa
               velocities/mach le 
systems/cadc/manuv-flap-fully-extended-max-mn
               gear/gear-cmd-norm eq 0
               fcs/wing-sweep-cmd le 0.7352941176470588
             </and>
             systems/cadc/manuv-active ne 0
           </or>
         </test>
       </switch>

```
The above switch statement has been tested and is working on 2018.3 and a 2018.4 nightly prior to the change.

If I change the logic to use embedded <test logic="OR|AND"> then the 
switch works; however I suspect that this may well break a few existing models and retaining the ability of JSBSim to continue to provide backwards compatibility is important.

bcoconni:
I am afraid the keywords/tags `<and>` and `<or>` have never been supported by JSBSim.

I dug JSBSim code history and never found a mention of these 2 keywords being read and interpreted as a replacement for `<test logic="{AND|OR}">`.

However, JSBSim have always had a relaxed interpretation of the tags within `<switch>` and `<condition>` statements and that allows tags `<and>` to be interpreted as you meant them but it is merely a coincidence.

Actually **any** tag within a `<test>` statement is interpreted as `<test logic="AND">`: you could have used `<smuglob>` instead of `<and>` and JSBSim would have interpreted `<smuglob>` as `<test logic="AND">` nonetheless.

JSBSim only checks for the `logic` attribute so unless you specify `<or logic="OR">` the tag `<or>` will be interpreted as `<test logic="AND">`.

[...]

PR #183, opened 2019-04-21
Implementation of option 4 from issue #181

bcoconni:
Following the problem raised by @Zaretto in issue #181, this PR adds the tags `<and>` and `<or>` and issue a warning message when the latter is used.

This is a proposal implementation while waiting for the discussion on FlightGear mailing list to come to a conclusion.

### DIFF
--- a/src/math/FGCondition.cpp
+++ b/src/math/FGCondition.cpp
@@ -60,7 +60,16 @@ FGCondition::FGCondition(Element* element, FGPropertyManager* PropertyManager)
   InitializeConditionals();
 
   string logic = element->GetAttributeValue("logic");
+  const string& elName = element->GetName();
+
   if (!logic.empty()) {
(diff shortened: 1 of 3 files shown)


bcoconni:
This pull request is also available as a [merge request to the FlightGear code base](https://sourceforge.net/p/flightgear/flightgear/merge-requests/158). If the aforementioned merge request is pulled to FlightGear `next` branch then so the current PR should.

bcoconni:
The issue #181 has just been closed. This PR should normally not be needed. However I suggest to wait for full scale testing after the release of FlightGear 2019.1 which should happen sooner or later.

bcoconni:
The [merge request 158 to the FlightGear code base](https://sourceforge.net/p/flightgear/flightgear/merge-requests/158) has been dropped so this PR can be rejected and closed.

Issue #186, opened 2019-04-28
Transmit simulation time via FG UDP interface


simfinite:
Hey all,

over at ArduPilot/ArduPlane [we are having some trouble](https://github.com/ArduPilot/ardupilot/issues/10612) to reliably synchronize our autopilot simulation application to a JSBSim instance using the FG UDP interface. We would benefit a lot from having the JSBSim's internal simulation time transmitted via the interface along with the FDM state vector.

I noticed a member `cur_time` in the `FGNetFDM` struct that is being sent over this interface. This member is currently documented to hold the unix time but it is unused and set to a 'dummy' value of `1234567890`.  I would like to propose using this existing member to transmit the simulation time. The required changes [be seen in this commit](https://github.com/simfinite/jsbsim/commit/f91283583fabac2b271ce0f84e6484164c37e15a).

If you agree to make this change, I will prepare a pull-request with the required changes in FGNetFDM.hxx and FGOutputFG.cpp. If not, then we can discuss other possible ways of having the simulation time transmitted.

bcoconni:
> I noticed a member cur_time in the FGNetFDM struct that is being sent over this interface. This member is currently documented to hold the unix time but it is unused and set to a 'dummy' value of 1234567890. I would like to propose using this existing member to transmit the simulation time.

Makes sense to me.
> If you agree to make this change, I will prepare a pull-request with the required changes in `FGNetFDM.hxx` and `FGOutputFG.cpp`. 

I guess you mean `net_fdm.hxx` instead of `FGNetFDM.hxx` ? Anyway, I don't see any change on either file in the commit you pointed to. Be careful about the fact that the content of `net_fdm.hxx` is managed by the FlightGear project and that we cannot make any alteration to it before synchronizing with them.

simfinite:
> I guess you mean net_fdm.hxx

Yes. The documenting comment for the member `cur_time` should be changed, see the pull request for my proposal. Since it's "just a comment", I would hope we can go forward with it, but I don't know the procedure.

bcoconni:
Your PR has just been merged.
Thanks for your contribution.

simfinite:
Thank you for the review and help with this.

PR #187, opened 2019-04-28
PR for issue #186: Transmit simulation time via FG UDP interface

simfinite:
Resolves feature request https://github.com/JSBSim-Team/jsbsim/issues/186

Please check whether it's ok to modify the documenting comment in net_fdm.hxx which is managed by the FlightGear project.

### DIFF
--- a/src/input_output/FGOutputFG.cpp
+++ b/src/input_output/FGOutputFG.cpp
@@ -243,7 +243,7 @@ void FGOutputFG::SocketDataFill(FGNetFDM* net)
   }
 
   // Environment
-  net->cur_time    = (long int)1234567890;    // Friday, Feb 13, 2009, 23:31:30 UTC (not processed by FGFS anyway)
+  net->cur_time    = static_cast<uint32_t>(FDMExec->GetSimTime()*1e6);   // used for simulation time sync in ArduPlane (not processed by FGFS)
   net->warp        = 0;                       // offset in seconds to unix time
   net->visibility  = 25000.0;                 // visibility in meters (for env. effects)
 
--- a/src/input_output/FGOutputFG.cpp
+++ b/src/input_output/FGOutputFG.cpp
@@ -44,6 +44,7 @@ INCLUDES
 
 #include "FGOutputFG.h"
 #include "FGFDMExec.h"
+#include "FGXMLElement.h"
 #include "models/FGAerodynamics.h"
 #include "models/FGAuxiliary.h"
 #include "models/FGPropulsion.h"
@@ -119,7 +120,7 @@ CLASS IMPLEMENTATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
 FGOutputFG::FGOutputFG(FGFDMExec* fdmex) :
-  FGOutputSocket(fdmex)
+  FGOutputSocket(fdmex), outputOptions{false, 1e6}
 {
   memset(&fgSockBuf, 0x0, sizeof(fgSockBuf));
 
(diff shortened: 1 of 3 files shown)


bcoconni:
Thanks for the PR @simfinite .

Just a couple of comments:
1. The feature that you are submitting should be optional with the current behavior being the default (i.e. always send 1234567890). The fact that JSBSim does not use this parameter does not mean that FlightGear does not try to interpret it. I am concerned about the effect that this would have when using JSBSim as an external FDM to FlightGear.
2. Hard coded constants should be avoided. Please replace the `1E6` value by a parameter that can be modified by the user.

simfinite:
Fair enough. I am assuming this should be implemented in terms of properties that can be acquired via the FGPropertyManager? 
If so, then I would appreciate some hints on how to implement this since I haven't worked with the property manager before. Is there a good example somewhere in the code on what it should look like? How are new properties defined? Where in the property tree should the new ones be placed?

bcoconni:
You do not need to use properties but rather get information from the XML input files.

Let's say that the feature you are implementing will be requested with a syntax such that
```xml
<output name="localhost" type="FLIGHTGEAR" protocol="UDP" port="5550" rate="60">
  <time type="simulation" factor="1E+6"/>
</output>
```

So you need to check that there is an XML element `<time>` and that it has 2 attributes:
- `type` which should be set to `simulation` to indicate that the simulation time will be sent
- `factor` which provides the constant by which the time will be multiplied.

For that you need to add the inherited method `FGOutputFG::Load` that will set the boolean `useSimTime` and the double `factor` according to what is specified in the XML input:
```c++
FGOutputFG::Load(Element* el)
{
  // Check if there is a <time> element
  Element time_el = el->FindElement("time");
  if (time_el) {

    //Check if the attribute "type" is specified
    if (time_el->HasAttribute("type") && time_el->GetAttributeValue("type") == "simulation") {
      useSimTime = true;

      // Check if the attribute "factor" is specified
      if (time_el->HasAttribute("factor"))
        time_factor = time_el->GetAttributeValueAsNumber("factor");
    }
  }
}
```
Note that this (untested) code implies that `useSimTime` is initialized to `false` in the constructor and `factor` is initialized to `1E+6` in the constructor as well.

The code in `FGOutputFG::SocketDataFill` would then be something like
```c++
  if (useSimTime)
    net->cur_time = static_cast<uint32_t>(FDMExec->GetSimTime() * factor);
 else
    net->cur_time = (long int)1234567890;
```

This is just an example. Feel free to code the thing to your own taste.

simfinite:
Thanks for the detailed example, I think I can take it from here.. just need to find a bit of free time.

simfinite:
Ok, I finally got around to implementing this. I tested it with an accordingly modified ArduPlane SITL and it works as expected. We can now have a properly reliable lock-step mechanism to sync with JSBSim. If not activated via the tag `<time type="simulation">`, JSBSim exhibits the old behavior of sending 1234567890 as a dummy value.

src/input_output/FGOutputFG.cpp
```diff
@@ -146,6 +147,35 @@ FGOutputFG::FGOutputFG(FGFDMExec* fdmex) :
...
+      outputOptions.useSimTime = true;
+    }
+
+    // Check if the attribute "resolution" is specified and set to a valid value
+    if (time_el->HasAttribute("resolution")) {
+      if (time_el->GetAttributeValueAsNumber("resolution") <= 1 &&
+          time_el->GetAttributeValueAsNumber("resolution") >= 1e-9) {
+        outputOptions.timeFactor = 1./time_el->GetAttributeValueAsNumber("resolution");
+      } else {
+        return false;
```
bcoconni: Ideally, an error message should be issued when such a condition is failed.

simfinite: Makes sense. How about:
```suggestion
      } else {
        cerr << "Not a valid time resolution (valid range is [1e-9, 1]): " << time_el->GetAttributeValueAsNumber("resolution") << endl;
        return false;
      }  
```

Issue #184, opened 2019-04-23
Approximating mean sea level / geoid radius with WGS-84 spheroid/ellipsoid


simfinite:
Hey all,
I am using JSBSim together with the ArduPlane framework to simulate fixed-wing drones. Looking at JSBSim CSV logs, I noticed that the ECEF position coordinates were way off from what I expected. The aircraft was initialized at ~9000m height above the WGS-84 ellipsoid (GPS altitude) when initialized with altitude AGL set to '0'. 

Looking around the code, I found out that this is because the default mean sea level radius (in FGDefaultGroundCallback) is initialized to the major semi-axis (equatorial earth radius). This constant value is used as the default reference for altitude AGL and terrain elevation everywhere.. the earth's a sphere again.

Now, I realize that there may be other implementations of FGGroundCallback, possibly with complex geoid and terrain models. However, for my use-case, all I really want is to be able to place the aircraft on an airfield, take-off and produce some reasonable log files. The WGS-84 spheroid is a good reference for that, especially since GPS altitude values can be found everywhere.

Does anyone agree that the WGS-84 spheroid/ellipsoid would be a much better approximation of the geoid than a simple sphere? Any reasons against it? Anything I may have missed/misunderstood?

I would be happy to implement this, if we can agree that it would make sense and wouldn't break anything...

simfinite:
After some further digging into the code, I realized that the spherical earth (or spherical mean sea level) assumption is quite deeply baked into the current implementation (e.g. with concepts such as a "geocentric terrain level radius" which is useful only under this assumption). I played around a bit to get a deeper understanding of what would have to be done to have both altitude and terrain level defined as a height above an ellipsoid instead of a radius of a sphere. I am currently stuck on getting the trim routine to work with my changes.
Anyway, before going further, I would appreciate any feedback on whether this has been adressed before and if any other parts of JSBSim / FlightGear rely on the spherical mean sea level assumptions.. Also, I would like to know how this is dealt with from the FlightGear side, where one can travel around a non-spherical earth right?

bcoconni:
Hi @simfinite 
The Earth mean radius is indeed used in several places in JSBSim for different purposes (off the top of my head: gravity computations, standard atmosphere model and altitude/location computations). This certainly needs some love to make sure that:
1. The mean radius value is the same everywhere which I am pretty sure is not the case at the moment.
2. The mean radius is only used where it makes sense which is not guaranteed at the moment.

[...]

PR #190, opened 2019-05-11
Add geodetic altitude to CSV log

simfinite:
Hi,
I would like to have the geodetic altitude added to the CSV log. Besides my personal use cases, this may be helpful for future work on issue #184. Also, it'll simply make the log more complete.

I noticed, that the altitude AGL / distance above ground is currently being logged twice. I removed one of the entries in the second commit, because I think there shouldn't be any redundant data in the output file. If this shall be kept for backwards-compatibility, just omit the 2nd commit.


simfinite:
Just realized I wasn't running all of the tests locally.. I can reproduce this now and will look into it.

simfinite:
Ok, here is a new attempt. The failure of test case `RunCheckCases` was due to the reference log file `check_cases/orbit/logged_data/BallOut.csv` missing the new column. In order not to change the existing reference data (there are minor numerical differences when comparing to logs of the current version), I produced a new log, but only merged the new column into the existing file. Since this was a bit tricky to accomplish, here is the code I used:

```
new = pandas.read_csv("BallOut.new.csv", index_col="Time", float_precision='round_trip')
old = pandas.read_csv("BallOut.old.csv", index_col="Time", float_precision='round_trip')
merged = old.copy()
merged["Altitude Geodetic (ft)"] = new["Altitude Geodetic (ft)"]
merged.to_csv("BallOut.csv", float_format="%.16g")
```

It's a bit hard to see in a textual diff, but what this accomplishes is that all values of the old logfile are *exactly* preseverd, such that...
```
diff = merged.loc[:]-old.loc[:]
(diff == 0).all()
```
... outputs `True` for all columns but the new one (`Altitude Geodetic (ft)`).

I also found some comments listing the log columns, so I updated those.

The last commit still includes all changes to remove one of the duplicate altitude/distance AGL columns. May be omitted, if it's not desired to remove that column.

simfinite:
hmm, the failures on Travis CI, I cannot reproduce reliably with my local setup. Although I have seen the segfault in `TestInputSocket` and also failure of `CheckScripts` occasionally when running tests in parallel with `-j8`.

bcoconni:
You can get the same result when the property `position/geod-alt-ft` is specified in the `<output>` element : 
```xml
<output name="result.csv" type="CSV" rate="10">
  <property> position/geod-alt-ft </position>
</output>
```
So there is no reason to add C++ code to get that value (and that saves the trouble about the file `check_cases/orbit/logged_data/BallOut.csv`).

bcoconni:
You can get the same result when the property `position/geod-alt-ft` is specified in the `<output>` element : 
```xml
<output name="result.csv" type="CSV" rate="10">
  <property> position/geod-alt-ft </position>
</output>
```
So there is no reason to add C++ code to get that value (and that saves the trouble about the file `check_cases/orbit/logged_data/BallOut.csv`).

simfinite:
Ok, thank's for the help. I didn't know that one could specify additional properties, so that's good to know. For what it's worth, I still think removing the duplicate distance/altitude AGL would be an improvement to the C++ code, but probably not worth the trouble. Closing this now, as it's not required to be merged anymore.

PR #193, opened 2019-06-08
Modify includes in header files, to all use relative paths

repagh:
After trying to use JSBSim as a library (from packages created with rpm), I found I needed to add the JSBSim include folder to the include path for compilation. 

This PR corrects the #include statements, to use relative paths to all headers under the JSBSim include folder.


seanmcleod:
Double-checking, are you wanting to copy the JSBSim source into your application's source tree and don't want to explicitly define an include path for JSBSim?

repagh:
No, I use the following steps:

* created an rpm package with jsbsim (on suse build service)
* want to use the installed headers under /usr/include/JSBSim, by simply doing `#include <JSBSim/FGFDMExec,h>`

It appears that the FGFDMExec.h then more or less pulls in the whole tree of headers. These do things like `#include <math/FGModelFunctions.h>`. That works, but only if I add a `-I /usr/include/JSBSim` to my compile flags. However that shouldn't be necessary. Maybe I am just being picky, but when using quotes (`""`) in an include statement, you shouldn't need the include path to find the header, but indicate the location relative to the currently included header (which will be the first place where the preprocessor looks).

repagh:
I am puzzled now about the build fault with python 2.7 ........

seanmcleod:
> I am puzzled now about the build fault with python 2.7 ........

That's a known issue with the TestInputSocket test - https://github.com/JSBSim-Team/jsbsim/issues/31

seanmcleod:
Okay so you want to be able to install the JSBSim headers to a compiler's known/standard include path and not have to specify an include path to your compiler for the JSBSim headers.

I guess the other option is to follow the pattern used by the boost C++ library for example.

So for example if you take a look at `<boost/core/addressof.hpp>` it has an include of `<boost/config.hpp>` as opposed to `"../config.h"`

repagh:
Yes, that is also an option. Boost then uses the <> brackets for the include file, and specifies the full path. But that works for boost, it will by default not work for JSBSim, since, when compiling the source, you use the header in the extracted tarball or git repository, and there is no guarantee that the main folder is then called JSBSim.

Limiting the number of -I directives in your compiler flags is anyhow a good thing; the fewer paths where you look for header files, the fewer clashes happen where two headers accidentally have the same name.

bcoconni:
@repagh 
The philosophy of the headers inclusion in JSBSim had never been questioned until you published this PR so I had to perform a bit of investigation to see if the design of JSBSim has anything unusual among C++ libraries. I must admit that I did not find any decisive argument that would allow to make a decision one way or another.

Our current requirement of specifying the path to the headers via the `-I` compiler option is quite common among C++ libraries and can therefore be considered good practice or, at the very least, reasonable practice.
I also found [this discussion on Stack Overflow](https://stackoverflow.com/questions/32117873/c-c-relative-include-paths-vs-makefile-include-flags) but it does not advocate in favor or against your proposal. The only statement that seems to be shared among the participants is that the use of the parent path `..` is considered bad practice:
> always use the project root as working directory. All explicit relative paths are from this root. That way you can avoid the problematic parent `..` path.

JSBSim is currently meeting this recommendation but if your PR was pulled, it would no longer be the case.

All in all, I do not find that your PR brings much added value to the project while it is impacting the whole source code. The design decision has been made more than 20 years ago and I do not see any strong argument to challenge it. So unless someone brings a decisive argument to the discussion, my opinion is to stay on the safe side and keep the current design as it is.

repagh:
OK, keep it as it is then.

PR #198, opened 2019-07-20
Do not define local boost-like functionality, if boost has already be…

repagh:
…en included

This solves cases where boost is used elsewhere in a project.


bcoconni:
Merged. Thanks for tour contribution.

PR #222, opened 2019-10-11
Modify some includes

bgaudron:
Change `#include <expat_config.h>` to `#include "expat_config.h"` to be able to compile under mingw and macOS.


bcoconni:
Committed. Thanks for your contribution.

Issue #219, opened 2019-09-30
function: <random> and <urandom>: fixed value throughout simulation


mike402:
A fellow aircraft dev has just found out `<random/>` and `<urandom/>` from `<function>` have the same value throughout the whole session, so it is impossible to make such things as random failures in terms of MTBF without external hacks.

But it would have been far more useful to have those funcs pick a new number every iteration, in that case if one would want the old behaviour they could use a `<switch>` that would latch the values on command.

Or at least, if too many aircraft rely on them staying the same now (but I don't think so) it could be easier to make some property that would trigger their regeneration when set to 1 and keep the old way when set to 0?

abassign:
I confirm this exactly if I insert the method <random> in the function I do not get the variation of the value both with <random> and with <urandom>

    <function name="systems/dragchute/get-rnd">
        <random/>
    </function>

I can not understand as it seems to me a mistake too obvious for not being already reported.

legoboyvdlp:
No need to do anything; just reinitialize the random seed per documentation:
http://www.cplusplus.com/reference/cstdlib/rand/
e.g.
`srand (time(NULL));`

mike402:
Seed defines the _sequence_ of random numbers, but the problem is that those functions only take the first random number out of the sequence and that's it, it never ever changes again. But of course it would be nice to have different seed for each run.

jonsberndt:
You need to be careful when considering this behavior. There is a distinction between randomizing a value for each simulation frame, and for providing an initial random value for something such as a set of monte carlo simulation runs. I haven’t looked at this code in a while, but may get a chance this weekend. 

Jon

mike402:
On Mon, Sep 30, 2019 at 09:25:55AM -0700, Jon S. Berndt wrote:
> You need to be careful when considering this behavior. There is a distinction between randomizing a value for each simulation frame, and for providing an initial random value for something such as a set of monte carlo simulation runs. I haven’t looked at this code in a while, but may get a chance this weekend. 

Of course, that's why it would probably need to be opt-in until maybe
it's clear none of the aircraft rely on old behaviour.

You would be able to freeze the ever-changing value easily with e.g.
`<switch>` if you need randomized constant instead of a sequence. But
the current behaviour does not let you get a sequence at all.

jonsberndt:
[...]

PR #223, opened 2019-10-12
Allow functions to specify they return non-const values

seanmcleod:
Add the ability for functions to specify that they return non-const values even if all their input parameters are const or if they don't take any input parameters.

This is to fix the issue with the `random` and `urandom` functions return a single const value every time they were invoked - https://github.com/JSBSim-Team/jsbsim/issues/219

### DIFF
--- a/src/math/FGFunction.cpp
+++ b/src/math/FGFunction.cpp
@@ -74,9 +74,10 @@ class aFunc: public FGFunction
 {
 public:
   aFunc(const func_t& _f, FGFDMExec* fdmex, Element* el,
-        const string& prefix, FGPropertyValue* v)
+        const string& prefix, FGPropertyValue* v, bool constForConstParams=true)
     : FGFunction(fdmex->GetPropertyManager()), f(_f)
   {
+    this->constForConstParams = constForConstParams;
     Load(el, v, fdmex, prefix);
     CheckMinArguments(el, Nmin);
     CheckMaxArguments(el, Nmax);
@@ -518,12 +519,12 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
       auto f = [](const decltype(Parameters)& p)->double {
                  return GaussianRandomNumber();
                };
-      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var));
+      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var, false));
     } else if (operation == "urandom") {
       auto f = [](const decltype(Parameters)& p)->double {
                  return -1.0 + (((double)rand()/double(RAND_MAX))*2.0);
                };
-      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var));
+      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var, false));
     } else if (operation == "switch") {
       string ctxMsg = element->ReadFrom();
       auto f = [ctxMsg](const decltype(Parameters)& p)->double {
@@ -802,12 +803,14 @@ FGFunction::~FGFunction()
 
 bool FGFunction::IsConstant(void) const
 {
-  for (auto p: Parameters) {
-    if (!p->IsConstant())
-      return false;
-  }
-
-  return true;
+  if (constForConstParams) {
+    for (auto p: Parameters) {
+      if (!p->IsConstant())
+        return false;
+    }
+    return true;
+  } else
+    return false;
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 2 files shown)


src/math/FGFunction.cpp
```diff
@@ -802,12 +803,14 @@ FGFunction::~FGFunction()
...
-
-  return true;
+  if (constForConstParams) {
+    for (auto p: Parameters) {
+      if (!p->IsConstant())
+        return false;
+    }
+    return true;
+  } else
+    return false;
```
bcoconni: This seems too complicated to me. Since all the functions with no parameters are meant to be non-const why not simply test if Parameters is empty ?
```diff
bool FGFunction::IsConstant(void) const {
+  if (Parameters.empty()) return false;

  for (auto p: Parameters) {
    if (!p->IsConstant())
      return false;
  }

  return true;
}
```

src/math/FGFunction.cpp
```diff
@@ -74,9 +74,10 @@ class aFunc: public FGFunction
 {
 public:
   aFunc(const func_t& _f, FGFDMExec* fdmex, Element* el,
-        const string& prefix, FGPropertyValue* v)
+        const string& prefix, FGPropertyValue* v, bool constForConstParams=true)
```
bcoconni: Remove the `constForCostParams` member, it is not needed (see comment below in `FGFunction::IsConstant`)

src/math/FGFunction.cpp
```diff
@@ -74,9 +74,10 @@ class aFunc: public FGFunction
...
     : FGFunction(fdmex->GetPropertyManager()), f(_f)
   {
+    this->constForConstParams = constForConstParams;
```
bcoconni: Remove the `constForCostParams` member, it is not needed (see comment below in `FGFunction::IsConstant`)

src/math/FGFunction.cpp
```diff
@@ -518,12 +519,12 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
       auto f = [](const decltype(Parameters)& p)->double {
                  return GaussianRandomNumber();
                };
-      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var));
+      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var, false));
```
bcoconni: Remove the `constForCostParams` member, it is not needed (see comment below in `FGFunction::IsConstant`)

src/math/FGFunction.cpp
```diff
@@ -518,12 +519,12 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
...
     } else if (operation == "urandom") {
       auto f = [](const decltype(Parameters)& p)->double {
                  return -1.0 + (((double)rand()/double(RAND_MAX))*2.0);
                };
-      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var));
+      Parameters.push_back(new aFunc<decltype(f), 0>(f, fdmex, element, Prefix, var, false));
```
bcoconni: Remove the `constForCostParams` member, it is not needed (see comment below in `FGFunction::IsConstant`)

src/math/FGFunction.h
```diff
@@ -735,7 +735,7 @@ class FGFunction : public FGParameter, public FGJSBBase
   /// Default constructor.
   FGFunction()
     : cached(false), cachedValue(-HUGE_VAL), pNode(nullptr), pCopyTo(nullptr),
-      PropertyManager(nullptr) {}
+      PropertyManager(nullptr), constForConstParams(true) {}
```
bcoconni: Remove the `constForCostParams` member, it is not needed (see comment below in `FGFunction::IsConstant`)

src/math/FGFunction.h
```diff
@@ -798,6 +798,8 @@ class FGFunction : public FGParameter, public FGJSBBase
   double cachedValue;
   std::vector <FGParameter_ptr> Parameters;
 
+  bool constForConstParams;
+
```
bcoconni: Remove the `constForCostParams` member, it is not needed (see comment below in `FGFunction::IsConstant`)

src/math/FGFunction.cpp
seanmcleod: Yep, I did mention that as an option in the issue report a while ago, but hadn't received any feedback on which approach was the better option.

> or modify FGFunction::IsConstant() to return false if there are no parameters, i.e. assume functions with no parameters aren't constant.

So I went with the more flexible approach, e.g. maybe we add a function that doesn't take any arguments but it does return a constant value. Although there is no real harm in terms of an incorrect result, just that we could potentially lose some optimization if this constant function was expensive to run. But that is a lot of ifs....

But I guess there is `YAGNI` to consider :wink:

bcoconni: > Yep, I did mention that as an option in the issue report a while ago, but hadn't received any feedback on which approach was the better option.

To be honest, I did not read all the details of the conversation, what I grasped was that the functions '<switch>` and `<urandom>` were incorrectly considered constant by the code and that this should be fixed.

> But I guess there is `YAGNI` to consider :wink:

I could not tell any better :smile:

PR #226, opened 2019-10-21
Changes to compile python extension directly with msvc 

bgaudron:
Change include `<expat.h>` in `'expat.h'` to prevent [file not found error](https://travis-ci.org/galleon/gym-jsbsim/jobs/599072729#L702) and change `std::min` in `(std::min)` to prevent [conflict error](https://travis-ci.org/galleon/gym-jsbsim/jobs/599116888#L776) compiling python extension directly under msvc with a setup.py as in https://github.com/galleon/gym-jsbsim/blob/new/setup.py.

### DIFF
--- a/src/simgear/misc/strutils.cxx
+++ b/src/simgear/misc/strutils.cxx
@@ -376,9 +376,9 @@ namespace simgear {
         vector<string> v1parts(split(v1, "."));
         vector<string> v2parts(split(v2, "."));
 
-        int lastPart = std::min(v1parts.size(), v2parts.size());
+        int lastPart = (std::min)(v1parts.size(), v2parts.size());
         if (maxComponents > 0) {
-            lastPart = std::min(lastPart, maxComponents);
+            lastPart = (std::min)(lastPart, maxComponents);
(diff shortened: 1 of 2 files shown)


bcoconni:
I am a bit skeptical that these source code modifications are needed given that JSBSim and its Python module currently successfully compile on Windows with AppVeyor. In addition I am concerned that this patch would compromise the use of the system `expat` library when the flag `SYSTEM_EXPAT` is set to `ON`.

However I saw in your file `setup.py` that the path to `gym_jsbsim/jsbsim/src/simgear/xml` is missing in the `include_dirs` parameter.
https://github.com/galleon/gym-jsbsim/blob/1d8fa981ded7452179c41312534ed797aa64de6c/setup.py#L63
This is most likely the cause of the build failure.

Regarding the issue with `std::min` I think this is due to the fact that you did not use the flag `/D NOMINMAX` while compiling with MSVC.

bgaudron:
Indeed I applied your advice and the builds passed successfully.
Thank you for your help !

Issue #256, opened 2020-02-18
Segmentation fault in AeromatiC++ 3.3.10


NikolaiVChr:
These input will fail it:

```
$ ./aeromatic-3.3.10.exe

** AeromatiC++ version 3.3.10
Aeromatiic is a JSBSim configuration file generation utility.
Please enter aircraft data when prompted.

You can always enter 'h' to get verbose help

** General Information **

Output directory [] (\Users\Nikolai): .
Create a subdirectory? [] (yes): yes
Overwrite? [] (yes): yes
Aircraft name [] (my_aircraft): trainer
Use dedicates System files? [] (yes): yes
Select a system of measurement [] (no):
 0: English (feet, pounds)
 1: Metric (meters, kilograms)
1
Type of aircraft [] (0):
 0: Light General Aviation (Glider, Small Commuter)
 1: High Performance (WWII Fighter, Military Trainer, Aerobatic, Air Racer)
 2: Fighter Jet
 3: Jet Transport (Passenger Jet Airliner, Transonic Jet Transport)
 4: Propeller Transport (Propeller Airliner, Propeller Transport)
1

** Weight and Balance **

Stall speed VS1 (clean, no flaps) [km/h] (0): 185
Maximum takeoff weight [kg] (4535.15): 6150
Empty weight [kg] (0): 3430
Inertia Ixx [kg/m2] (0):
Inertia Iyy [kg/m2] (0):
Inertia Izz [kg/m2] (0):

** Geometry **

Length [m] (12.192): 10.97
Select a wing shape [] (0):
 0: Straight
 1: Elliptical
 2: Delta
0
Wing span [m] (12.192): 10.86
Wing area [m2] (0): 19.3
Wing aspect ratio [] (0): 5.26
Wing taper ratio [] (1): 0.598
Wing root chord [m] (0): 2.24
Wing incidence [] (0): 0.3
Wing dihedral [] (0): 3.44
Wing sweep (quarter chord) [] (0): 8.483
Htail area [m2] (0): 4.12
Htail arm [m] (0): 4.6
Vtail area [m2] (0): 2.34
Vtail arm [m] (0): 3.97

** Systems **

Segmentation fault
```

NikolaiVChr:
Hmm, I just tried Release_Candidate_v1.0.0 aeromatic.

That one still was called 3.3.10 but it did not fail.

Thats good. Maybe its version number should be bumped, so people know its better.

bcoconni:
[...]

Issue #257, opened 2020-02-18
Aeromatic incidence input


NikolaiVChr:
When I input incidence 0 then it assumes I want it to make a guess of incidence instead of making an actual incidence of 0.

Prepar3DGuy:
If you set 0 or leave it default than this code will set it to 2.
https://github.com/JSBSim-Team/jsbsim/blob/2f2398cbafc0fc4f14dc6157d1923548ec8804ba/utils/aeromatic%2B%2B/Aircraft.cpp#L318-L321

ermarch:
Good point, it might be better to initialize it to something like -1 and test for that when setting it to 2.0

seanmcleod:
@NikolaiVChr can we close this issue?

NikolaiVChr:
yes

PR #260, opened 2020-03-17
Fix issues #256 and #257

ermarch:
Initialize wing incidence to 2.0 directly so 0.0 would be a valid input.
Bump the version number.


bcoconni:
Pull request merged.
Thanks Erik.

PR #261, opened 2020-03-18
Schema definition update (XSD)

Zaretto:
Added elements to JSBSim.xsd

* tank was missing standpipe
* output item missing caption
* copyright (information only)
* description (information only)
* added reference to URL (information only)


bcoconni:
Pull request merged.
Thanks.

Issue #287, opened 2020-05-17
Broken accelerations/Nx, mismatched sign of accelerations/Nz and accelerations/pilot-z-norm.


mike402:
1. Stand the aircraft on the ground, slightly tilt nose up and one wing down.
2. Compare `accelerations/N{x,y,z}` and `accelerations/pilot-{x,y,z}-norm`.
Expected:
- same values because the aircraft is not rotating, so accelerations should be the same no matter where;
- `Nx` not zero when aircraft is tilted, because of gravity.
Get: 
- sign of `Nz` and `pilot-z-norm`are the opposite;
- `Nx` zero unless accelerating, but `pilot-x-norm` works as expected.

bcoconni:
> [...] slightly tilt nose up and one wing down
> [...] Expected: same values because the aircraft is not rotating

My English might well be wrong but as far as I know "tilt" means (slightly) rotate. So I am confused about what you mean there and if it is my English that's wrong then no offense will be taken 😄 
Also could you specify if the aircraft was motionless when you made the two checks ?
> `Nx` not zero when aircraft is tilted, because of gravity.

That's quite strange because there is no `Nx`, just `Ny` and `Nz`:
https://github.com/JSBSim-Team/jsbsim/blob/d73d76c7e875141797d51eba0c45a0fd21fa9e5b/src/models/FGAuxiliary.cpp#L336-L339
> sign of `Nz` and `pilot-z-norm`are the opposite;

That's purposely, `Nz` is meant to be positive equal to 1g when sitting on ground (see the minus sign on line 208):
https://github.com/JSBSim-Team/jsbsim/blob/d73d76c7e875141797d51eba0c45a0fd21fa9e5b/src/models/FGAuxiliary.cpp#L206-L209
> same values because the aircraft is not rotating, so accelerations should be the same no matter where

Well, if the aircraft attitude has been modified (at least one angle among `phi`, `theta`, `psi`) then the orientation of gravity (1g) is modified in the aircraft body frame and so are the components (`x`, `y`, `z`) of the accelerations.

For information, below is the code at stake. All the properties you mentioned are based on accelerations measured in the body frame (variable `in.vBodyAccel`) so it's not unexpected that the components (`x`, `y`, `z`) are modified when the aircraft attitude is modified.
https://github.com/JSBSim-Team/jsbsim/blob/d73d76c7e875141797d51eba0c45a0fd21fa9e5b/src/models/FGAuxiliary.cpp#L205-L216

[...]

PR #290, opened 2020-05-18
Nx acceleration

seanmcleod:
Export Nx body acceleration to match current exports of Nz and Ny.


bcoconni:
Merged, thanks.

[PR linked to the discussion that took place in issue #287]

Issue #291, opened 2020-05-23
Actuator <lag> expects a number only


Octal450:
Hi guys,
I am needing dynamic lag on an actuator in my control system, seems that JSBsim only accepts a number there, and not a property like lag_filter does.

My syntax is as follows:

		<actuator name="fcs-actual/rudder/final-actuator">
			<input>fcs-actual/rudder/cmd-deg</input>
			<rate_limit>fcs-actual/rudder/rate</rate_limit>
			<lag>fcs-actual/rudder/lag</lag>
			<output>fcs-actual/rudder/final-deg</output>
		</actuator>

Can this be added please, that I can put a property (or table) there?

Kind Regards,
Josh

seanmcleod:
@Octal450 about 10 years ago I needed dynamic property support for `rate_limit, lag, delay` in the actuator for a project I was working on, so I added support for properties but never got round to submitting it to JSBSim.

In the meantime I see `rate_limit` does support being specified as a property as well as a numeric.

I'm busy adding support for `lag` to also support being specified as a property or numeric, I'll submit a pull request for review.

Octal450:
Thank you much @seanmcleod 
Kind Regards,
Josh

bcoconni:
Hi all, while I was reviewing @seanmcleod PR #292, I realized that the exact purpose of this change is not clear to me: is this feature intended to change the lag parameter τ a couple of times during a simulation or is it intended to allow a continuous modification of τ ?

I am asking the question because currently the numerical integration of the lag filter is made assuming that the lag parameter τ is constant. So if τ is continuously varying then the numerical scheme will introduce a non negligible error that will accumulate as the simulation progresses. On the hand, if the lag parameter τ will seldomly be changed during a simulation then I guess the error can be deemed acceptable.

The use case is of primary importance because it will tell whether or not we need to modify the numerical algorithm. If we choose not to modify the numerical integration scheme, then I think we should document the limitation.

seanmcleod:
My personal use case and the way I've used it over the years is for the lag parameter (and the rate limit and delay parameters) to be set typically once for a lesson/task, e.g. offset landing task, but in particular if the task is an up and away task the parameter can be changed during the lesson by the instructor via an instructor station.

So these are discrete changes, typically with the aircraft trimmed in-between a change as opposed to continuous changes based on say a table lookup based on for example air-density or something.

[...]

PR #292, opened 2020-05-24
Property support for actuator lag element

seanmcleod:
For feature request - https://github.com/JSBSim-Team/jsbsim/issues/291


src/models/flight_control/FGActuator.h
```diff
@@ -80,7 +80,7 @@ CLASS DOCUMENTATION
 @code
 <actuator name="name">
   <input> {[-]property} </input>
-  <lag> number </lag>
+  <lag> {property name | value} </lag>
```
bcoconni, line 83: According to the discussion in issue #291, it is needed to document that the property is not meant to be continuously modified during a simulation.

PR #311, opened 2020-08-05
Fixed one off error in array iteration

### LINKED ISSUE #310: Possible out of bounds access on array in debug mode
In `src/models/flight_control/FGFilter.cpp` and when `(debug_lvl & 1) && from == 0` the array `C` with 7 slots will be accessed at index `7` and evaluate that value in a boolean context during the last iteration of the loop. cppcheck 1.82 warns on that one ... but maybe this is not a problem? It is a 1-based indexing on that array giving an extra slot never used at index `0` but, index `6` should be the largest one accessed.

```cpp
void FGFilter::Debug(int from)
{
  if (debug_lvl <= 0) return;

  if (debug_lvl & 1) { // Standard console startup message output
    if (from == 0) { // Constructor
      cout << "      INPUT: " << InputNodes[0]->GetName() << endl;

      for (int i=1; i <= 7; i++) {
        if (!C[i]) break;  // here access of C[7] in final iteration - test random value or seg fault?
// ...
```

sthagen:
Closes #310 

Since I maintain an active fork (via github web ui) there maybe quite some merge commits in my history, but this is a single character change.

The deployment stages of travis failed in my fork - is this OK? Hope the checks in this upstream repo succeed though.


bcoconni:
> Since I maintain an active fork (via github web ui) there maybe quite some merge commits in my history, but this is a single character change.

No problem. We are always squashing pull requests anyway to keep the project history clean.
> The deployment stages of travis failed in my fork - is this OK? Hope the checks in this upstream repo succeed though.

The tests have passed in the PR so it's OK. As a general rule, failures in the pull request are a no-go (except of course when pushing new regression tests :smile:)

Thanks for your contribution @sthagen

sthagen:
Thank you for accepting and merging @bcoconni

#### On failures:
In forks (I maintain quite some forks) and in my experience these external build tools running the tests often do not work in the same way as in the upstream repository. 

So, as an outside contributor to a project I am then reluctant to even request the pull when in my [forked repository the change seems to cause failures](https://github.com/sthagen/jsbsim/pull/27) - when in the [original the same change simply works](https://github.com/JSBSim-Team/jsbsim/pull/311) as here is the case.

PR #313, opened 2020-08-05
Implemented iterator update after erase (suggested by Sean McLeod)

### LINKED ISSUE #309: Possible iterator use after call of erase on element
In `src/initialization/FGTrimAnalysis.cpp` the iterator `iControls` may be used after calling `erase`on the element. `cppcheck 1.82` warns on that one ... but maybe this is not a problem?

```cpp
bool FGTrimAnalysis::RemoveControl( TaControl control ) {

  FGTrimAnalysisControl* tac;
  bool result=false;

  mode = taCustom;
  vector <FGTrimAnalysisControl*>::iterator iControls = vTrimAnalysisControls.begin();
  while (iControls != vTrimAnalysisControls.end()) {
      tac=*iControls;
      if( tac->GetControlType() == control ) {
        delete tac;
        vTrimAnalysisControls.erase(iControls);  // conditional erase here
        result=true;
        continue;
      }
      iControls++;  // iterator use here
  }

  return result;
}
```

sthagen:
Details at #309 discussion esp. [latest comment from Sean](https://github.com/JSBSim-Team/jsbsim/issues/309#issuecomment-669171296)

Closes #309


bcoconni:
Thanks for the PR @sthagen.

However this PR is not addressing exactly the same issue than the one reported in #309. Here, you are submitting some modifications to `FGTrimAnalysis::EditState` while we have discussed `FGTrimAnalysis::RemoveControl` in #309.

In addition, I am suggesting that we insert `#error` pragmas wherever iterators are used incorrectly rather than submitting tentative fixes that we cannot test.

sthagen:
I reverted in my fork merging in the enhancements from @bcoconni and suggest to either close this PR #313 for now or that I amend as suggested.

sthagen:
Here is variant two (the team can decide if close or merge - happy with either result).

bcoconni:
Thanks for the update @sthagen.

However as was shown by the initial version of your PR, it seems the same issue exists for `FGTrimAnalysis::EditState()`. I guess a similar PR could be pushed to address that issue as well.

sthagen:
@bcoconni  I will take a look, but there is a break after the erase call in that method, thus no iterator use after erase. ok?

bcoconni:
>  I will take a look, but there is a break after the erase call in that method, thus no iterator use after erase. ok?

Oh yes, right. Forget about my previous comment: since there is a `break` statement, indeed we do not need to update the iterator of `erase()` has been called.

PR #315, opened 2020-08-18
Clean up virtual methods.

bcoconni:
* Fix some warnings from the clang compiler about hidden methods.
* Use C++11 `override` wherever possible
* Renamed `FGModel::Load()` to `FGModel::Upload()` since it is an helper method to load models.

### DIFF
--- a/src/FGFDMExec.cpp
+++ b/src/FGFDMExec.cpp
@@ -721,7 +721,7 @@ bool FGFDMExec::LoadModel(const string& model, bool addModelToPath)
     // Process the metrics element. This element is REQUIRED.
     element = document->FindElement("metrics");
     if (element) {
-      result = ((FGAircraft*)Models[eAircraft])->Load(element);
+      result = Models[eAircraft]->Load(element);
       if (!result) {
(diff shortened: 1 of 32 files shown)


seanmcleod:
Forces me to update my C++11 knowledge 😉 

Not sure I follow the reason for the required name change for `FGModel::Load` to `FGModel::Upload` though.

bcoconni:
> Not sure I follow the reason for the required name change for `FGModel::Load` to `FGModel::Upload` though.

Because we would have had 2 methods named `FGModel::Load`:
https://github.com/JSBSim-Team/jsbsim/blob/69bcaccace2eb70fa3489406e1f51f0952b46108/src/models/FGModel.h#L98
and
https://github.com/JSBSim-Team/jsbsim/blob/69bcaccace2eb70fa3489406e1f51f0952b46108/src/models/FGModel.cpp#L110-L149
The first method being virtual, the other (`Upload`) not. Strictly speaking this would have been perfectly legal to keep the 2 methods with the same name `Load()` since their signatures are different. However I have had the feeling that, for maintenance, the situation would have been confusing since the 2 methods have different (albeit related) purposes. Hence the renaming.

Issue #316, opened 2020-08-29
Plans for Ubuntu Focal release


Jaeyoung-Lim:
I was wondering if there were any plans for a release on Ubuntu Focal, since this is the latest LTS of Ubuntu.

I have tried installing the Ubuntu 18.04 rolling release deb package from this repo, but it doesn't seem to work. 

The way currently I installed jsbsim on 20.04 was by installing flightgear. However, it would be nice if we could have a jsbsim deb package as a rolling release. 

Related: https://github.com/PX4/containers/pull/276

bcoconni:
Generally speaking, we are constrained by the availability of the different versions of Ubuntu LTS on GitHub servers. Fortunately, it seems that [Ubuntu 20.04 LTS is now available on GitHub hosted runners](https://docs.github.com/en/actions/reference/software-installed-on-github-hosted-runners) so we need to update our GitHub workflow to build packages for 20.04.

I have added this issue to our project "Migrating our CI workflow to GitHub".

Jaeyoung-Lim:
@bcoconni Thanks!

bcoconni:
> If it is a matter of just extending this workflow to a matrix build, I think I can also try to extend it. Is there anything more that needs to be done?

Yes, just that workflow. Nothing else.

PR #317, opened 2020-08-31
Enable Ubuntu focal release on github workflows

Jaeyoung-Lim:
This adds a matrix configuration to enable releasing rolling releases for Ubuntu 20.04(Focal)

Fixes https://github.com/JSBSim-Team/jsbsim/issues/316

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -45,6 +45,7 @@ jobs:
           cd build
           cpack
       - name: Build ManyLinux wheel packages (PEP 513)
+        if: ${{ matrix.os == 'ubuntu-20.04'}}
         run: |
           cd build
           rm -f CMakeCache.txt
@@ -57,6 +58,7 @@ jobs:
           # Build the manylinux1 wheel binary packages
           docker run --rm -e GITHUB_RUN_NUMBER=$GITHUB_RUN_NUMBER -e GITHUB_SHA=$GITHUB_SHA -v $PWD/..:/io quay.io/pypa/manylinux2010_x86_64 /io/build/build-wheels.sh
       - name: Test wheel package
+        if: ${{ matrix.os == 'ubuntu-20.04'}}
         run: |
           pip install jsbsim --no-index -f build/python/dist
           python -c "import jsbsim;fdm=jsbsim.FGFDMExec('.', None);print(jsbsim.FGAircraft.__doc__)"
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -4,7 +4,10 @@ on: [push]
 
 jobs:
   Linux:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Install Ubuntu packages
         run: sudo apt-get install doxygen cxxtest valgrind
@@ -89,8 +92,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          name: ${{ github.job}} rolling release
-          tag_name: ${{ github.job }}
+          name: ${{ github.job}} rolling release for ${{ matrix.os }}
+          tag_name: ${{ github.job }}-${{ matrix.os }}
           prerelease: true
           body: |
             Contains the ${{ github.job}} packages built from the bleeding edge code
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -92,8 +92,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          name: ${{ github.job}} rolling release for ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


Jaeyoung-Lim:
@bcoconni Is the appveyor CI failure related to this PR? I don't see how it can be related

seanmcleod:
Looks like an issue with the generation of JSBSIM_VERSION define?

Jaeyoung-Lim:
@seanmcleod How is this related to my PR? As far as I can see, github actions should not be executed with my change until it is merged

seanmcleod:
@Jaeyoung-Lim it's not. I was pointing out that the error seems to be due to the generation of the JSBSIM_VERSION define.

bcoconni:
I haven't had the time to investigate much but for some reasons, the Python wheel packages seem not to be built. Neither for 20.04 nor for 18.04.

bcoconni:
Correction: they are built but not released.

Jaeyoung-Lim:
@bcoconni You mean on the github workflow? Can you show me the log of this happening please?

bcoconni:
Regarding wheel packages, I've investigated a bit much and the problem is actually in our `master` branch, not in your PR.
The problem is that wheel packages are now built with the `ManyLinux2010` version while they used to be `ManyLinux1`. In my personal fork, I've tried to force `auditwheel` to build `ManyLinux1` but it fails with an error message `auditwheel: error: cannot repair "python/dist/JSBSim-1.1.0.dev1-257-cp35-cp35m-linux_x86_64.whl" to "manylinux1_x86_64" ABI because of the presence of too-recent versioned symbols.`.

What this means is that we are no longer releasing wheel packages for Linux for some time (weeks ? months ?) since the regular expression below restricts to `manylinux1*` while the built wheels were named `manylinux2010*`:
https://github.com/JSBSim-Team/jsbsim/blob/679bda6980d8f8e47791c66a0c9bb0d6c9e42a83/.github/workflows/cpp-python-build.yml#L98-L100

@Jaeyoung-Lim, you can disregard this problem: I'll fix it directly in `master`.

.github/workflows/cpp-python-build.yml
```diff
@@ -89,8 +92,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          name: ${{ github.job}} rolling release
-          tag_name: ${{ github.job }}
+          name: ${{ github.job}} rolling release for ${{ matrix.os }}
+          tag_name: ${{ github.job }}-${{ matrix.os }}
```
bcoconni, line 96:
```suggestion
          name: ${{ github.job}} rolling release
          tag_name: ${{ github.job }}
```
I would prefer to release everything under the `Linux rolling release` rather than have a different release for each Ubuntu version: since the package file names are tagged with the Ubuntu version, there is no possible confusion between 18.04 packages and 20.04 packages.

bcoconni (changes requested):
I'd rather only have one Linux release that would contain both Ubuntu 18.04 and 20.04 packages.

Also, to reduce the build time, I think the wheel packages should only be built once (either for 18.04 or 20.04). I think this can be achieved with the command `if:`
```diff
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -42,6 +42,7 @@ jobs:
           cd build
           cpack
       - name: Build ManyLinux wheel packages (PEP 513)
+        if: ${{ matrix.os == 'ubuntu-20.04'}}
         run: |
           cd build
           rm -f CMakeCache.txt
@@ -54,6 +55,7 @@ jobs:
           # Build the manylinux1 wheel binary packages
           docker run --rm -e GITHUB_RUN_NUMBER=$GITHUB_RUN_NUMBER -e GITHUB_SHA=$GITHUB_SHA -v $PWD/..:/io quay.io/pypa/manylinux2010_x86_64 /io/build/build-wheels.sh
       - name: Test wheel package
+        if: ${{ matrix.os == 'ubuntu-20.04'}}
         run: |
           pip install jsbsim --no-index -f build/python/dist
           python -c "import jsbsim;fdm=jsbsim.FGFDMExec('.', None);print(jsbsim.FGAircraft.__doc__)"```

bcoconni:
> Looks like an issue with the generation of JSBSIM_VERSION define?

@seanmcleod yes I agree. That's definitely weird since the Windows build from GitHub Actions seems to be happy with the same build instructions... And I'm struggling to see any relation with the current PR !?! 

@Jaeyoung-Lim since Windows builds successfully on GitHub, I'd suggest to ignore AppVeyor failure.

seanmcleod:
@bcoconni is there a reason/some advantage to continue using AppVeyor in addition to Github Actions?

bcoconni:
> is there a reason/some advantage to continue using AppVeyor in addition to Github Actions?

The long term goal is indeed to drop AppVeyor and Travis CI but before doing that there are a number of tasks that need to be completed. The task list is compiled in the project [Migrating our CI workflow to GitHub](https://github.com/JSBSim-Team/jsbsim/projects/3) that I've created especially for that. Not all of the tasks need to be completed to drop AppVeyor and TravisCI: as soon as GitHub actions is on par with AppVeyor and Travis CI build processes we can get rid of them.

bcoconni (approved):
Looks good to me :+1: 
Thanks @Jaeyoung-Lim

Jaeyoung-Lim:
@bcoconni Thanks!

PR #318, opened 2020-09-06
Add warning messages for inconsistent gravity settings.

bcoconni:
As per the discussion in issue #184, this PR adds warnings when the user select gravity settings that are not entirely consistent with the planet shape (spherical vs non-spherical).

### DIFF
--- a/src/models/FGInertial.cpp
+++ b/src/models/FGInertial.cpp
@@ -107,6 +107,12 @@ bool FGInertial::Load(Element* el)
 
   GroundCallback->SetEllipse(a, b);
 
+  // Messages to warn the user about possible inconsistencies.
+  if (a != b && J2 == 0.0)
+    cout << "Gravitational constant J2 is null for a non-spherical planet." << endl;
+  if (a == b && J2 != 0.0)
+    cout << "Gravitational constant J2 is non-zero for a spherical planet." << endl;
+
   Debug(2);
 
   return true;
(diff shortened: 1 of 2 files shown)


src/models/FGInertial.cpp
```diff
@@ -217,11 +223,31 @@ void FGInertial::SetAltitudeAGL(FGLocation& location, double altitudeAGL)
...
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+void FGInertial::SetGravityType(int gt)
+{
+  // Messages to warn the user about possible inconsistencies.
+  switch (gt)
+  {
+  case eGravType::gtStandard: 
+    if (a != b)
+      cout << "Set standard gravity for a non-spherical planet" << endl;
```
seanmcleod, line 233: Suggest the following:

"Warning: Standard gravity model has been set for a non-spherical planet"

src/models/FGInertial.cpp
```diff
@@ -217,11 +223,31 @@ void FGInertial::SetAltitudeAGL(FGLocation& location, double altitudeAGL)
...
+    break;
+  case eGravType::gtWGS84:
+    if (J2 == 0.0)
+      cout << "Set WGS84 with a null gravitational constant J2." << endl;
```
seanmcleod, line 237: Suggest the following:

"Warning: WGS84 gravity model has been set without specifying the J2 gravitational constant."

seanmcleod (commented):
Commented on the wording for some of the warning messages.

src/models/FGInertial.cpp
bcoconni: Done.

PR #319, opened 2020-09-08
AeromatiC++: Also output a linear coefficient only JSON configuration file

ermarch:
In addition to outputting to a comprehensive JSBSim XML configuration file also output a simplified linear coefficients JSON configuration. This may be used by FlightGears upcoming AISim but could be useful for other purposes too. All positions are relative to C.G. already.

### DIFF
--- a/utils/aeromatic++/Aircraft.cpp
+++ b/utils/aeromatic++/Aircraft.cpp
@@ -190,7 +190,7 @@ Aeromatic::Aeromatic() : Aircraft(),
     _CDalpha.resize(4, 0.0f);
     _CYp.resize(4, 0.0f);
     _Clbeta.resize(8, 0.0f);
-    _Clr.resize(8, 0.0f);
+    _Clr.resize(9, 0.0f);
     _Cnp.resize(4, 0.0f);
 
     _Cna.resize(8, 0.0f);
@@ -223,8 +223,7 @@ Aeromatic::~Aeromatic()
 bool Aeromatic::fdm()
 {
     Aircraft *aircraft = _aircraft[_atype];
-    std::vector<System*> systems = _aircraft[_atype]->get_systems();
-
+    systems = _aircraft[_atype]->get_systems();
     _engines = _MIN(_no_engines, 4);
     aircraft->_engines = _engines;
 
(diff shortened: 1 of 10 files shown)


bcoconni:
Should not `aeromatic++` be a project of its own now that it provides features beyond JSBSim ?

ermarch:
On 9/8/20 8:08 PM, Bertrand Coconnier wrote:
> Should not |aeromatic++| be a project of its own now that it provides 
> features beyond JSBSim ?

It's a bit of a difficult situation: The code is too useful not to 
implement and too small for a separate project of it's own, which I did 
consider for a brief moment.

If you want to make it a separate project I'm okay with that but I think 
it would be very useful to keep it under control of the JSBSim 
developers. Generating JSBSim configuration files still is, and always 
will be, it's main purpose.

This was just of one of those nice things to have.

Erik

bcoconni:
OK. I just asked the question because with this new addition of the JSON feature, I thought it was worth clarifying the situation. As far as I am concerned, it's perfectly fine to keep `aeromatic++` in JSBSim. However everyone has to keep in mind that this is a viable option as long as you hang around the JSBSim project to maintain this code. Unfortunately we have a couple of examples of rotten code that had to be dropped due to the lack of maintenance such as `FGTrimAnalysis` and `FGSimplexTrim`.

ermarch:
On 9/8/20 10:53 PM, Bertrand Coconnier wrote:
> OK. I just asked the question because with this new addition of the JSON 
> feature, I thought it was worth clarifying the situation. As far as I am 
> concerned, it's perfectly fine to keep |aeromatic++| in JSBSim. However 
> everyone has to keep in mind that this is a viable option as long as you 
> hang around the JSBSim project to maintain this code. Unfortunately we 
> have a couple of examples of rotten code that had to be dropped due to 
> the lack of maintenance such as |FGTrimAnalysis| and |FGSimplexTrim|.

I never envisioned me being the only contributor to AeromatiC++ but so 
far it turns out I am. I'm not really worried about bit-rot since it is 
a stand-alone project which only interacts with JSBSim by creating it's 
configuration files. So if anything changes there (which I don't see 
happening very often, if at all) then the code needs some attention but 
that should be fairly straight forward.

PR #322, opened 2020-09-13
AeromatiC++:  Fix the air density at sea level for propellers, add CG x-position.

ermarch:
Fix the air density at sea level for propellers (it was in metric units by mistake). Calculate the X-position of the center of gravity relative to the aerodynamic reference point. Bump the version number and add thrust for turbine and rocket engines in the JSON configuration.


ermarch:
Discovered a small problem. Will create a new pull request when this is fixed.

Issue #173, opened 2019-03-13
AeromatiC++


swampmo:
With piston engines I wonder if we could ask how many blades the prop has.

bcoconni:
@ermarch
Could you please review this request ? Thanks.

ermarch:
The idea has always been to keep the number of parameters low, and adjust the generated configurations afterwards by hand. It's not a replacement for DATCOM and JavaProp. So in that sense   I'm inclined to say no.

But the propeller code needs a good overhaul anyhow since it's output is not the best in the world to say it mildly. I could take the request into account when working on it.

bcoconni:
> I could take the request into account when working on it.

Thanks. I've assigned this issue to you and it is now labeled as a `Feature request`.

PR #323, opened 2020-09-14
Ovehaul of the propeller code, other small fixes

ermarch:


ermarch:
> Compute the x-offset of the center of gravity from aircraft parameters
> Use the correct value for rho, not the metric one
> Overhaul of the propeller thrust and power math: It's getting pretty close to propeller data generated with JavaProp or from other data.

bcoconni:
@ermarch Is the PR complete, ready to be merged ? Or are you planning to commit some additional changes ?

ermarch:
No, this should be it. I've certainly not planned anything for the 
foreseeable future.

Erik

bcoconni:
Great ! Thanks.
Does this PR fixes the issue #173 as well ? If so, could you please tell the issue poster that the feature request has been addressed ?
Thanks.

ermarch:
On 9/15/20 11:10 PM, Bertrand Coconnier wrote:
> Great ! Thanks.
> Does this PR fixes the issue #173 
> <https://github.com/JSBSim-Team/jsbsim/issues/173> as well ? If so, 
> could you please tell the issue poster that the feature request has been 
> addressed ?

No this issue has not been solved with this patch since it needs some 
considerations on how to keep it backwards compatible:

AeromatiC++ can output it's parameters to a file and read them back for 
a second run, for instance when AeromatiC++ has been improved. And 
messing with the number of inputs is a bit tricky at the moment since it 
just reads the file as if it where the standard input.

Erik

ermarch:
The estimation code for the number of blades has changes though and this 
also influences the propeller material: wood, aluminum or carbon fiber 
for calculation of the propeller inertia.

Issue #332, opened 2020-10-18
Magnetometer Startup Sequence Causes ~5 Seconds of Erroneous Data


mvacanti:
**Issue Description**
Using an airframe with a [magnetometer system](https://github.com/JSBSim-Team/jsbsim/blob/master/src/models/flight_control/FGMagnetometer.cpp) the startup sequence does not correctly apply the magnetic variation until `counter` reaches 1,000 - effectively delaying the correct publication of magnetic values for several seconds after startup. 

https://github.com/JSBSim-Team/jsbsim/blob/3743a5c3d849e16cb7d23bdfdf7322608f25246a/src/models/flight_control/FGMagnetometer.cpp#L103-L117

**Observed Issue (output csv)**

Time | /fdm/jsbsim/px4/mag-x | /fdm/jsbsim/px4/mag-y | /fdm/jsbsim/px4/mag-z
-- | -- | -- | --
0 | 0 | -nan | 1.27066231393875E+236
0.0083333 | 0 | -nan | 1.27062309229309E+236
0.0166666 | 0 | -nan | 1.27062309229387E+236
0.0249999 | 0 | -nan | 1.2706230898948E+236
0.0333332 | 0 | -nan | 1.27062308521534E+236
0.0416665 | 0 | -nan | 1.27062307843842E+236
0.0499998 | 0 | -nan | 1.27062306968209E+236
0.0583331 | 0 | -nan | 1.27062305912773E+236
0.0666664 | 0 | -nan | 1.27062304689168E+236
0.0749997 | 0 | -nan | 1.27062303315325E+236
0.083333 | 0 | -nan | 1.27062301802605E+236

**To Duplicate Observed Issue**
Using existing release version of JSBSim and forked versions of [737.xml](https://github.com/mvacanti/jsbsim/blob/pr-update-mag-startup/aircraft/737/737.xml), [B737_Runway](https://github.com/mvacanti/jsbsim/blob/pr-update-mag-startup/scripts/B737_Runway.xml) and [px4_default_mag_sensor.xml](https://github.com/mvacanti/jsbsim/blob/pr-update-mag-startup/systems/px4_default_mag_sensor.xml) system file:  `JSBSim --script=scripts/B737_Runway --realtime`

**Proposed Solution**
https://github.com/mvacanti/jsbsim/blob/468dec7ff7a0b697c1ef5d576961ea0fabb313a3/src/models/flight_control/FGMagnetometer.cpp#L99-L112

```c++
[...]

PR #334, opened 2020-10-20
Resolve Mag startup sequence (#332)

mvacanti:
Resolve mag startup sequence per #332.


seanmcleod (changes requested):
You need to delete the `counter++;` on line 105, otherwise the magnetic variation won't be calculated on the 1st simulation frame, it will only be calculated for the 1st time after 500 frames now.

mvacanti:
Thanks for the catch.... Apologize, I created a new branch for the PR ([original issue branch is correct](https://github.com/mvacanti/jsbsim/blob/84e182628ffa9a441a4adf2c2d7a4c2777670bde/src/models/flight_control/FGMagnetometer.cpp#L101)) and manually edited... Updated now.

Issue #357, opened 2020-12-11
Negative drag due to elevator deflection


seanmcleod:
While debugging a difference in trim results I noticed that the drag due to elevator deflection was negative for the 737 model.

```sh
    aero/coefficient/CD0 = 2801.679561
    aero/coefficient/CDi = 3813.129823
    aero/coefficient/CDflap = 5689.318388
    aero/coefficient/CDgear = 1928.582504
    aero/coefficient/CDsb = 0.000000
    aero/coefficient/CDsb = 0.000000
    aero/coefficient/CDsp = 0.000000
    aero/coefficient/CDbeta = 0.007089
    aero/coefficient/CDde = -181.721539
```

Taking a look at the 737 model's calculation for elevator deflection drag it's using `elevator-pos-norm` `[-1, +1]` without converting it to an absolute value, hence the reason you can see negative drag values.

```xml
<function name="aero/coefficient/CDde">
    <description>Drag_due_to_Elevator_Deflection</description>
    <product>
        <property>aero/qbar-psf</property>
        <property>metrics/Sw-sqft</property>
        <property>fcs/elevator-pos-norm</property>
        <value>0.059</value>
    </product>
</function>
```

In context that means at around the critical AoA of ~12.5 degrees the negative drag from a full elevator deflection will completely cancel out the total induced drag!

![image](https://user-images.githubusercontent.com/824856/101910664-eadfd380-3bc7-11eb-802d-6de39613db17.png)

Also if the aircraft was trimmed at say ~2 degrees AoA and you put in a sudden step of full elevator deflection the aircraft would actually start accelerating with this extra negative drag 'thrust'.

In addition to the sign issue with the elevator deflection the use of a normalized position `[-1, +1]` is unusual since the coefficient is normally per radian. It's not impossible that it's been converted/scaled, however the value of `0.059` is very similar to the typical values published that are per radian.

The 737 model for example has a maximum elevator deflection of `0.3 rad` so if the figure of `0.059` is per radian then that would equate to a value of `0.018` for a normalized range.

I did a search through all the JSBSim aircraft for `/CDde` to see whether the 737 model was an exception or not.

Unfortunately it's not. The following aircraft models follow the same setup as the 737 model.

## Incorrect use of `elevator-pos-norm`

https://github.com/JSBSim-Team/jsbsim/blob/eccfd64bbe09fc209c399bead031a8ccc5873118/aircraft/sgs233/sgs233.xml#L321

[...]

PR #359, opened 2020-12-12
Update CDDe from 0 to 0.06

seanmcleod:
Use a reasonable CDDe coefficient as opposed to the current value of 0.

### DIFF
--- a/aircraft/c172p/c172p.xml
+++ b/aircraft/c172p/c172p.xml
@@ -554,7 +554,7 @@
                     <property>aero/qbar-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>fcs/mag-elevator-pos-rad</property>
-                    <value>0.0000</value>
+                    <value>0.06</value>
                 </product>
             </function>
             <function name="aero/coefficient/CDbeta">
--- a/aircraft/c182/c182.xml
+++ b/aircraft/c182/c182.xml
@@ -486,7 +486,7 @@
                     <property>aero/qbar-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>fcs/mag-elevator-pos-rad</property>
-                    <value>0.0000</value>
+                    <value>0.06</value>
                 </product>
             </function>
         </axis>
--- a/aircraft/c310/c310.xml
+++ b/aircraft/c310/c310.xml
@@ -673,7 +673,7 @@
                     <property>aero/qbar-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>fcs/mag-elevator-pos-rad</property>
-                    <value>0.0000</value>
+                    <value>0.06</value>
                 </product>
             </function>
 


bcoconni:
May be worth getting some feedback from the c172p folks about this value ?
@dany93, @wlbragg any opinion ?

dany93:
Personally, I have no idea for this value. Megginson's table gives no value (rather, 0).
As a clue: CDDf (flaps) is about 0.034 to 0.040 / rd. Hence, 0.06 for the elevator seems plausible.

Question: sign?
My reasoning:
The horizontal stabilizer (considered as a whole) gives a negative lift. Which makes me think that:
- elevator up (negative defection, pitch up) should make the drag to increase. "More" negative lift (in absolute value) due to the horizontal stabilizer taken as a whole.
- elevator down (positive defection, pitch down) should make the drag to decrease. "Less" negative lift (in absolute value) due to the horizontal stabilizer taken as a whole.

Which would give a negative value (-0.06).
What do you think?

dany93:
But it wouldn't work with `mag-elevator-pos-rad`, which is always positive.

bcoconni:
> The horizontal stabilizer (considered as a whole) gives a negative lift. Which makes me think that:
>
> * elevator up (negative defection, pitch up) should make the drag to increase. "More" negative lift (in absolute value) due to the horizontal stabilizer taken as a whole.

This is an interesting point that you are bringing. We have just had a discussion about that particular topic in issue #357 from where this PR originated (as well as PR #358). @seanmcleod and myself are thinking that drag always pull you back or said otherwise that it should always be negative. Positive drag means thrust which cannot happen. However I did not envision the drag from the elevator as a difference/delta that is applied to the global drag :thinking:. From that point of view, the elevator drag could indeed be positive as you explain, as long as the global drag is still negative of course.

bcoconni:
On second thoughts, drag is always parallel to the air velocity and, on an elevator, air flows from the leading edge to the trailing edge in all cases (whether the lift is up or down). Since the air flow is moving backward in all cases then the drag is always backward as well i.e. negative. So I am not convinced after all that your reasoning is correct...

seanmcleod:
@dany93 yes you're thinking that given there is a decrease in lift relative to the zero deflection that there is less induced drag relative to the zero deflection, and hence the reasoning for 'negative drag' relative to the zero deflection.

However what you've forgotten to take into account is the combination of the change in parasitic/profile drag and induced drag relative to the zero deflection. The increase in profile drag is larger than the reduction in induced drag so there is still a net increase in drag relative to the zero deflection.

However using a single coefficient value for both deflection directions is a simplification/average since although there is an increase in drag for both deflection directions they're not equal, since in the one direction there is an increase in induced drag and profile drag while in the other there is a decrease in induced drag and an increase in profile drag. Think of the aileron case where this contributes to adverse yaw.

So if you wanted a higher fidelity model of the elevator drag you would use different coefficients for each direction, e.g. say 0.06 for one direction and 0.04 for the other direction. But in both cases it's a net increase in drag.

Remember in terms of elevator direction that the horizontal tail doesn't always produce negative lift, e.g.

image

So if you are going to have different coefficients for each direction you also need to be sure of the horizontal tail setup.

Lastly I came across a Flightgear user in 2009 complaining that he noticed the aircraft increasing it's acceleration during the take-off run by deflecting the elevator.

https://forum.flightgear.org/viewtopic.php?t=2924&p=25689

> I was enjoying flying the 747-200 which was available on the flightgear website and noticed some unrealistic behavior in it. During the take off run on the runway I noticed the acceleration increases when I pull the control column(deflecting elevators upward) and wise versa. Later when I analysed the internal properties I saw the drag coefficient(CDde) was turning negative

bcoconni:
> So if you wanted a higher fidelity model of the elevator drag you would use different coefficients for each direction, e.g. say 0.06 for one direction and 0.04 for the other direction. But in both cases it's a net increase in drag.

So as you mentioned in issue #357, the only aircraft model in our repository that went that far is the f22. That's good to know :smile:
https://github.com/JSBSim-Team/jsbsim/blob/eccfd64bbe09fc209c399bead031a8ccc5873118/aircraft/f22/f22.xml#L1931-L1948

seanmcleod:
Although the F-22 isn't an ideal example since it doesn't have any elevators! :wink:

It only has all moving horizontal stabilizers. So it's drag should follow the typical drag polar for a wing, without any complications in terms of an elevator moving in different directions generating different amounts of drag.

I wonder if the `0.077` value is a typo and shouldn't also be `0.075` or vice-versa.

image

There is also the extra complication of the downwash angle affecting things, maybe the slightly different slope with maximums of `0.077` and `0.075` aren't typos and they take into account some average downwash angle?

If you look hard enough nothing is simple :wink:

dany93:
> Remember in terms of elevator direction that the horizontal tail doesn't always produce negative lift, e.g.
> So if you are going to have different coefficients for each direction you also need to be sure of the horizontal tail setup.

Agreed, that's an important point.

I was considering only the case of an aircraft with CoG in front of the aerocenter. 
My reasoning was more simple that @seanmcleod [explanation above](https://github.com/JSBSim-Team/jsbsim/pull/359#issuecomment-743985732). It implicitely assumed that the parasitic/profile aircraft drag included the elevator drag at its neutral position (as a whole with the horizontal stabilizer). In this case, a contribution change can decrease the total drag. Assuming that this is not the case can be more logical.

> The increase in profile drag is larger than the reduction in induced drag so there is still a net increase in drag relative to the zero deflection

I hadn't gone so far in my reasoning. In this case, you are right for a drag increase in all cases.

A question for my understanding:
- In the [sketch above](https://github.com/JSBSim-Team/jsbsim/pull/359#issuecomment-743803570), you seem to call '_induced drag_' the drag due the angle change between the '_Relative airflow (free stream)'_ and the '_Effective relative airflow_'.
- For me, the induced drag is due to the vortice edge effect, at the free edges of the wings. Which is a different contribution.

This is a difficult subject, but after these explanations I tend to agree that taking a positive drag contribution for every surface change is the best compromise, even if it is an approximation in some  cases.

I don't know if it can be found, but this error may happen too in some aircraft with the rudder drag, if not calculated with magnitude or absolute deflection. Although this case in more obvious.

seanmcleod:
> I don't know if it can be found, but this error may happen too in some aircraft with the rudder drag, if not calculated with magnitude or absolute deflection. Although this case in more obvious.

I haven't seen any JSBSim aircraft (in the JSBSim repo) modeling drag due to rudder deflection, at least the couple that I manually checked. Also searching for `/CDdr` doesn't turn up any hits compared to `CDde` pretty much turning up a hit for every aircraft. But yes you would need to double-check if they're using a signed or unsigned version of the rudder deflection angle if it was modelled.

bcoconni:
It seems we now have a common ground. Commit pulled !
Thanks every one !

dany93:
Is there a connection with the c172p from the main FGAddon( fgdata pack)? Do we (c172p team) have to do a merge ourselves?

bcoconni:
@dany93 
> Is there a connection with the c172p from the main FGAddon( fgdata pack)? Do we (c172p team) have to do a merge ourselves?

No, this model is only maintained in our repo and has no connection with FGAddon and c172p-team's. So if you decide that this change is relevant to your project you are much welcome to merge it to your model.

dany93:
@bcoconni 
Thank you.

I don't know if it can be of interest for you, but I've added a stall and spin feature, up to flat spin, in the FDM. Initially experimental, but still alive since 2014.

bcoconni:
> I don't know if it can be of interest for you, but I've added a stall and spin feature, up to flat spin, in the FDM. Initially experimental, but still alive since 2014.

Of course we are interested, @seanmcleod has initiated a discussion #361 precisely on that topic. May be you'd like to join in ?

PR #364, opened 2020-12-28
Allow property based delay parameter for FCSComponent

seanmcleod:


bcoconni:
Thanks for the PR.

With this patch, the delay is set once during initialization and never will be modified afterwards during the simulation whatever the property value will be. Is that the intended behaviour ?

seanmcleod:
It is for now. I'm still thinking whether my current update mechanism, which currently simply re-sizes the output buffer and fills it with zeros is the best option.

bcoconni:
> I'm still thinking whether my current update mechanism, which currently simply re-sizes the output buffer and fills it with zeros is the best option.

Yeah, filling the buffer with zeros is most likely not a good idea. We should rather initialize it with the first value of `Output` just as is done for the bound property:
https://github.com/JSBSim-Team/jsbsim/blob/868fcd59f4fea8bc051bd44b18a913c1d86de40c/src/models/flight_control/FGFCSComponent.cpp#L288-L293
Similarly, the buffer should also be updated after a trim has been successfully performed. However, these changes are not linked to your PR so I have merged and these topics should be addressed in a separate PR.

Issue #227, opened 2019-10-22
Generate Linear Model from XML Script


benhager:
Does anyone have any experience with, or is it possible to generate a linear model via XML scripting? Looking to call into FGStateSpace to get x0, u0, A, B, C, D vectors. Thanks!

bcoconni:
Your request is too vague. What do you mean by "generate a linear model" ?

bcoconni:
No response from the issuer for 3 weeks: issue closed.

benhager:
Sorry for the delay, i was not notified of a comment on this ticket. I thought generating a linear model is pretty standard, and I believe it is a part of the source code already. It is implemented in [JSBSim Advanced Triming Branch](https://github.com/arktools/jsbsim) by calling into FGLinearization.cpp.

The output should be something like this:
`x0=..
[  3.3500000001e+002;
   1.0833708379e-001;
   1.0833708380e-001;
   0.0000000000e+000;
  -5.8423256804e-004;
   0.0000000000e+000;
   0.0000000000e+000;
   0.0000000000e+000;
   0.0000000000e+000;
   0.0000000000e+000;
   0.0000000000e+000;
   1.5000000000e+002];
u0=..
[  4.1507273608e-001;
  -3.5248398193e-003;
  -2.0575829692e-001;
  -2.9106407215e-003];
sys = syslin('c',..
[ -4.3427060360e-002,  -9.8096550611e+000,  -3.2087058179e+001,  -2.0560557019e-010,   3.3305725641e+002,  -1.8636402009e-002,  -2.0618344994e-010,  -2.1078986357e-010,   6.3607575515e-010,   2.0773530465e-010,  -2.1567803515e-001,   1.4236057475e-005;
  -8.1166258462e-005,  -1.3702931926e-001,   1.1657013512e-002,   1.0000000112e+000,  -2.3830098715e-007,   1.5667392399e-005,   5.7283788879e-004,   6.2980734972e-005,  -1.4375970466e-004,  -1.7085689860e-007,  -2.1476001587e-008,   3.9049977087e-007;
   0.0000000000e+000,   0.0000000000e+000,   0.0000000000e+000,   1.0000000000e+000,   0.0000000000e+000,   0.0000000000e+000,   0.0000000000e+000,   6.0540241507e-036,   0.0000000000e+000,   0.0000000000e+000,   0.0000000000e+000,   0.0000000000e+000;
[...]

PR #370, opened 2021-01-07
Exposing linearization procedures

aarondewindt:
In this pull request I'm exposing the linearization procedures written by @jgoppert back in 2012. For now I'm making them only available through the `DoLinearization(...)` function on the `FGFDMExec` class. The reason I'm doing is, is because I'm interested in trying them out and based on issue #227 there seems to be some interested in this.

Next to the changes needed to expose the code, there are some other changes I had to make in order to build and test the code on my PC. These should not interfere with any current setup, but just to be sure this is the list of extra changes I made.

 - **.gitignore** Added excludes for the Clion (Jetbrains) project files and for the python cache. I'm excluding all project files, not just the user specific files.
 - **tests/JSBSim_utils.py** Changed the way in which the jsbsim repository root path is found when initializing a sandbox. The previous implementation assumed the tests are started by a script in the test directory. This is not the case when using the Clion/Pycharm unit testing integration. So this path is now found by starting with the path to the _tests/JSBSim_utils.py_ file.

I started by modifying the `FGLinearization` class.
- Moved the Scicoslab script generator code to it's own function.
- Made the state space matrices and vectors private members of the class.
- Added `~Get####(...)` functions to get these matrices and vectors.

I added the `DoLinearization(...)` function to the C++ `FGFDMExec` class. The only thing this does, is create a new instance of the `FGLinearization` class and write the results to the referenced vectors.

I added the `do_linearization()` function to the Cython `FGFDMExec` class. It calls it's C++ equivalent, converts the results to python objects (numpy arrays and tuples) and returns the results.

Since these new functions seem to return quite a lot of items I'm considering letting them return the `FGLinearization` object instead. Any objections?

An example on how to use this function can be found in the `tests/TestLinearization.py` unittest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,11 @@ TAGS
 .dir-locals.el
 
 #################
+## Python
+#################
+__pycache__/
+
+#################
 ## Visual Studio
 #################
 
@@ -75,3 +80,11 @@ ipch/
 ## Mac OSX
 #################
 .DS_Store
+
+#################
+## Clion (Jetbrains IDEs)
+#################
+# Project files
+.idea/*
+# CMake
+cmake-build-*/
--- a/python/jsbsim.pxd
+++ b/python/jsbsim.pxd
@@ -31,6 +31,29 @@ cdef extern from "initialization/FGInitialCondition.h" namespace "JSBSim":
         c_FGInitialCondition(c_FGInitialCondition* ic)
         bool Load(const c_SGPath& rstfile, bool useStoredPath)
 
+cdef extern from "initialization/FGLinearization.h" namespace "JSBSim":
+    cdef cppclass c_FGLinearization "JSBSim::FGLinearization":
+        c_FGLinearization(c_FGFDMExec* fdme)
+        
+        void WriteScicoslab() const
+        void WriteScicoslab(string& path) const
+
+        void GetStateSpace(vector[vector[double]]& A_, vector[vector[double]]& B_,
+                           vector[vector[double]]& C_, vector[vector[double]]& D_) const
+
+        vector[double] GetInitialState() const
+        vector[double] GetInitialInput() const
+        vector[double] GetInitialOutput() const
+
+        vector[string] GetStateNames() const
+        vector[string] GetInputNames() const
+        vector[string] GetOutputNames() const
+
+        vector[string] GetStateUnits() const
+        vector[string] GetInputUnits() const
+        vector[string] GetOutputUnits() const
+
+
 cdef extern from "input_output/FGPropertyManager.h" namespace "JSBSim":
     cdef cppclass c_FGPropertyManager "JSBSim::FGPropertyManager":
         c_FGPropertyManager()
@@ -206,19 +229,3 @@ cdef extern from "FGFDMExec.h" namespace "JSBSim":
         shared_ptr[c_FGAircraft] GetAircraft()
         shared_ptr[c_FGAtmosphere] GetAtmosphere()
         shared_ptr[c_FGMassBalance] GetMassBalance()
-
-        void DoLinearization(
-            vector[double]& out_x0,
-            vector[double]& out_u0,
-            vector[double]& out_y0,
-            vector[vector[double]]& out_A,
-            vector[vector[double]]& out_B,
-            vector[vector[double]]& out_C,
-            vector[vector[double]]& out_D,
-            vector[string]& out_state_names,
-            vector[string]& out_input_names,
-            vector[string]& out_output_names,
-            vector[string]& out_state_units,
-            vector[string]& out_input_units,
-            vector[string]& out_output_units,
-        )
--- a/python/jsbsim.pxd
+++ b/python/jsbsim.pxd
@@ -38,20 +38,22 @@ cdef extern from "initialization/FGLinearization.h" namespace "JSBSim":
         void WriteScicoslab() const
         void WriteScicoslab(string& path) const
 
-        void GetStateSpace(vector[vector[double]]& A_, vector[vector[double]]& B_,
-                           vector[vector[double]]& C_, vector[vector[double]]& D_) const
+        vector[vector[double]]& GetSystemMatrix() const
+        vector[vector[double]]& GetInputMatrix() const
+        vector[vector[double]]& GetOutputMatrix() const
+        vector[vector[double]]& GetFeedforwardMatrix() const
 
-        vector[double] GetInitialState() const
-        vector[double] GetInitialInput() const
-        vector[double] GetInitialOutput() const
+        vector[double]& GetInitialState() const
+        vector[double]& GetInitialInput() const
+        vector[double]& GetInitialOutput() const
 
-        vector[string] GetStateNames() const
-        vector[string] GetInputNames() const
-        vector[string] GetOutputNames() const
+        vector[string]& GetStateNames() const
+        vector[string]& GetInputNames() const
+        vector[string]& GetOutputNames() const
 
-        vector[string] GetStateUnits() const
-        vector[string] GetInputUnits() const
-        vector[string] GetOutputUnits() const
+        vector[string]& GetStateUnits() const
+        vector[string]& GetInputUnits() const
+        vector[string]& GetOutputUnits() const
 
 
 cdef extern from "input_output/FGPropertyManager.h" namespace "JSBSim":
--- a/python/jsbsim.pxd
+++ b/python/jsbsim.pxd
@@ -21,6 +21,7 @@
 from libcpp cimport bool
 from libcpp.string cimport string
 from libcpp.memory cimport shared_ptr
+from libcpp.vector cimport vector
 
 cdef extern from "ExceptionManagement.h":
     cdef void convertJSBSimToPyExc()
@@ -205,3 +206,19 @@ cdef extern from "FGFDMExec.h" namespace "JSBSim":
         shared_ptr[c_FGAircraft] GetAircraft()
         shared_ptr[c_FGAtmosphere] GetAtmosphere()
         shared_ptr[c_FGMassBalance] GetMassBalance()
+
+        void DoLinearization(
+            vector[double]& out_x0,
+            vector[double]& out_u0,
+            vector[double]& out_y0,
+            vector[vector[double]]& out_A,
+            vector[vector[double]]& out_B,
+            vector[vector[double]]& out_C,
+            vector[vector[double]]& out_D,
+            vector[string]& out_state_names,
+            vector[string]& out_input_names,
+            vector[string]& out_output_names,
+            vector[string]& out_state_units,
+            vector[string]& out_input_units,
+            vector[string]& out_output_units,
+        )
--- a/src/initialization/FGLinearization.cpp
+++ b/src/initialization/FGLinearization.cpp
@@ -23,12 +23,8 @@
 
 namespace JSBSim {
 
-// TODO make FGLinearization have X,U,Y selectable by xml config file
-
 FGLinearization::FGLinearization(FGFDMExec * fdm) : fdm(fdm), ss(fdm)
 {
-    cout << "Initializing linearization" << std::endl;
-
     ss.x.add(new FGStateSpace::Vt);
     ss.x.add(new FGStateSpace::Alpha);
     ss.x.add(new FGStateSpace::Theta);
@@ -72,23 +68,30 @@ FGLinearization::FGLinearization(FGFDMExec * fdm) : fdm(fdm), ss(fdm)
     u0 = ss.u.get();
     y0 = x0; // state feedback
 
-
-    cout << "Linearizing..." << std::endl;
     ss.linearize(x0,u0,y0,A,B,C,D);
-    cout << "Done linearizing..." << std::endl;
 }
 
-void FGLinearization::GetStateSpace(std::vector<double> & x0_, std::vector<double> & u0_, std::vector<double> & y0_,
-                                    std::vector<std::vector<double>> & A_, std::vector<std::vector<double>> & B_,
-                                    std::vector<std::vector<double>> & C_, std::vector<std::vector<double>> & D_) {
+void FGLinearization::GetStateSpace(std::vector<std::vector<double>> & A_,
+                                    std::vector<std::vector<double>> & B_,
+                                    std::vector<std::vector<double>> & C_,
+                                    std::vector<std::vector<double>> & D_) const {
     A_ = A;
     B_ = B;
     C_ = C;
     D_ = D;
-    x0_ = x0;
-    u0_ = u0;
-    y0_ = y0;
 }
+
+std::vector<double> FGLinearization::GetInitialState() const {
+    return x0;
+};
+
+std::vector<double> FGLinearization::GetInitialInput() const {
+    return u0;
+};
+
+std::vector<double> FGLinearization::GetInitialOutput() const {
+    return y0;
+};
 
 std::vector<std::string> FGLinearization::GetStateNames() const {
     return ss.x.getName();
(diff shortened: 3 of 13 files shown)


bcoconni:
Thanks for contributing to JSBSim !

> - **.gitignore** Added excludes for the Clion (Jetbrains) project files and for the python cache. I'm excluding all project files, not just the user specific files.

OK. I will push that change separately from your PR because it is not linked to the PR purpose.

> - **tests/JSBSim_utils.py** Changed the way in which the jsbsim repository root path is found when initializing a sandbox. The previous implementation assumed the tests are started by a script in the test directory. This is not the case when using the Clion/Pycharm unit testing integration. So this path is now found by starting with the path to the _tests/JSBSim_utils.py_ file.

That's incorrect. In Python [`sys.argv[0]` is the script name](https://docs.python.org/3/library/sys.html#sys.argv) so `tests/JSBSim_utils.py` builds the relative path from the script directory so that it can be executed from **any** folder. For example, if I run `python ../../tests/TestSwitch.py` from `build/tests` where the Python module has been compiled then the test succeeds (`sys.argv[0]` contains `../../tests/TestSwitch.py`). So your change does nothing since `JSBSim_utils.py` and all the tests scripts are located in the same directory.
> I added the `DoLinearization(...)` function to the C++ `FGFDMExec` class. The only thing this does, is create a new instance of the `FGLinearization` class and write the results to the referenced vectors.

Your method `FGFDMExec::DoLinearization()` is way too complicated with a lot of cryptic arguments. Why don't you return an instance of `FGLinearization` instead ?
> I added the `DoLinearization(...)` function to the C++ `FGFDMExec` class. The only thing this does, is create a new instance of the `FGLinearization` class and write the results to the referenced vectors.

There again that's way too complicated. You'd better create a Python interface to `FGLinearization`. There are quite a number of examples in `jsbsim.pxd.in`.
> Since these new functions seem to return quite a lot of items I'm considering letting them return the `FGLinearization` object instead. Any objections?

Yeah, that would be much better.
> An example on how to use this function can be found in the `tests/TestLinearization.py` unittest.

Unfortunately, [your example fails with a SEGFAULT](https://github.com/JSBSim-Team/jsbsim/pull/370/checks#step:10:757) (and so are the tests `RunCheckCases`, `CheckDebugLvl`, and `TestChannelRate`). These failures must be fixed for your PR to be merged.

bcoconni:
> I will push that change separately from your PR because it is not linked to the PR purpose.

Done in commit 2d84c4182712432ae3402e3ccee771b5dc04e471

aarondewindt:
> That's incorrect. In Python sys.argv[0] is the script name so tests/JSBSim_utils.py builds the relative path from the script directory so that it can be executed from any folder. For example, if I run python ../../tests/TestSwitch.py from build/tests where the Python module has been compiled then the test succeeds (sys.argv[0] contains ../../tests/TestSwitch.py). So your change does nothing since JSBSim_utils.py and all the tests scripts are located in the same directory.

@bcoconni Sorry, I had assumed it was started from the test directory. The problem I was having is that the script Clion uses to start the unittests is located within the its install directory, far away from the JSBSim files. So `sys.argv[0]` was pointing to this script within the Clion files.
What my code is doing is starting with the path to `tests/JSBSim_utils.py` and go one directory up to find the repository root. The reason is because the path to that file should always be known and remain unchanged regardless from where the test is started. The current code assumes it's started from a script located one directory down from the root.

> Yeah, that would be much better.

Ok, I'll do that then.

> Unfortunately, your example fails with a SEGFAULT (and so are the tests RunCheckCases, CheckDebugLvl, and TestChannelRate). These failures must be fixed for your PR to be merged.

I'm not sure why this is happening in the CI/CD, but I'll figure it out. For now my focus is on checking the results of this function though. Based on the conversation in issue #227 the linearization may not be perfect, so I want to make sure that it's at least useful.

bcoconni:
You said
> my code [...] is starting with the path to `tests/JSBSim_utils.py` and go one directory up to find the repository root

and
> The current code assumes it's started from a script located one directory down from the root.

Given that `JSBSim.py` is located in the **same** folder than the test scripts and unless I'm mistaken, these 2 statements are equivalent. I am reluctant to hardcode the location of `JSBSim_utils.py` to being one directory down from the root. Unlike the test scripts, `JSBSim_utils.py` is meant to be used from anywhere (see the Jupyter notebook example [doc/basic/00_JSBSim_sandbox.ipynb](https://github.com/JSBSim-Team/jsbsim/blob/master/doc/basic/00_JSBSim_sandbox.ipynb) for instance)

Actually your problem is coming from the fact that:
> the script Clion uses to start the unittests is located within the its install directory

Could you please provide the exact command that Clion uses to launch the tests ?
> Since these new functions seem to return quite a lot of items I'm considering letting them return the `FGLinearization` object instead. Any objections?

I slept over it and I am wondering why you the method `FGFDMExec::DoLinearization` at all ? I  mean you could make a Python interface that would allow to use the `FGLinearization` object like this:
```python
fdm = jsbsim.FGFDMExec('.')

lin = jsbsim.FGLinearization(fdm)
A, B, C, D = lin.GetStateSpace(x0, u0, y0)
state_names = lin.GetStateNames()
input_names = lin.GetInputNames()
output_names = lin.GetOutputNames()
state_units = lin.GetStateUnits()
input_units = lin.GetInputUnits()
output_units = lin.GetOutputUnits()
```

bcoconni:
> I'm not sure why this is happening in the CI/CD, but I'll figure it out. For now my focus is on checking the results of this function though. Based on the conversation in issue #227 the linearization may not be perfect, so I want to make sure that it's at least useful.

This is not limited to the CI/CD since I get the same failures when building your branch locally on my PC (running `ctest` from the build folder). In general, when the 3 platforms Linux, Windows and MacOSX it really means something is wrong with the code :smile:

aarondewindt:
> Given that JSBSim.py [..] for instance)

Ahhh, that makes now. The command Clion uses to for example start my test is:
`/home/adewindt/miniconda3/envs/jsbsim_dev/bin/python /home/adewindt/.local/share/JetBrains/Toolbox/apps/CLion/ch-0/203.5981.106/plugins/python-ce/helpers/pycharm/_jb_unittest_runner.py --target TestLinearization.TestLinearization.test_do_linearization`

So a bit clearer:
`<python> <_jb_unittest_runner.py> --target TestLinearization.TestLinearization.test_do_linearization`


> I slept over it and I am wondering why you need the method FGFDMExec::DoLinearization at all ? I mean you could make a Python interface that would allow to use the FGLinearization object like this:

Makes sense, I'll change it do that.

aarondewindt:
Ok, I'm getting the segfault by running ctest. I'll look for the mistake.

aarondewindt:
Nvm, my Clion unittesting issue. Running the tests through ctest fixes it.

bcoconni:
> Nvm, my Clion unittesting issue. Running the tests through ctest fixes it.

That's fortunate because when I saw the command line issued by Clion I told myself that we'll have to be very creative to handle that case :smile:

aarondewindt:
Ok, the tests are passing and I'm satisfied with the interface so far. If you want you can merge it. For now I'll be using these functions and do some checks. I'll create a new issue and PR if I come across anything that needs to be fixed.

src/initialization/FGLinearization.h
```diff
@@ -19,25 +19,74 @@
...
+    /**
+     * Get the state space model matrices.
+     *
+     * @param A_ System matrix
+     * @param B_ Input matrix
+     * @param C_ Output matrix
+     * @param D_ Feedforward matrix
+     */
+    void GetStateSpace(Vector2D<double> & A_, Vector2D<double> & B_,
+                       Vector2D<double> & C_, Vector2D<double> & D_) const;
```
bcoconni, line 76: I would split `GetStateSpace` in 4 separate methods rather than modifying the method parameters which I always find awkward.
In addition the 4 method names would document what is the returned matrix rather than using obscure names `A_`, `B_`, `C_` and `D_`.
```suggestion
    const Vector2D<double>& GetSystemMatrix(void) const { return A; }
    const Vector2D<double>& GetInputMatrix(void) const { return B; }
    const Vector2D<double>& GetOutputMatrix(void) const { return C; }
    const Vector2D<double>& GetFeedforwardMatrix(void) const { return D; }
```

src/initialization/FGLinearization.cpp
```diff
@@ -69,47 +65,88 @@ FGLinearization::FGLinearization(FGFDMExec * fdm, int mode)
...
+    return x_units;
+}
+
+std::vector<std::string> FGLinearization::GetInputUnits() const {
+    return u_units;
+}
+
+std::vector<std::string> FGLinearization::GetOutputUnits() const {
+    return y_units;
+}
```
bcoconni, line 126: Such trivial methods should be inlined in the header. I suggest to remove all of them from `FGLinearization.cpp`
```suggestion
```

src/initialization/FGLinearization.h
```diff
@@ -19,25 +19,74 @@
...
+    std::vector<double> GetInitialInput() const;
+    std::vector<double> GetInitialOutput() const;
+
+    std::vector<std::string> GetStateNames() const;
+    std::vector<std::string> GetInputNames() const;
+    std::vector<std::string> GetOutputNames() const;
+
+    std::vector<std::string> GetStateUnits() const;
+    std::vector<std::string> GetInputUnits() const;
+    std::vector<std::string> GetOutputUnits() const;
```
bcoconni, line 88: Trivial methods like that must be inlined in the header and return a const reference to avoid calling the copy constructor whenever possible (note that the `inline` keyword is not needed for the compiler to inline methods)
```suggestion
    const std::vector<double>& GetInitialState(void) const { return x0; }
    const std::vector<double>& GetInitialInput(void) const { return u0; }
    const std::vector<double>& GetInitialOutput(void) const { return y0; }

    const std::vector<std::string>& GetStateNames(void) const { return x_names; }
    const std::vector<std::string>& GetInputNames(void) const { return u_names; }
    const std::vector<std::string>& GetOutputNames(void) const { return y_names; }

    const std::vector<std::string>& GetStateUnits(void) const { return x_units; }
    const std::vector<std::string>& GetInputUnits(void) const { return u_units; }
    const std::vector<std::string>& GetOutputUnits(void) const { return y_units; } 
```

python/jsbsim.pxd
```diff
@@ -30,6 +31,29 @@ cdef extern from "initialization/FGInitialCondition.h" namespace "JSBSim":
...
 
+cdef extern from "initialization/FGLinearization.h" namespace "JSBSim":
+    cdef cppclass c_FGLinearization "JSBSim::FGLinearization":
+        c_FGLinearization(c_FGFDMExec* fdme)
+        
+        void WriteScicoslab() const
+        void WriteScicoslab(string& path) const
+
+        void GetStateSpace(vector[vector[double]]& A_, vector[vector[double]]& B_,
+                           vector[vector[double]]& C_, vector[vector[double]]& D_) const
```
bcoconni, line 42: Replace this method with 4 separate and more explicitly named methods `GetSystemMatrix`, `GetInputMatrix`, `GetOutputMatrix` and `GetFeedforwardMatrix` as suggested in the header `FGLinearization.h` below.
```suggestion
        vector[vector[double]]& GetSystemMatrix() const
        vector[vector[double]]& GetInputMatrix() const
        vector[vector[double]]& GetOutputMatrix() const
        vector[vector[double]]& GetFeedforwardMatrix() const
```

src/initialization/FGLinearization.cpp
```diff
@@ -20,23 +20,19 @@
...
-
+    auto ss = FGStateSpace(fdm);
     ss.x.add(new FGStateSpace::Vt);
     ss.x.add(new FGStateSpace::Alpha);
     ss.x.add(new FGStateSpace::Theta);
     ss.x.add(new FGStateSpace::Q);
 
     // get propulsion pointer to determine type/ etc.
-    FGEngine * engine0 = fdm->GetPropulsion()->GetEngine(0);
+    auto engine0 = fdm->GetPropulsion()->GetEngine(0);
```
bcoconni, line 35: What for ? Current code doesn't hurt: I suggest to keep the existing code.
```suggestion
    FGEngine * engine0 = fdm->GetPropulsion()->GetEngine(0);
```

.gitignore
```diff
@@ -33,6 +33,11 @@ src/.libs/
 TAGS
 .dir-locals.el
 
+#################
+## Python
+#################
+__pycache__/
+
```
bcoconni, line 40: FYI, this change has already been committed to `master` (commit 2d84c4182712432ae3402e3ccee771b5dc04e471) so it will be ignored when your PR will be merged.

.gitignore
```diff
@@ -75,3 +80,11 @@ ipch/
...
 #################
 .DS_Store
+
+#################
+## Clion (Jetbrains IDEs)
+#################
+# Project files
+.idea/*
+# CMake
+cmake-build-*/
```
bcoconni, line 90: This change has also already been committed to `master` (commit 2d84c4182712432ae3402e3ccee771b5dc04e471) so it will be ignored when your PR will be merged.

bcoconni (changes requested):
Overall it looks fine to me.

However I have requested a number of minor changes:
- Split the C++ method `FGLinearization::GetStateSpace` in 4 smaller, simpler methods with more explicit names. This helps documenting the code and I dislike functions/methods that modify their parameters :wink: 
- Systematically calling `self.__intercept_invalid_pointer()` before dereferencing it in `jsbsim.pyx.in`.
- Trivial C++ methods should be inlined
- Trivial C++ methods should return a const reference rather than returning by value
- Simplify the test `TestLinearization.py` there is a lot of boilerplate code that is really not needed here.

Once these modifications will be applied, I will merge your PR.
Thanks again for your contribution.

src/initialization/FGLinearization.cpp
aarondewindt: I changed that because the `GetEngine()` function has changed since this code was last compiled. I think it now returns a smart pointer. Changing it to `auto` was a fast fix when I was trying to get the code to compile. I should've changed it to the new type afterwards though, but forgot.

src/initialization/FGLinearization.cpp
```diff
@@ -20,23 +20,19 @@
 #include "FGLinearization.h"
 #include <ctime>
 
-namespace JSBSim {
 
-// TODO make FGLinearization have X,U,Y selectable by xml config file
+namespace JSBSim {
 
-FGLinearization::FGLinearization(FGFDMExec * fdm, int mode)
+FGLinearization::FGLinearization(FGFDMExec * fdm) : aircraft_name(fdm->GetAircraft()->GetAircraftName())
```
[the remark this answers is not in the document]
aarondewindt: Ahhh, I assumed a 120 columns limit.

src/initialization/FGLinearization.cpp
```diff
@@ -20,23 +20,19 @@
...
 {
-    std::cout << "\nlinearization: " << std::endl;
-    std::clock_t time_start=clock(), time_linDone;
-    FGStateSpace ss(fdm);
-
+    auto ss = FGStateSpace(fdm);
```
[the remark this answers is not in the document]
aarondewindt: I originally changed it so that the FGStateSpace instance was a member of the class so that FGLinearization can access it later on. This was eventually not necessary so I changed it "back". I didn't check what was originally there when I started.

tests/TestLinearization.py
```diff
@@ -0,0 +1,49 @@
...
+                                         aircraft_name+'.xml'))
+
+        IC_file = self.sandbox('aircraft', aircraft_name, 'cruise_init.xml')
+        tree = et.parse(IC_file)
+        heading_el = tree.find('psi')
+        heading_el.text = '270.0'
+        tree.write(IC_file)
+
+        fdm = self.create_fdm()
+        fdm.set_aircraft_path(self.sandbox('aircraft'))
```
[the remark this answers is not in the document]
aarondewindt: I wanted a working FDM to test with so I just copy pasted what I thought was the "initialization" of one of the tests. I'll clean the test up.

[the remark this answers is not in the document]
aarondewindt: Isn't this necessary to initialize the fdm in a trimmed condition? I assume the `cruise_init.xml` sets the 737 model IC in a trimmed condition. I wasn't sure what was required to get the IC to be loaded properly. I haven't had much experience with JSBSim yet.

aarondewindt:
I'll make the changes in a day or two.

src/initialization/FGLinearization.cpp
bcoconni: Yes, indeed. I forgot that the list of engines in `FGPropulsion` has moved to smart pointers. No problem in using `auto` in that particular case.

tests/TestLinearization.py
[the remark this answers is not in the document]
bcoconni: The code that I suggest you remove is intended to modify the XML file that describes the initial conditions (IC) before it is read by JSBSim. If you want to trim the aircraft you just need to set the property `simulation/do_simple_trim` to 1 with the following code:
```python
fdm['simulation/do_simple_trim'] = 1
```
However, given that your test only checks the matrices shapes it doesn't make much difference whether or not the aircraft is trimmed.

bcoconni:
> I'll make the changes in a day or two.

OK. No rush.

src/initialization/FGLinearization.cpp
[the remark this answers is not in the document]
bcoconni: > Ahhh, I assumed a 120 columns limit.

JSBSim is 20+ years old : we were using CRTs computer monitor at the time 😄

src/initialization/FGLinearization.cpp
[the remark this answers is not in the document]
bcoconni: No problem. That's what code review are made for 😉

bcoconni (approved):
Looks good to me

PR #382, opened 2021-01-26
Tsfc function

seanmcleod:
Allow TSFC and ATSFC to be specified as functions for an engine model.

### DIFF
--- a/src/models/propulsion/FGTurbine.cpp
+++ b/src/models/propulsion/FGTurbine.cpp
@@ -69,7 +69,7 @@ FGTurbine::FGTurbine(FGFDMExec* exec, Element *el, int engine_number, struct Inp
   MaxN1 = MaxN2 = 100.0;
   Augmented = AugMethod = Injected = 0;
   BypassRatio = BleedDemand = 0.0;
-  IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = 0;
+  IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = TSFCLookup = ATSFCLookup = nullptr;
   N1_spinup = 1.0; N2_spinup = 3.0; IgnitionN1 = 5.21; IgnitionN2 = 25.18; N1_start_rate = 1.4; N2_start_rate = 2.0; 
   N1_spindown = 2.0; N2_spindown = 2.0;
   InjectionTime = 30.0;
--- a/src/models/propulsion/FGTurbine.cpp
+++ b/src/models/propulsion/FGTurbine.cpp
@@ -94,6 +94,9 @@ FGTurbine::~FGTurbine()
   if (dynamic_cast<FGSpoolUp*>(N2SpoolUp)) delete N2SpoolUp;
   if (dynamic_cast<FGSpoolUp*>(N2SpoolDown)) delete N2SpoolDown;
 
+  delete TSFC;
+  delete ATSFC;
+
   Debug(1);
 }
 
--- a/src/models/propulsion/FGTurbine.cpp
+++ b/src/models/propulsion/FGTurbine.cpp
@@ -234,6 +234,12 @@ double FGTurbine::Run()
   OilPressure_psi = N2 * 0.62;
   OilTemp_degK = Seek(&OilTemp_degK, 366.0, 1.2, 0.1);
 
+  // Check if there are functions for TSFC and ATSFC
+  if (TSFCLookup)
+    TSFC = TSFCLookup->GetValue();
+  if (ATSFCLookup)
+    ATSFC = ATSFCLookup->GetValue();
+
   if (!Augmentation) {
     correctedTSFC = TSFC * sqrt(T/389.7) * (0.84 + (1-N2norm)*(1-N2norm));
     FuelFlow_pph = Seek(&FuelFlow_pph, thrust * correctedTSFC, 1000.0, 10000.0);
@@ -434,7 +440,7 @@ bool FGTurbine::Load(FGFDMExec* exec, Element *el)
     string name = function_element->GetAttributeValue("name");
     if (name == "IdleThrust" || name == "MilThrust" || name == "AugThrust"
         || name == "Injection" || name == "N1SpoolUp" || name == "N1SpoolDown"
-        || name == "N2SpoolUp" || name == "N2SpoolDown")
+        || name == "N2SpoolUp" || name == "N2SpoolDown" || name == "TSFC" || name == "ATSFC")
       function_element->SetAttributeValue("name", string("propulsion/engine[#]/") + name);
 
     function_element = el->FindNextElement("function");
@@ -502,6 +508,9 @@ bool FGTurbine::Load(FGFDMExec* exec, Element *el)
   MilThrustLookup = GetPreFunction(property_prefix+"/MilThrust");
   MaxThrustLookup = GetPreFunction(property_prefix+"/AugThrust");
   InjectionLookup = GetPreFunction(property_prefix+"/Injection");
+
+  TSFCLookup = GetPreFunction(property_prefix+"/TSFC");
+  ATSFCLookup = GetPreFunction(property_prefix+"/ATSFC");
 
   // Pre-calculations and initializations
   N1SpoolUp = GetPreFunction(property_prefix+"/N1SpoolUp");
--- a/src/models/propulsion/FGTurbine.cpp
+++ b/src/models/propulsion/FGTurbine.cpp
@@ -44,6 +44,7 @@ INCLUDES
 
 #include "FGFDMExec.h"
 #include "math/FGFunction.h"
+#include "math/FGRealValue.h"
 #include "FGTurbine.h"
 #include "FGThruster.h"
 #include "input_output/FGXMLElement.h"
@@ -62,14 +63,14 @@ FGTurbine::FGTurbine(FGFDMExec* exec, Element *el, int engine_number, struct Inp
   Type = etTurbine;
 
   MilThrust = MaxThrust = 10000.0;
-  TSFC = 0.8;
-  ATSFC = 1.7;
+  TSFC = new FGRealValue(0.8);
+  ATSFC = new FGRealValue(1.7);
   IdleN1 = 30.0;
   IdleN2 = 60.0;
   MaxN1 = MaxN2 = 100.0;
   Augmented = AugMethod = Injected = 0;
   BypassRatio = BleedDemand = 0.0;
-  IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = TSFCLookup = ATSFCLookup = nullptr;
+  IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = nullptr;
   N1_spinup = 1.0; N2_spinup = 3.0; IgnitionN1 = 5.21; IgnitionN2 = 25.18; N1_start_rate = 1.4; N2_start_rate = 2.0; 
   N1_spindown = 2.0; N2_spindown = 2.0;
   InjectionTime = 30.0;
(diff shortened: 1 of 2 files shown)


src/models/propulsion/FGTurbine.h
```diff
@@ -311,7 +311,9 @@ class FGTurbine : public FGEngine
   FGFunction *MilThrustLookup;
   FGFunction *MaxThrustLookup;
   FGFunction *InjectionLookup;
-  FGFDMExec* FDMExec;
+  FGFunction *TSFCLookup;
+  FGFunction *ATSFCLookup;
```
bcoconni, line 315: I'd rather replace the members `TSFC` and `ATSFC` by `FGParameter*` members

src/models/propulsion/FGTurbine.cpp
```diff
@@ -69,7 +69,7 @@ FGTurbine::FGTurbine(FGFDMExec* exec, Element *el, int engine_number, struct Inp
   MaxN1 = MaxN2 = 100.0;
   Augmented = AugMethod = Injected = 0;
   BypassRatio = BleedDemand = 0.0;
-  IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = 0;
+  IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = TSFCLookup = ATSFCLookup = nullptr;
```
bcoconni, line 72: I'd rather replace the `TSFC` and `ATSFC` members by `FGParameter*`
```suggestion
  IdleThrustLookup = MilThrustLookup = MaxThrustLookup = InjectionLookup = nullptr;
  TSFC = new FGRealValue(0.8);
  ATSFC = new FGRealValue(1.7);
```

src/models/propulsion/FGTurbine.cpp
```diff
@@ -234,6 +234,12 @@ double FGTurbine::Run()
   OilPressure_psi = N2 * 0.62;
   OilTemp_degK = Seek(&OilTemp_degK, 366.0, 1.2, 0.1);
 
+  // Check if there are functions for TSFC and ATSFC
+  if (TSFCLookup)
+    TSFC = TSFCLookup->GetValue();
+  if (ATSFCLookup)
+    ATSFC = ATSFCLookup->GetValue();
+
```
bcoconni, line 242: The point of making `TSFC` and `ATSFC` pointers of type `FGParameter*` is that you no longer need to test if `TSFCLookup` and `ATSFCLookup` exist and just call `GetValue()` on them:
```suggestion
  double _TSFC = TSFC->GetValue();
  double _ATSFC = ATSFC->GetValue();
```

src/models/propulsion/FGTurbine.cpp
```diff
@@ -434,7 +440,7 @@ bool FGTurbine::Load(FGFDMExec* exec, Element *el)
     string name = function_element->GetAttributeValue("name");
     if (name == "IdleThrust" || name == "MilThrust" || name == "AugThrust"
         || name == "Injection" || name == "N1SpoolUp" || name == "N1SpoolDown"
-        || name == "N2SpoolUp" || name == "N2SpoolDown")
+        || name == "N2SpoolUp" || name == "N2SpoolDown" || name == "TSFC" || name == "ATSFC")
```
bcoconni, line 443: Don't use "pre-functions". Instead re-use the existing `tsfc` and `atsfc` attributes (lines 455-458 of `FGTurbine.cpp`)
```c++
  if (el->FindElement("tsfc")) {
    string value = el->GetDataLine();
    if (is_number(value))
      TSFC = new FGRealValue(atof(value.c_str()));
    else
      TSFC = new FGFunction(FDMExec, el);
  }
  if (el->FindElement("atsfc")) {
    string value = el->GetDataLine();
    if (is_number(value))
      ATSFC = new FGRealValue(atof(value.c_str()));
    else
      ATSFC = new FGFunction(FDMExec, el);
  }
```

src/models/propulsion/FGTurbine.cpp
```diff
@@ -503,6 +509,9 @@ bool FGTurbine::Load(FGFDMExec* exec, Element *el)
   MaxThrustLookup = GetPreFunction(property_prefix+"/AugThrust");
   InjectionLookup = GetPreFunction(property_prefix+"/Injection");
 
+  TSFCLookup = GetPreFunction(property_prefix+"/TSFC");
+  ATSFCLookup = GetPreFunction(property_prefix+"/ATSFC");
+
```
bcoconni, line 514: Don't use "prefunctions" as mentioned above.
```suggestion
```

bcoconni (changes requested):
I would implement the feature slightly differently. Rather than having a parameter `<tsfc>` and a lookup function, I would merge the 2 into one using the `FGParameter*` class.

Hopefully the comments I made are clear enough to expose my point of view.

seanmcleod:
All good suggestions. I'll implement them, test and then push another commit.

Just wondering whether I should add in logic to bypass the TSFC correction if the user supplies a function now, or leave that for a follow-up pull request.

bcoconni:
> Just wondering whether I should add in logic to bypass the TSFC correction if the user supplies a function now, or leave that for a follow-up pull request.

My suggestion would rather be to address them separately (i.e. 2 different pull requests) that will make things simpler to discuss.

bcoconni:
Following [your comment about making a reference to the current engine](https://github.com/JSBSim-Team/jsbsim/discussions/381#discussioncomment-320910), the code for the creation of `FGFunction` should be slightly modified to pass the engine number value to `FGFunction` so that the hash/sharp `#` character can be replaced:
```c++
  if (el->FindElement("tsfc")) {
    string value = el->GetDataLine();
    if (is_number(value))
      TSFC = new FGRealValue(atof(value.c_str()));
    else
      TSFC = new FGFunction(FDMExec, el, to_string((int)EngineNumber));
  }
  if (el->FindElement("atsfc")) {
    string value = el->GetDataLine();
    if (is_number(value))
      ATSFC = new FGRealValue(atof(value.c_str()));
    else
      ATSFC = new FGFunction(FDMExec, el, to_string((int)EngineNumber));
  }
```

seanmcleod:
Was getting very weird errors with the loading of engine model:

```sh
In file ./engine/CFM56.xml: line 15
Bad operation <milthrust> detected in configuration file

In file ./engine/CFM56.xml: line 16
Bad operation <bypassratio> detected in configuration file

....

In file ./engine/CFM56.xml: line 42
Bad operation <function> detected in configuration file

In file ./engine/CFM56.xml: line 14
Malformed function name with number: 0 and property name: CFM56 but no "#" sign for substitution.
    Function: CFM56
FATAL ERROR: JSBSim terminated with an unknown exception.
```

Took a while, but it turned out that

```c++
  if (el->FindElement("tsfc")) {
    string value = el->GetDataLine();
```

Needed to be changed to:

```c++  
   if (el->FindElement("tsfc")) {
    string value = el->FindElement("tsfc")->GetDataLine();
```

seanmcleod:
Last gotcha for me :wink:

Wasn't sure why the following wasn't working as before.

```xml
  <function name="tsfc">
    <table>
      <independentVar lookup="row">velocities/mach</independentVar>
      <independentVar lookup="column">atmosphere/density-altitude</independentVar>
      <tableData>
             -10000     0     10000   20000   30000   40000   50000   60000
        0.0  0.2        0.3   0.4     0.5     0.6     0.7     0.8     0.9
        0.2  0.3        0.4   0.5     0.6     0.7     0.8     0.9     1.0
        0.4  0.4        0.5   0.6     0.7     0.8     0.9     1.0     1.1
        0.6  0.5        0.6   0.7     0.8     0.9     1.0     1.1     1.2
        0.8  0.6        0.7   0.8     0.9     1.0     1.1     1.2     1.3
        1.0  0.7        0.8   0.9     1.0     1.1     1.2     1.3     1.4
      </tableData>
    </table>
  </function>
```

Was sort of assuming the source code changes to `FGParameter*` would result in `FindElement("tsfc")` finding the function with that name.

But it looks like this is required, I had initially had a `<function>` element inside the `<tsfc>` element as well.

```xml
  <tsfc>
    <table>
      <independentVar lookup="row">velocities/mach</independentVar>
      <independentVar lookup="column">propulsion/engine[#]/thrust-lbs</independentVar>
      <tableData>
        -10000     0     10000   20000   30000   40000   50000   60000
        0.0  0.2        0.3   0.4     0.5     0.6     0.7     0.8     0.9
        0.2  0.3        0.4   0.5     0.6     0.7     0.8     0.9     1.0
        0.4  0.4        0.5   0.6     0.7     0.8     0.9     1.0     1.1
        0.6  0.5        0.6   0.7     0.8     0.9     1.0     1.1     1.2
        0.8  0.6        0.7   0.8     0.9     1.0     1.1     1.2     1.3
        1.0  0.7        0.8   0.9     1.0     1.1     1.2     1.3     1.4
      </tableData>
    </table>
  </tsfc>    
```

Tested it using both `atmosphere/density-altitude` and `propulsion/engine[#]/thrust-lbs`.

src/models/propulsion/FGTurbine.cpp
```diff
@@ -503,6 +500,26 @@ bool FGTurbine::Load(FGFDMExec* exec, Element *el)
   MaxThrustLookup = GetPreFunction(property_prefix+"/AugThrust");
   InjectionLookup = GetPreFunction(property_prefix+"/Injection");
 
+  JSBSim::Element* tsfcElement = el->FindElement("tsfc");
+  if (tsfcElement) {
+    delete TSFC;
```
bcoconni, line 505: It is customary to protect such calls with a test that checks that `TSFC` is not a `nullptr` because, with exceptions, you never know... And the check doesn't hurt anyway
```suggestion
    if (TSFC) {
      delete TSFC;
      TSFC = nullptr;
    }
```
At some point, we'll need to replace all the `new` and `delete` in this class with `std::unique_ptr<>`

src/models/propulsion/FGTurbine.cpp
```diff
@@ -503,6 +500,26 @@ bool FGTurbine::Load(FGFDMExec* exec, Element *el)
...
+    string value = tsfcElement->GetDataLine();
+    if (is_number(value))
+      TSFC = new FGRealValue(atof(value.c_str()));
+    else
+      TSFC = new FGFunction(FDMExec, tsfcElement, to_string((int)EngineNumber));
+  }
+
+  JSBSim::Element* atsfcElement = el->FindElement("atsfc");
+  if (atsfcElement) {
+    delete ATSFC;
```
bcoconni, line 515: Same as above
```suggestion
    if (ATSFC) {
      delete ATSFC;
      ATSFC = nullptr;
    }
```

bcoconni (changes requested):
Looks good to me.

The pointers `TSFC` and `ATSFC` need to be deleted in `FGTurbine` destructor otherwise we'll have dangling pointers.
```c++
  if (TSFC) delete TSFC;
  if (ATSFC) delete ATSFC;
```

bcoconni:
Sorry for having misled you with my untested code by the way :confused:

src/models/propulsion/FGTurbine.cpp
seanmcleod: Definitely missed adding the deletes in the destructor. And yep would be better/easier with `std::unique_ptr<>`.

In terms of the `nullptr` check before the call to `delete`, it may be customary but also old-fashioned :wink:

https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null

seanmcleod:
> Sorry for having misled you with my untested code by the way

No problem, I was a bit lazy blindingly using it without double-checking, but the weird errors threw for a while thinking I'd messed something up in the engine xml.

src/models/propulsion/FGTurbine.cpp
bcoconni: > In terms of the `nullptr` check before the call to `delete`, it may be customary but also old-fashioned :wink:

Oops ! You are absolutely right :smile: Forget the check.

bcoconni (approved):

agodemar:
@seanmcleod @bcoconni thank you guys for working in this!

PR #387, opened 2021-01-30
For user supplied TSFC functions don't perform correction/denormalisation

seanmcleod:


src/models/propulsion/FGTurbine.cpp
```diff
@@ -220,7 +220,11 @@ double FGTurbine::Run()
   OilTemp_degK = Seek(&OilTemp_degK, 366.0, 1.2, 0.1);
 
   if (!Augmentation) {
-    correctedTSFC = TSFC->GetValue() * sqrt(T/389.7) * (0.84 + (1-N2norm)*(1-N2norm));
+    // If TSFC function is supplied don't perform any correction/denormalisation
+    if (dynamic_cast<FGFunction*>(TSFC))
+      correctedTSFC = TSFC->GetValue();
+    else
+      correctedTSFC = TSFC->GetValue() * sqrt(T/389.7) * (0.84 + (1-N2norm)*(1-N2norm));
```
bcoconni, line 227: I'd rather create a new class `FGTSFC` to specialize `GetValue()` with the expression above. Similar to what I did for `FGSpoolUp`
https://github.com/JSBSim-Team/jsbsim/blob/a00129aa73c87d5322ca086aa9ecbf26c0be4cc5/src/models/propulsion/FGTurbine.h#L325-L340
The point being that I think the fuel flow management at idle will need to move to that class as well.

bcoconni (changes requested):
Better create a new class deriving from `FGParameter` than doing an `if/else`.

src/models/propulsion/FGTurbine.h
```diff
@@ -337,6 +341,34 @@ class FGSpoolUp : public FGParameter
...
+
+class FGTSFC : public FGParameter
+{
+public:
+  FGTSFC(FGTurbine* _turb, double tsfcVal)
+    : turb(_turb), tsfc(new FGRealValue(tsfcVal)) {}
+
+  FGTSFC(FGTurbine* _turb, FGFDMExec* fdmexec, Element* element, int enginenumber)
+    : turb(_turb), tsfc(new FGFunction(fdmexec, element, to_string(enginenumber))) {}
+
```
bcoconni, line 353: Sorry, the idea was to leave the `FGFunction` case outside of the class.
```suggestion
```

src/models/propulsion/FGTurbine.cpp
seanmcleod: Just pushed a commit to go with the suggested approach.

src/models/propulsion/FGTurbine.h
```diff
@@ -337,6 +341,34 @@ class FGSpoolUp : public FGParameter
...
+  double GetValue(void) const {
+    // If there is a user supplied function don't perform any correction/denormalisation
+    if(dynamic_cast<FGFunction*>(tsfc.get()))
+      return tsfc->GetValue();
+    else {
+      double T = turb->in.Temperature;
+      double N2norm = turb->N2norm;
+      return tsfc->GetValue() * sqrt(T / 389.7) * (0.84 + (1 - N2norm) * (1 - N2norm));
+    }
```
bcoconni, line 364: Just manage the case where TSFC is not specified as a function:
```suggestion
    double T = turb->in.Temperature;
    double N2norm = turb->N2norm;
    return tsfc * sqrt(T / 389.7) * (0.84 + (1 - N2norm) * (1 - N2norm));
```

src/models/propulsion/FGTurbine.h
```diff
@@ -337,6 +341,34 @@ class FGSpoolUp : public FGParameter
...
+  }
+
+private:
+  FGTurbine* turb;
+  std::unique_ptr<FGParameter> tsfc;
```
bcoconni, line 369:
```suggestion
  double tsfc;
```

src/models/propulsion/FGTurbine.h
```diff
@@ -337,6 +341,34 @@ class FGSpoolUp : public FGParameter
   FGTurbine* turb;
   double delay; ///< Inverse spool-up time from idle to 100% (seconds)
 };
+
+class FGTSFC : public FGParameter
+{
+public:
+  FGTSFC(FGTurbine* _turb, double tsfcVal)
+    : turb(_turb), tsfc(new FGRealValue(tsfcVal)) {}
```
bcoconni, line 349: This class just need to manage the case where `<tsfc>` is specified as a number. So you don't need `FGRealValue`.
```suggestion
  FGTSFC(FGTurbine* _turb, double tsfcVal)
    : turb(_turb), tsfc(tsfcVal) {}
```

src/models/propulsion/FGTurbine.h
```diff
@@ -246,7 +249,7 @@ class FGTurbine : public FGEngine
   double MilThrust;        ///< Maximum Unaugmented Thrust, static @ S.L. (lbf)
   double MaxThrust;        ///< Maximum Augmented Thrust, static @ S.L. (lbf)
   double BypassRatio;      ///< Bypass Ratio
-  std::unique_ptr<FGParameter> TSFC;   ///< Thrust Specific Fuel Consumption (lbm/hr/lbf)
+  std::unique_ptr<FGTSFC> TSFC;        ///< Thrust Specific Fuel Consumption (lbm/hr/lbf)
```
bcoconni, line 252: Keep the type of member `TSFC` as `FGParameter` because it will contain either an `FGFunction` or a `FGTSFC` instance.
```suggestion
  std::unique_ptr<FGParameter> TSFC;   ///< Thrust Specific Fuel Consumption (lbm/hr/lbf)
```

src/models/propulsion/FGTurbine.h
```diff
@@ -41,6 +41,9 @@ INCLUDES
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
 #include "FGEngine.h"
+#include "math/FGFunction.h"
+#include "math/FGRealValue.h"
+
```
bcoconni, line 46: You don't need to include these headers because the class `FGTSFC` will store the `<tsfc>` as a `double`
```suggestion
```

src/models/propulsion/FGTurbine.h
```diff
@@ -49,7 +52,7 @@ FORWARD DECLARATIONS
 namespace JSBSim {
 
 class Element;
-class FGFunction;
+class FGTSFC;
```
bcoconni, line 55: Once the headers, will be removed you'll need to restore the declaration of `class FGFunction;`
```suggestion
class FGFunction;
```

src/models/propulsion/FGTurbine.cpp
```diff
@@ -488,9 +485,9 @@ bool FGTurbine::Load(FGFDMExec* exec, Element *el)
   if (tsfcElement) {
     string value = tsfcElement->GetDataLine();
     if (is_number(value))
-      TSFC = std::make_unique<FGRealValue>(atof(value.c_str()));
+      TSFC = std::make_unique<FGTSFC>(this, atof(value.c_str()));
     else
-      TSFC = std::make_unique<FGFunction>(FDMExec, tsfcElement, to_string((int)EngineNumber));
+      TSFC = std::make_unique<FGTSFC>(this, FDMExec, tsfcElement, EngineNumber);
```
bcoconni, line 490: Just use the plain `FGFunction` class in that case.
```suggestion
      TSFC = std::make_unique<FGFunction>(FDMExec, tsfcElement, to_string((int)EngineNumber));
```

src/models/propulsion/FGTurbine.cpp
```diff
@@ -592,7 +589,7 @@ void FGTurbine::bindmodel(FGPropertyManager* PropertyManager)
   PropertyManager->Tie( property_name.c_str(), this,
                         &FGTurbine::GetInjN2increment, &FGTurbine::SetInjN2increment);
   property_name = base_property_name + "/tsfc";
-  PropertyManager->Tie(property_name.c_str(), TSFC.get(), &FGParameter::GetValue);
+  PropertyManager->Tie(property_name.c_str(), &correctedTSFC);
```
bcoconni, line 592: May be keep the old way so that the property is read only ?
```suggestion
  PropertyManager->Tie(property_name.c_str(), TSFC.get(), &FGParameter::GetValue);
```

bcoconni (changes requested):
Sorry I did not make myself clear. The idea was to get rid of the test:
```c++
  if (dynamic_cast<FGFunction>) {
    ...
  } else {
    ...
  }
```
The member `TSFC` should either be an `FGFunction` or an `FGTSFC` instance. The C++ inheritance mechanism will automatically resolve which `GetValue()` should be called based upon the `TSFC` type:
* If `TSFC` is an `FGFunction` then `FGFunction::GetValue()` will be called.
* If `TSFC` is an `FGTSFC` then `FGTSFC::GetValue()` will be called.

In your proposal you moved the `if (dynamic_cast<>)` in `FGTSFC` but really the point is to no longer have this `if` (it is the C++ compiler job to resolve it).

I've tried to illustrate the code modifications in the comments but as usual this untested code :confused:

seanmcleod:
I had understood your request, partly I guess based on the suggested name of `FGTSFC` to encapsulate TSFC in the new class to handle both cases of it being specified as a number with correction needed or a function version. Also the comment about adding additional logic for idle in future wasn't entirely clear if that would only apply to the one case, i.e. simple number version versus applying to both.

Trying to think of a more descriptive name for `FGTSFC` to indicate it applies specifically to the single tsfc value approach with correction.

src/models/propulsion/FGTurbine.cpp
seanmcleod: The reason I tied the property to the `correctedTSFC` member was so that there wouldn't be another function lookup when the property was requested for the case where a function was supplied for TSFC, i.e. return the cached version calculated in the last call to `run()`.

bcoconni:
> I had understood your request, partly I guess based on the suggested name of FGTSFC to encapsulate TSFC in the new class to handle both cases of it being specified as a number with correction needed or a function version.

Yes my suggestion was misleading. Sorry about that.
> Also the comment about adding additional logic for idle in future wasn't entirely clear if that would only apply to the one case, i.e. simple number version versus applying to both.

Regarding the fuel consumption at idle, my point is the same than for the temperature correction: the user who specifies TSFC with a table will most likely provide data that cover idling as well. In that case, that user doesn't want `FGTurbine` to take over the TSFC tables with the code:
https://github.com/JSBSim-Team/jsbsim/blob/a00129aa73c87d5322ca086aa9ecbf26c0be4cc5/src/models/propulsion/FGTurbine.cpp#L525
https://github.com/JSBSim-Team/jsbsim/blob/a00129aa73c87d5322ca086aa9ecbf26c0be4cc5/src/models/propulsion/FGTurbine.cpp#L225
So my idea is that these 2 lines of code will end up in the class `FGTSFC` (or whatever will its name) as well. The point being that this correction will only be applied when `<tsfc>` is specified with a number.
> Trying to think of a more descriptive name for `FGTSFC` to indicate it applies specifically to the single tsfc value approach with correction.

I guess something like `FGSimplifiedTSFC`, `FGBasicTSFCManagement` ?!? You are the native English speaker, I am sure you'll come with better names :smile:

seanmcleod:
> Yes my suggestion was misleading. Sorry about that.

No problem, re-reading it I can also understand your actual intent.

> You are the native English speaker, I am sure you'll come with better names

Coming up with useful names for variables, classes etc. one of the toughest things in programming 😉

src/models/propulsion/FGTurbine.cpp
seanmcleod: Ah, missed the read-only distinction. Originally I meant to make it read-only, and was used to only providing a getter and not a setter, so sort of assumed I was only passing a getter reference `&correctedTSFC`, but double-checking the `Tie` class I see this form enables read and write.

bcoconni: > The reason I tied the property to the `correctedTSFC` member was so that there wouldn't be another function lookup when the property was requested for the case where a function was supplied for TSFC, i.e. return the cached version calculated in the last call to `run()`.

Yes, that's a good point.
> Ah, missed the read-only distinction. Originally I meant to make it read-only, and was used to only providing a getter and not a setter, so sort of assumed I was only passing a getter reference `&correctedTSFC`, but double-checking the `Tie` class I see this form enables read and write.

Well, this can be worked around by explicitly calling `SGPropertyNode::setAttribute` on the property:
```c++
    auto node = FGPropertyManager->GetNode(property_name.c_str(), false);
    node->setAttribute(SGPropertyNode::WRITE, false);
```

seanmcleod: Ah, I had just reverted to providing just the `&FGParameter::GetValue` getter, convincing myself that the calls to read `tsfc` wouldn't be at the simulation rate of say 120Hz, but probably only for reporting, e.g. every 5s, or maybe 10Hz or so for graphing etc.

I'll make use of `SGPropertyNode::setAttribute` instead.

bcoconni (approved):

bcoconni:
Excellent ! :+1:
Now we'll need to decide how we manage the fuel flow logic at idle (back in discussion #381 ?)

PR #394, opened 2021-02-10
Properties added for Propeller power, torque, and sense

rega0051:


src/models/propulsion/FGPropeller.cpp
```diff
@@ -163,6 +163,12 @@ FGPropeller::FGPropeller(FGFDMExec* exec, Element* prop_element, int num)
   property_name = base_property_name + "/prop-induced-velocity_fps";
   PropertyManager->Tie( property_name.c_str(), this, &FGPropeller::GetInducedVelocity,
                       &FGPropeller::SetInducedVelocity );
+  property_name = base_property_name + "/propeller-power";
+  PropertyManager->Tie( property_name.c_str(), &PowerRequired );
+  property_name = base_property_name + "/propeller-torque";
```
bcoconni, line 168: Could you please append the unit name to the property names ? We are trying to display the units in which the quantity is given so that users can implement conversion factors when needed. For example the propeller induced velocity on line 163 is `prop-induced-velocity_fps` where `_fps` means `ft/s`.

The unit of Power is `slug*ft^2*s^-3` in `FGPropeller`
https://github.com/JSBSim-Team/jsbsim/blob/77ecd97fa759ab0b8d1fbb33a6ea1748d3aa6732/src/models/propulsion/FGPropeller.cpp#L358
and the unit of Torque is `slug*ft^2*s^-2.`
https://github.com/JSBSim-Team/jsbsim/blob/77ecd97fa759ab0b8d1fbb33a6ea1748d3aa6732/src/models/propulsion/FGPropeller.cpp#L272
I am not much familiar with Imperial/British/US units for power and torques so there might be shorter names for the units above and you are welcome to use them :smiley:

bcoconni (changes requested):
Thanks for the PR. Looks good to me.

Just a minor request: could you please add the unit names to the property names as described below ?

src/models/propulsion/FGPropeller.cpp
rega0051: I agree about the Imperial units, its a dreadful state. Should I only add units to the scope of this PR (the 3 new property tree elements), or throughout more of the code?

The variable naming and units is actually why I didn't want to touch FGElectric with this PR, there is a variable 'HP' and 'PowerWatt'; HP is the current power (in horsepower of course) and PowerWatt, is the maximum power.

bcoconni: > Should I only add units to the scope of this PR (the 3 new property tree elements), or throughout more of the code?

Oops, you just need to update the 3 new properties you are adding. Unfortunately regarding the other properties it's already too late because renaming would break backward compatibility.
> The variable naming and units is actually why I didn't want to touch FGElectric with this PR, there is a variable 'HP' and 'PowerWatt'; HP is the current power (in horsepower of course) and PowerWatt, is the maximum power.

Indeed, we have not always been consistent about internal usage of SI units vs Imperial units. I guess people tend to use the system unit they feel more comfortable with and convert units from/to the Imperial system when needed.

bcoconni: Another option is to add methods for conversion to another unit (for example `FGPropeller::GetPowerRequiredHP()` or `FGPropeller::GetPowerRequiredWatt()`) and use these methods to output the power in the corresponding unit.
You can even create a different property for each unit:
```c++
  property_name = base_property_name + "/propeller-power_watt";
  PropertyManager->Tie( property_name.c_str(), &PowerRequiredWatt );
  property_name = base_property_name + "/propeller-power_hp";
  PropertyManager->Tie( property_name.c_str(), &PowerRequiredHP );
```
Do as you see fit your need as long as the unit name is specified in the property name.

seanmcleod (commented):
For consistency the underscore before the units should be replaced with a dash, e.g.

https://github.com/JSBSim-Team/jsbsim/blob/b7401df2b8a103e45dd758f7c282377fde534a03/src/models/propulsion/FGEngine.cpp#L244-L250

seanmcleod:
Although I see why you went with the underscore, because @bcoconni used it in his comment about including the units in the property's name.

But that doesn't seem to be consistent with how units are generally included as part of the property name.

```c++
PropertyManager->Tie("accelerations/pdot-rad_sec2", this, eP, (PMF)&FGAccelerations::GetPQRdot);
PropertyManager->Tie("forces/fbx-weight-lbs", this, eX, (PMF)&FGAccelerations::GetWeight);
PropertyManager->Tie("accelerations/udot-ft_sec2", this, eU, (PMF)&FGAccelerations::GetUVWdot);
PropertyManager->Tie("moments/l-total-lbsft", this, eL, (PMF)&FGAccelerations::GetMoments);
```

An underscore is typically used within the units to indicate / or 'per', although not consistently in terms of things like 'fps', 'gph' etc.

rega0051:
The only other property with units specified in that file is:
`property_name = base_property_name + "/prop-induced-velocity_fps";`
Changing to a dash for the new properties would make it inconsistent within the file, changing _fps to -fps would break backward compatibility... which is least bad?

seanmcleod:
> The only other property with units specified in that...

Hmm I see 3 other properties, all using the dash convention. So the underscore example `/prop-induced-velocity_fps` is the outlier.

https://github.com/JSBSim-Team/jsbsim/blob/b7401df2b8a103e45dd758f7c282377fde534a03/src/models/propulsion/FGPropeller.cpp#L148

https://github.com/JSBSim-Team/jsbsim/blob/b7401df2b8a103e45dd758f7c282377fde534a03/src/models/propulsion/FGPropeller.cpp#L156

https://github.com/JSBSim-Team/jsbsim/blob/b7401df2b8a103e45dd758f7c282377fde534a03/src/models/propulsion/FGPropeller.cpp#L158

bcoconni:
> Although I see why you went with the underscore, because @bcoconni used it in his comment about including the units in the property's name.

Oops ! My mistake ! @seanmcleod is correct, units name should be appended with a dash not an underscore. Unfortunately there exist some inconsistencies in the code where obviously the rule has not been enforced everywhere :disappointed: 
Sorry for my misleading comment.

bcoconni (approved):

Issue #400, opened 2021-03-19
AeromatiC++ Segmentation fault


kwanzapili:
I downloaded jsbsim-1.1.5 and built on an opensuse linux platform so that I could run aeromatic. The consistent problem I am facing is a segmentation fault at the same step every time I run it. I go through the Q/A session and as soon as I enter "Systems", it fails. The last few of the session are as follows:

```
Htail` arm [ft] (0): 
Vtail area [ft2] (0): 150.76
Vtail arm [ft] (0): 

** Systems **

Segmentation fault (core dumped)

```
A brief look at the backtrace shows that it fails at:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055cfc4444441 in Aeromatic::Propulsion::param_reset() ()

```
Could someone please advise on what is going on?
Thanks.

kwanzapili:
I took a closer look at the code and found that the offending lines are the two commented out lines in the file _utils/aeromatic++/Systems/Propulsion.cpp_

```
void Propulsion::param_reset()
{
    _param = 0;
//    _propulsion[_ptype]->param_reset();
//    _propulsion[_ptype]->_thruster->param_reset();
}
```

With this modification, I can get the program to run without crashing. Is this the right thing to do, please?

bcoconni:
@kwanzapili thanks for reporting the issue.

Aeromatic++ is Erik Hoffman's area of expertise. @ermarch could you please review this issue and @kwanzapili's fix ?

ermarch:
Personally I think _propulsion[_ptype]->_thruster is NULL somehow and causes the segmentation fault.
It would be interesting to know which parameters you parsed to cause this.

ermarch:
Okay, I can only trigger this when I select an aircraft type number which is not listed.
The types range from 0..4 and only when I select 5 this bug is triggered.

kwanzapili:
For me the effect was very consistent regardless of the parameters I entered. For reference, by session was as follows:
```

** AeromatiC++ version 3.3.12
Aeromatiic is a JSBSim configuration file generation utility.
Please enter aircraft data when prompted.

You can always enter 'h' to get verbose help

** General Information **

Output directory [] (/home/kwanza): /home/kwanza/SR71
Create a subdirectory? [] (yes): 
Overwrite? [] (yes): 
Aircraft name [] (my_aircraft): SR71
Use dedicates System files? [] (yes): 
Select a system of measurement [] (no): 
[...]

PR #406, opened 2021-03-26
Fix a delta-wing Lift_alpha and Drag_alpha problem

ermarch:
Fix a delta-wing Lift_alpha and Drag_alpha problem where the values of alpha increase improperly. (issue #400)
Fix a bug introduced by the previous pull request.
Convert more code to C++11


utils/aeromatic++/Systems/Controls.cpp
```diff
@@ -143,6 +141,7 @@ void CableControls::set(const float* cg_loc)
     _aircraft->_CLalpha.at(1) = CLaw[1];
     _aircraft->_CLalpha.at(2) = CLaw[2];
 
+printf("CLalpha: %f\n", _aircraft->_CLalpha.at(0));
```
bcoconni, line 144: Probably a debug message that should be removed ?

bcoconni (commented):
It seems a debug message has been left in the code.
Should it be removed prior to merging the PR ?

utils/aeromatic++/Systems/Controls.cpp
ermarch: Ah, yeas please. It's a debugging message indeed.

PR #434, opened 2021-05-30
AutoPilot for Global5000

gasilva:
The autopilot is meant to assess the fuel consumption in a mission of cruise at 30 kft. It tries to keep the L/D_max in terms of angle of attack and altitude by varying elevator, engine thrust and consequently airspeed.


python/aoa.py
```diff
@@ -0,0 +1,48 @@
+import jsbsim
```
seanmcleod, line 1: The jsbsim/python directory is for files related to building the python wrapper module for JSBSim, not for semi-random useful python scripts. It would potentially be useful to include some of these sorts of scripts as an example of what the python wrapper can be used for, but we should create a separate directory for them.

aircraft/global5000/Systems/GNCUtilities.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0"?>
```
seanmcleod, line 1: Is this version of GNCUtilities.xml any different to the original version in jsbsim/systems? If not can it not be used/included from there?

aircraft/global5000/Systems/Autopilot.xml
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0"?>
```
seanmcleod, line 1: Is this version any different to the version in jsbsim/systems?

gasilva: no they are the same

aircraft/global5000/Systems/GNCUtilities.xml
gasilva: how to do it? just remove them all?

python/aoa.py
gasilva: where can I include them?

gasilva:
pushed again..see new commit

gasilva:
Is it acceptable or not?

PR #435, opened 2021-06-06
AOA vs. airspeed varying altitude, weight and CG

gasilva:
Python script developed by JSBSim Teamd and modified by Guilherme Lima da Silva.


gasilva:
@seanmcleod @bcoconni the code ran here but the test did not.

bcoconni:
Indeed there is something wrong with the scripts `c1723.xml` and  `c1724.xml` which now trigger an FPE (floating point exception). Unfortunately I am unable to reproduce this failure locally.

It is very unlikely that it has any connection with your PR and the test failure is most likely due to some modification to the compilers settings/environment on GitHub. So I would tend to think that this should not hold the merge of your PR, but unfortunately I did not have the time to review it thus far. I'll try later this week unless @seanmcleod beat me to it 😉 

On a side note, these test failures might be triggered by genuine bugs so I'll have to investigate using distant debugging, a feature which I am very unfamiliar with so there is a lot of fun in perspective 😞

gasilva:
Information got in the smartcockpit site [``](https://www.smartcockpit.com)

54570482-e10b-49b4-9809-68abb18bf35a
c6cddacb-8262-4ebd-aad7-c7da6c43e46a

The wing is a little to forward compared to aeromatic

examples/python/aoa.py
```diff
@@ -0,0 +1,98 @@
+# Originally developed by JSBSim Team
+# Modified by Guilherme A. L. da Silva - aerothermalsolutions.co
+# Calculation required by aircraft icing enginering
+
+import jsbsim
+from IPython.core.display import display, HTML
```
bcoconni, line 6:
```suggestion
```
These symbols `display` and `HTML` are not used in the script.

examples/python/aoa.py
```diff
@@ -0,0 +1,98 @@
...
+import xml.etree.ElementTree as ET
+import matplotlib.pyplot as plt
+import matplotlib.style as style
+style.use('fivethirtyeight')
```
bcoconni, line 10:
```suggestion
```
Such customizations should not be used in an example to keep things as simple as possible.

examples/python/aoa.py
```diff
@@ -0,0 +1,98 @@
...
+        speed, alpha = zip(*results)
+        plt.plot(speed, alpha,label="{0} weight {1:.0f} kft {2:.2f} % cg".format(weight[i],h_ft[j]/1000,(float(cgPos[j])/float(cgOrig)-1)*100))
+
+#plot final results
+plt.legend(frameon=False)
+plt.xlabel('KCAS (kt)')
+plt.ylabel('AoA (deg)')
+plt.title('AoA vs KCAS')
+
+plt.show()
```
bcoconni, line 95:
```suggestion
```
Not sure if these plots should be included in an example. You already print the results in the console a few lines above ? It is also unclear if someone learning to use JSBSim would have matplotlib installed (or would know how to install it ?)

I am open to suggestions.

examples/python/aoa.py
```diff
@@ -0,0 +1,98 @@
+# Originally developed by JSBSim Team
+# Modified by Guilherme A. L. da Silva - aerothermalsolutions.co
+# Calculation required by aircraft icing enginering
+
```
bcoconni, line 4: May be you should add to the comments that this script is designed to be executed from the folder which contains the jsbsim module or that the module should be installed via 'pip' or 'conda' ?

examples/python/aoa.py
```diff
@@ -0,0 +1,98 @@
...
+from IPython.core.display import display, HTML
+import xml.etree.ElementTree as ET
+import matplotlib.pyplot as plt
+import matplotlib.style as style
+style.use('fivethirtyeight')
+
+# function to change CG in aircraft xml
+# change the directory to the aircraft to be studied
+def changeCG(cgPos,readOnly):
+    tree = ET.parse('../../aircraft/global5000/global5000.xml')
```
bcoconni, line 15: I don't feel comfortable with hardcoded relative paths because they assume the folder structure of the user is exactly the same than yours. At the very least, you should document what the directory structure is expected to be for this script to be run.

examples/python/aoa.py
```diff
@@ -0,0 +1,98 @@
...
+    root = tree.getroot()
+
+    for x in root.findall('mass_balance/location'):
+        cg = x.find('x').text
+        if not readOnly:
+             x.find('x').text=str(cgPos)
+             tree.write('../../aircraft/global5000/global5000.xml')
```
bcoconni, line 22: Some more hardcoded relative paths: same comment than above. May be you could use a variable to store the file name and save the duplication ?

examples/python/aoa.py
```diff
@@ -0,0 +1,98 @@
...
+#vary CG in the study
+cgPos=[cgOrig*1.02,cgOrig*0.98]
+
+#vary altitude
+h_ft=[8000,30000]
+
+#run the simulation varying CG, altitude, speed and total weight
+for j in range(2):
+    cg=changeCG(cgPos[j],False)
+    fdm = jsbsim.FGFDMExec('../../')
```
bcoconni, line 50: Yet another hardcoded relative path: same comment than above.

bcoconni (changes requested):
Thanks for your PR @gasilva :+1:

Your PR is made of 10 new/updated files but they seem to address different purpose, don't they ? Usually a PR addresses only one topic and I am unsure that this is indeed the case here ?

I have only reviewed the Python script because this was what was initially intended to be included in a PR according to [our discussion ](https://github.com/JSBSim-Team/jsbsim/discussions/428#discussioncomment-826244) but I may be wrong in which case I will happily review all the files in this PR.

gasilva:
I agree with all comments. Thank for your time and patience.

seanmcleod (commented):
> Your PR is made of 10 new/updated files but they seem to address different purpose, don't they ? Usually a PR addresses only one topic and I am unsure that this is indeed the case here ?

Yep, this was also confusing to me. I think the issue is that the PR was made from the master branch of your fork, and then while you waited for feedback from us you continued your development and testing by pushing commits to your master branch which then pollute/confuse this pull request.

What I typically do if I want to submit a pull request is create a separate branch in my fork, and issue the pull request from that branch. I can then continue developing and testing in my master branch or other branches while waiting for feedback, and when needed just commit any required changes to the branch tied to the pull request. Keeps things simpler and separated.

gasilva:
Sorry for the confusions

gasilva:
Yes ... the CG is the one from Aeromatic...It is confirmed by the paper [https://www.researchgate.net/publication/303902873_Conceptual_Flight_Dynamic_Model_for_Bombardier_Global_5000_Aircraft](https://www.researchgate.net/publication/303902873_Conceptual_Flight_Dynamic_Model_for_Bombardier_Global_5000_Aircraft)

examples/python/aoa.py
```diff
@@ -0,0 +1,96 @@
...
+import matplotlib.pyplot as plt
+
+# function to change CG in aircraft xml
+# change the directory to the aircraft to be studied
+def changeCG(cgPos,readOnly):
+    tree = ET.parse('../../aircraft/global5000/global5000.xml')
+    root = tree.getroot()
+
+    for x in root.findall('mass_balance/location'):
+        cg = x.find('x').text
```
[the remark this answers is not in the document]
bcoconni: Good catch !

gasilva:
Yes, now is OK

examples/python/aoa.py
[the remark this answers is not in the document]
bcoconni: This could easily be fixed with the [parallel axis theorem](https://en.wikipedia.org/wiki/Parallel_axis_theorem) however.

bcoconni:
@gasilva As we previously mentioned, may be you should try removing extra files from the PR and only keep the Python script ?
I also forgot to mention that the script should include an open source license to be pushed to JSBSim.

gasilva:
I tried to delete the files form commit...I do not know if it worked...please let me know how to do it.

gasilva:
I just can not do it...

bcoconni:
> I just can not do it...

Have you tried using `git rm` ?

The usual procedure to remove a file `path/file.xml` is first tell `git` to the files you plan to remove from the next commit:
```bash
> git rm path1/file1.xml path2/file2.xml path3/file3.cpp
```
then commit the change
```bash
> git commit -m "Removed obsolete files"
```
then push the commit
```
> git push my_repo my_branch
```

seanmcleod:
> I just can not do it...

Try this, which will also be good practice for future pull requests.

Create a branch for this pull request, e.g. `python-plot-example`. Now normally this new branch would be branched from the head of your master branch which matches some point in time the main JSBSim repo. However since you've pushed other commits to your master branch that aren't relevant/needed/wanted in this case I suggest you create this branch from an earlier point in time when your master branch was still in sync with the main JSBSim repo. So for example:

```sh
git checkout -b python-plot-example a35be5c
```

Where a35be5c is the commit at the time your autopilot changes were merged into the main JSBSim repo via a pull request.

Now on this new branch `python-plot-example` create the python sample directory and add in the python sample file, btw I suggest a better name than `aoa.py`, something more descriptive, e.g. `trimmed-aoa-vs-speed-plot.py` or something.

Then generate a pull request from this `python-plot-example` branch. The pull request will then only show this single new python file. Which we can comment on, make suggestions etc. Based on possible comments/suggestions make the changes via subsequent commits in this branch. Don't commit any other files you might be working on into this branch, e.g. autopilot updates, keep those in a separate branch.

bcoconni:
FWIW the commit history is not relevant for PRs in JSBSim since we "squash & merge" which, as the name implies, squashes the PR commit history into one single change which is then merged with `master`.

gasilva:
I opened a new PR with clean files from a new branch as requested...This one is very polluted.

Issue #436, opened 2021-06-06
Aeromatic alpha value output


kwanzapili:
**I'm submitting a ...**
  - [X] bug report
  - [X] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**
1. For the bug part, Aeromatic does not generate a logical "_aero/force/Lift_alpha_" function. The "_aero/alpha-rad_" values are in indeterminate order
2. Aeromatic does **not** offer generate a propulsion system or a drag parachute. The output then states that the aircraft has no engine

**What is the current behavior?**
I used the parameters used in the attached file "_vulcan.log_" and the output is shown in "_vulcan.txt_". See the first function in the "LIFT" axis.

The program outputs the following error when run;
```
*** ERROR: The alpha value for maximum lift is too high.
           This means the specified Stall Speed was too low.
           Make sure it is for a clean (no gear and no flaps) configurtion.

```
I am not sure what stall speed it expects. I have tried entering nothing and other reasonable values without any better luck.

**What is the expected behavior?**
In Aeromatic v1.1.5, we had the offer to generate propulsion and other system tiles. This seems to have be omitted in v1.1.6.
The alpha lift behaviour seems to be similar to issue #400 that I had reported before.

**What is the motivation / use case for changing the behavior?**
Simply revert to prior features.

**Please tell us about your environment:**
-  Linux
-  JSBSim 1.1.6

**Other information**
Refer to issue #400 for some prior problems with the output.
[vulcan.log](https://github.com/JSBSim-Team/jsbsim/files/6604709/vulcan.log)
[vulcan.txt](https://github.com/JSBSim-Team/jsbsim/files/6604710/vulcan.txt)

bcoconni:
@kwanzapili Thanks for the bug report.

Just to clarify things: when you are referring to `Aeromatic`, are you talking about the [Aeromatic web site](http://jsbsim.sourceforge.net/aeromatic2.html) or are you talking about the command line utility `aeromatic++` ?

kwanzapili:
Sorry for the lapse in responding. I mean the Aeromatic CLI that you get from compiling the JSBSim tar ball (`AeromatiC++ version 3.3.14`). I have tried two web-based versions: one never generates any output (it rigidly stays at the input form page) and the other works - (http://jsbsim.sourceforge.net/aeromatic2.html). The only reason I was trying the CLI version was because I wanted finer controls of the inputs.

[...]

PR #450, opened 2021-06-22
Fix delta wings

ermarch:
This fixes issue #436


bcoconni:
Thanks for the PR ! It has been pulled in JSBSim even though one of the tests fails since this PR is deemed unrelated to that failure.

PR #457, opened 2021-06-27
only aoa.py

gasilva:
pure file...deleted fork..and commited again

### DIFF
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -8,8 +8,8 @@ import matplotlib.pyplot as plt
 
 # function to change CG in aircraft xml
 # change the directory to the aircraft to be studied
-def changeCG(cgPos,readOnly):
-    tree = ET.parse('../../aircraft/global5000/global5000.xml')
+def changeCG(fdm, cgPos, readOnly):
+    tree = ET.parse(os.path.join(fdm.get_root_dir(), 'aircraft/global5000/global5000.xml'))
     root = tree.getroot()
 
     for x in root.findall('mass_balance/location'):
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -5,8 +5,6 @@
 import jsbsim
 import xml.etree.ElementTree as ET
 import matplotlib.pyplot as plt
-import matplotlib.style as style
-style.use('fivethirtyeight')
 
 # function to change CG in aircraft xml
 # change the directory to the aircraft to be studied
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -85,15 +85,6 @@ for j in range(2):
             print(result[0], result[1])
 
         speed, alpha = zip(*results)
-        plt.plot(speed, alpha,label="{0} weight {1:.0f} kft {2:.2f} % cg".format(weight[i],h_ft[j]/1000,(float(cgPos[j])/float(cgOrig)-1)*100))
-
-#plot final results
-plt.legend(frameon=False)
-plt.xlabel('KCAS (kt)')
-plt.ylabel('AoA (deg)')
-plt.title('AoA vs KCAS')
-
-plt.show()
 
 #restore original CG for the aircraft xml
 cgx=changeCG(str(cgOrig),False)
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -64,7 +64,6 @@ for j in range(2):
             fdm['propulsion/tank[1]/contents-lbs'] = fuel[i]
             fdm['propulsion/tank[2]/contents-lbs'] = fuel[i]
             fdm['inertia/pointmass-weight-lbs[0]'] = payload[i]
-            fdm['inertia/mass_balance-positionx-in'] = cgPos[j]
 
             fdm.run_ic() # Initialize the aircraft with initial conditions
             fdm.run()
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -16,7 +16,7 @@ def changeCG(fdm, cgPos, readOnly):
         cg = x.find('x').text
         if not readOnly:
              x.find('x').text=str(cgPos)
-             tree.write('../../aircraft/global5000/global5000.xml')
+             tree.write(os.path.join(fdm.get_root_dir(), 'aircraft/global5000/global5000.xml'))
     return cg
 
 #Fuel Max for Global5000
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -1,6 +1,11 @@
 # Originally developed by JSBSim Team
 # Modified by Guilherme A. L. da Silva - aerothermalsolutions.co
 # Calculation required by aircraft icing enginering
+# GNU Lesser General Public License v2.1
+
+# to install jsbsim module in Python
+# pip install jsbsim
+# or conda install jsbsim (if you have anaconda, this is the best way)
 
 import jsbsim
 import xml.etree.ElementTree as ET
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -43,8 +43,8 @@ h_ft=[8000,30000]
 
 #run the simulation varying CG, altitude, speed and total weight
 for j in range(2):
-    cg=changeCG(cgPos[j],False)
-    fdm = jsbsim.FGFDMExec('../../') 
+    fdm = jsbsim.FGFDMExec(PATH_TO_JSBSIM_FILES) 
+    cg=changeCG(fdm, cgPos[j],False)
     fdm.load_model('global5000')
     # Set engines running
     fdm['propulsion/engine[0]/set-running'] = 1
--- a/examples/python/aoa.py
+++ b/examples/python/aoa.py
@@ -9,7 +9,6 @@
 
 import jsbsim
 import xml.etree.ElementTree as ET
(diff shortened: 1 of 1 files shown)


seanmcleod:
Yep, only single file. But you've committed it to your master branch so remember you can't push any further commits to your master branch that aren't specifically related to this pull request until this pull request is merged.

gasilva:
> Yep, only single file. But you've committed it to your master branch so remember you can't push any further commits to your master branch that aren't specifically related to this pull request until this pull request is merged.

OK I understood

gasilva:
I would like to submit some python scripts:

- long flight for fuel consumption calculation
- design maneuver speeds and AoA plots
- TAs variation with AoA
- fuel consumption with Mach variation as discussed with @seanmcleod by considering fuel absolute consumption and fuel/miles consumption

However, I need to close this PR first. Or should I make a new branch?

gasilva:
True Airspeed (TAS) variation
Figure_3 copy 2

Design Maneuver Speeds
Figure_6

**Long flight fuel consumption**
Figure_1  conflicted  copy

**Fuel used**
Figure_2 copy 2

**Fuel used/miles**
Figure_1 copy 4

seanmcleod:
> However, I need to close this PR first. Or should I make a new branch?

Yes I'd recommend always creating a new branch, otherwise if there is a delay in merging an earlier pull request like this one then you're not held up.

JSBSim wise I've been busy looking into this issue https://github.com/JSBSim-Team/jsbsim/discussions/455 so I haven't had time to comment on this pull request yet.

gasilva:
No problem I will wait.

Enviado do meu iPhone

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
...
+import jsbsim
+import xml.etree.ElementTree as ET
+import matplotlib.pyplot as plt
+import matplotlib.style as style
+style.use('fivethirtyeight')
+
+# function to change CG in aircraft xml
+# change the directory to the aircraft to be studied
+def changeCG(cgPos,readOnly):
+    tree = ET.parse('../../aircraft/global5000/global5000.xml')
```
bcoconni, line 14:
```suggestion
def changeCG(fdm, cgPos, readOnly):
    tree = ET.parse(os.path.join(fdm.get_root_dir(), 'aircraft/global5000/global5000.xml'))
```
Don't use hardcoded paths in an example because the code will fail whenever the example is copied in another folder. You could use the FDM root directory(`fdm.get_root_dir()`) instead.

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
...
+    root = tree.getroot()
+
+    for x in root.findall('mass_balance/location'):
+        cg = x.find('x').text
+        if not readOnly:
+             x.find('x').text=str(cgPos)
+             tree.write('../../aircraft/global5000/global5000.xml')
```
bcoconni, line 21:
```suggestion
             tree.write(os.path.join(fdm.get_root_dir(), 'aircraft/global5000/global5000.xml'))
```
Another hardcoded path. Same comment as above try not to duplicate locations where hardcoded paths are used. Using `fdm.get_root_dir()` is one way to work around this problem.

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
...
+#vary CG in the study
+cgPos=[cgOrig*0.95,cgOrig*1.05]
+
+#vary altitude
+h_ft=[8000,30000]
+
+#run the simulation varying CG, altitude, speed and total weight
+for j in range(2):
+    cg=changeCG(cgPos[j],False)
+    fdm = jsbsim.FGFDMExec('../../')
```
bcoconni, line 49:
```suggestion
    fdm = jsbsim.FGFDMExec(PATH_TO_JSBSIM_FILES) 
    cg=changeCG(fdm, cgPos[j],False)
```
I suggest to switch the order of these 2 lines so that `changeCG` can make reference to the FDM root directory rather than using an hardcoded relative path.

Also I'd suggest to use a global variable `PATH_TO_JSBSIM_FILES` that is defined at the very top of the script and documented as being needed to be updated if the script is copied in another directory.

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
...
+        results = []
+        for speed in range(90, 460, 10):
+            fdm['ic/h-sl-ft'] = h_ft[j]
+            fdm['ic/vc-kts'] = speed
+            fdm['ic/gamma-deg'] = 0
+            fdm['propulsion/tank[0]/contents-lbs'] = fuel[i]
+            fdm['propulsion/tank[1]/contents-lbs'] = fuel[i]
+            fdm['propulsion/tank[2]/contents-lbs'] = fuel[i]
+            fdm['inertia/pointmass-weight-lbs[0]'] = payload[i]
+            fdm['inertia/mass_balance-positionx-in'] = cgPos[j]
```
bcoconni, line 64: As previously mentioned by @seanmcleod the property `inertia/mass_balance-positionx-in` doesn't exist.

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
...
+def changeCG(cgPos,readOnly):
+    tree = ET.parse('../../aircraft/global5000/global5000.xml')
+    root = tree.getroot()
+
+    for x in root.findall('mass_balance/location'):
+        cg = x.find('x').text
+        if not readOnly:
+             x.find('x').text=str(cgPos)
+             tree.write('../../aircraft/global5000/global5000.xml')
+    return cg
```
bcoconni, line 22: As previously mentioned by @seanmcleod in your PR #435:
> This cg is the cg for the empty weight aircraft with the associated moments of inertia for this empty weight aircraft. So you can't move it without also updating the moments of inertia.

This could be fixed with the [parallel axis theorem](https://en.wikipedia.org/wiki/Parallel_axis_theorem) however.

bcoconni (changes requested):
Sorry for taking so much time to review your code but I haven't had much opportunity recently. The comments are mostly a duplication of the PR #435 that you have discarded.

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
...
+        speed, alpha = zip(*results)
+        plt.plot(speed, alpha,label="{0} weight {1:.0f} kft {2:.2f} % cg".format(weight[i],h_ft[j]/1000,(float(cgPos[j])/float(cgOrig)-1)*100))
+
+#plot final results
+plt.legend(frameon=False)
+plt.xlabel('KCAS (kt)')
+plt.ylabel('AoA (deg)')
+plt.title('AoA vs KCAS')
+
+plt.show()
```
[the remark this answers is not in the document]
gasilva: matplotlib is pretty standard and comes with anaconda for example...simply install it with pip install matplotlib or conda install matplotlib...I think the visual results are important

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
+# Originally developed by JSBSim Team
+# Modified by Guilherme A. L. da Silva - aerothermalsolutions.co
+# Calculation required by aircraft icing enginering
```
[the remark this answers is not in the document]
gasilva: we can choose GNU Lesser General Public License v2.1 for example...same license than jsbsim

examples/python/aoa.py
gasilva: Yes...but it refers to fuel weight only not empty weight.

examples/python/aoa.py
gasilva: you can take it out since I am change CG directly in the XML file

examples/python/aoa.py
```diff
@@ -0,0 +1,97 @@
...
+
+import jsbsim
```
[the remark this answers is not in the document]
gasilva:
```python
# to install jsbsim module make 
# pip install jsbsim or conda install jsbsim
```

gasilva (commented):
changes made...despite I think plot was important.

bcoconni (approved):

bcoconni:
Thanks for the updates @gasilva :+1: 

> changes made...despite I think plot was important.

Well, by no means did I intend to force you to remove the plots. It was just that it was (and still is) unclear to me if plots are relevant for an example. Hence my statement _"I am open to suggestions"_. No hard feelings I hope :smile_cat: 

Anyway, I have merged the PR because I think we all agree that we should move forward on this topic. Further improvements will be introduced by new PRs.

Thanks again for your commitment @gasilva on this PR which has unexpectedly been quite a bumpy road.

gasilva:
I think the python example is important. For me it was very helpful. I like
gnuplot, but plotting in python is better for example. Another thing
that allows some math and variables/results manipulation.

I have another example running a script instead of running JSBSim itself. I
can PR a series of python examples if you are interested. They are
important for icing engineering, which is a sub-activity of aircraft
performance. In icing we need an aircraft operational envelope in order to
find the critical icing conditions to system design and ice shapes
prediction.

Guilherme A. Lima da Silva, Dr.

Em dom., 11 de jul. de 2021 às 09:28, Bertrand Coconnier <
***@***.***> escreveu:

bcoconni:
You are most welcome to submit new Python scripts but they have to be related to JSBSim, showing for example how to use one of its feature or maybe bringing a feature that is not available to the default vanilla JSBSim.

PR #466, opened 2021-07-03
Provide mode flag to ResetToInitialConditions() to skip the execution of RunIC()

seanmcleod:
See https://github.com/JSBSim-Team/jsbsim/discussions/455#discussioncomment-959614

### DIFF
--- a/src/FGFDMExec.cpp
+++ b/src/FGFDMExec.cpp
@@ -692,8 +692,6 @@ void FGFDMExec::ResetToInitialConditions(int mode)
   if (Constructing) return;
 
   // mode flags
-  const int START_NEW_OUTPUT    = 0x1;
-  const int DONT_EXECUTE_RUN_IC = 0x2;
 
   if (mode & START_NEW_OUTPUT) Output->SetStartNewOutput();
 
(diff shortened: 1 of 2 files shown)


src/FGFDMExec.cpp
```diff
@@ -691,7 +691,11 @@ void FGFDMExec::ResetToInitialConditions(int mode)
 {
   if (Constructing) return;
 
-  if (mode == 1) Output->SetStartNewOutput();
+  // mode flags
+  const int START_NEW_OUTPUT    = 0x1;
+  const int DONT_EXECUTE_RUN_IC = 0x2;
```
bcoconni, line 696:
```suggestion
```
May be we should take the opportunity of your PR to locate **and document** these constants in the header `FGFDMExec.h` ? Currently, only the most motivated users could unearth the existence of these constants and their role.

bcoconni (changes requested):
Thanks for the PR. I suggest to move the constants to the header `FGFDMExec.h` where they could be documented and therefore easier to get for someone looking for guidance about how to use `FGFDMExec::ResetToInitialConditions`.

seanmcleod:
@bcoconni I was in a bit of a rush originally. I've moved the constants to the header file and added some documentation in the latest commit.

bcoconni (approved):
Thanks for updating the code.

PR #470, opened 2021-07-10
Remove old unused reset code

seanmcleod:
As mentioned by @rjmccabe3701 the `ResetMode` variable and it's references are dead code. Looks like it used to be tied directly to `simulation/reset`.

```c++
instance->Tie("simulation/reset", (int*)&ResetMode);
```


bcoconni:
Good catch ! The change has been introduced by the commit 4599c458ce175dd951fbf4941e9f666d907709b3 then partially undone by the commit 6f0d7e5d4fa2b6c6b3c1f84407fa5cfd0f928445.

PR #472, opened 2021-07-11
Display the stack trace when an FPE is raised - Linux

bcoconni:
We have [an FPE (Floating Point Exception) which is raised when building JSBSim for MSVC](https://github.com/JSBSim-Team/jsbsim/runs/3040077190?check_suite_focus=true#step:12:964) for quite some time now. This exception is responsible for the build failure of our CI/CD (GitHub Actions). The failure is reproducible since it causes the build to fail repeatedly for a couple of weeks now. Unfortunately I am unable to reproduce the FPE on my computer, neither on Linux nor on Windows. The only option remaining is to display the stack trace directly on GitHub logs whenever an FPE is raised.

This PR modifies the module `fpectl` to do just that: it uses the library [Backward-cpp](https://github.com/bombela/backward-cpp) to output the stack trace in the GitHub tests log when an FPE is raised. This makes the code and build process of `fpectl` a bit more complex than you would expect for such a module. However I do not expect that to be an issue because:
* By default the flag `FPECTL_DISPLAY_STACK_TRACE` is set to `OFF` meaning that CMake will not try to include the stack trace feature by default.
* This is really a debug feature for the developers of JSBSim and is therefore not expected to be used by the average user (who most likely will not even realize the existence of the feature due to the default setting mentioned above).

This PR only introduces the display of the stack trace for the Linux OS. I am more familiar with this platform so I started coding for it before moving to Windows. The stack trace is successfully displayed by the test `test_fpectl`.

I am currently porting the feature to Windows with some success since [my experiments are currently pointing out the failure to be located in `src\GeographicLib\Geodesic.cpp` line 400](https://github.com/bcoconni/jsbsim/runs/3039985005?check_suite_focus=true#step:13:1008). I will submit another PR to include the feature for Windows on JSBSim.

Of course, the final step will be fixing the bug !


bcoconni:
The stack trace is [displayed successfully for `test_fpectl`](https://github.com/JSBSim-Team/jsbsim/runs/3040260823?check_suite_focus=true#step:13:60). I am merging the PR.

PR #476, opened 2021-07-11
Sample python script documentation tweaks

seanmcleod:
@gasilva I've made some documentation changes in the python script file.

I've also modified the global5000.xml file because I noticed that if I ran the sample then even with the cg position being reset git would show that there were changes to the file, in particular some of your comments were being lost. So I've tweaked it so that no changes show up. This is really just a temporary work-around, since a user editing the python sample to run it against another aircraft will run into this issue as well. The better approach, for a future update is to make these cg changes on a temporary version of the aircraft file.

### DIFF
--- a/aircraft/global5000/global5000.xml
+++ b/aircraft/global5000/global5000.xml
@@ -1,39 +1,31 @@
-<?xml version="1.0"?>
-<?xml-stylesheet type="text/xsl" href="http://jsbsim.sourceforge.net/JSBSim.xsl"?>
-<fdm_config name="global5000" version="2.0" release="ALPHA"
-   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-   xsi:noNamespaceSchemaLocation="http://jsbsim.sourceforge.net/JSBSim.xsd">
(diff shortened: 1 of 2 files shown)


gasilva:
Is it possible to make RW for the CG x-location?

seanmcleod:
One option in terms of moving the cg and keeping all the moments of inertia correct is to use a point mass and update the point mass's location, which is RW and/or it's mass.

https://github.com/JSBSim-Team/jsbsim/blob/11ea52ed340ea32a34e6e3dfc6df503f9ad16cf5/src/models/FGMassBalance.cpp#L424-L441

When the point mass's mass changes and/or it's location changes JSBSim will recalculate the moments of inertia.

gasilva:
How to do a variation of CG in terms of MAC? Or in percentage?

seanmcleod:
You would need to know where the start of the MAC is in the structural frame. It's not required and so it's typically not specified in an aircraft's .xml file. Then you would need to query what the current mass and cg is, and then use that to work out what point mass mass and location would move the cg to the position you want relative to the MAC.

But let's open a separate discussion on this as a new discussion item. In terms of this pull request let's keep the discussion to whether the proposed changes are useful or not, or whether there are any bugs in the proposed changes etc.

PR #478, opened 2021-07-15
Python simple scripts

gasilva:
To implement three python scripts to give an example of powerful integration between jsbsim and python. For engineering purposes may be the most recommended tool.

All have graphics outputs but they can be changed to a table format.
The plot are more useful basically.

- `catalog.py`: 
List properties of a script
This is good to know which variable to read and write

- `runScriptLongFlight.py`
Calculation required by aircraft icing enginering
Information required to calculate fuel consuption in leveled flight with no slats or flaps in order to provide data for mission profile analysis

Other files as changed like `global5000.xml, airborne.xml, trim-cruise_ap.xml`

- `runJSBSimFuelMach.py`
Calculation required by aircraft icing enginering
Find the optimum value for fuel consumption 
For greatest distance - low AoA and high speed (Carson cruise)
For greatest time - high AoA and low speed (miminum power required)


gasilva:
I have two other scripts but I need to change CG by putting a point mass instead of changing the aircraft XML file as already said by @seanmcleod .

Issue #125, opened 2018-10-29
Inertial tensor matrix initialization error


jinzhu1911:
The inertia matrix J for rigid body is  
FGMatrix33(  bixx,  -bixy,  -bixz,
                    -bixy,  biyy,  -biyz,
                    -bixz,  -biyz,  bizz )
  (ref. Stevens and Lewis, "Flight Control & Simulation"),
but in the function FGMassBalance::Load of FGMassBalance class , 
you pass FGMatrix33(  bixx,  -bixy,  bixz,
                                 -bixy,  biyy,  -biyz,
                                   bixz,  -biyz,  bizz ) to SetAircraftBaseInertias.
I think this is wrong.
Wait for your answer. Thanks very much!

bcoconni:
I first thought you were right but after some investigations, the code is actually correct :smiley: 

The trick is that the matrix in Steven & Lewis is given in the **body** frame while the inertia in JSBSim are specified in the **structural** frame (see [JSBSim reference manual](https://jsbsim-team.github.io/jsbsim-reference-manual/mypages/user-manual-frames-of-reference) about frame definitions).

The opposite sign of `bixz` with respect to Steven & Lewis is therefore a result of the transformation from the structural frame to the body frame (see the message of commit f04489499d477).

I will add some comments to the code since, I agree, the motivation behind the sign modification is not obvious.

bcoconni:
After some further investigations, the problem was raised in 2006 (see the [SF ticket 46](https://sourceforge.net/p/jsbsim/bugs/46/)) and, as I mentioned above, solved by the commit f044894 in 2007.

At the time @jonsberndt wrote
> The aircraft inertias are read in correctly. However, they are given in structural coordinates, with the X axis positive aft, and the Y axis positive out the right wing. When they are read in, they are transferred to body frame, which involves a 180 degree roation about the Y axis. That changes the sign of the x and z terms, so any product of inertia that involves the x and y axis, or z and y axis should flip it sign. Unfortunately, the xz term was also having its sign flipped, which it should not have. So, that has been fixed. The relevant setting function is now:
```c++
SetAircraftBaseInertias(FGMatrix33( bixx, -bixy,  bixz,
                                   -bixy,  biyy, -biyz,
                                    bixz, -biyz,  bizz ));
```
> Before, there was a minus sign in front of the "bixz" variable. Now, there is not.

[...]

PR #502, opened 2021-10-04
Fixed inertia matrix error

likangbei:
Under the assumption that in the structural frame the:
 - X-axis is directed afterwards,
 - Y-axis is directed towards the right,
 - Z-axis is directed upwards,
And then a 180 degree rotation is done about the Y axis so that the:
 - X-axis is directed forward,
 - Y-axis is directed towards the right,
 - Z-axis is directed downward.
The original ixx,iyy,izz,ixy,ixz,iyz in structural frame.
So, When transform the inertia products from the structural frame to the body frame, ixy and iyz should be reversed.

### DIFF
--- a/src/models/FGMassBalance.cpp
+++ b/src/models/FGMassBalance.cpp
@@ -117,9 +117,9 @@ static FGMatrix33 ReadInertiaMatrix(Element* document)
 
   // Transform the inertia products from the structural frame to the body frame
   // and create the inertia matrix.
-  return FGMatrix33( bixx, -bixy,  bixz,
-                    -bixy,  biyy, -biyz,
-                     bixz, -biyz,  bizz );
+  return FGMatrix33( bixx,  bixy, -bixz,
+                     bixy,  biyy,  biyz,
+                    -bixz,  biyz,  bizz );
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%


seanmcleod:
@likangbei have you taken a look at this issue - https://github.com/JSBSim-Team/jsbsim/issues/125

likangbei:
> @likangbei have you taken a look at this issue - #125
Ixz defined by DHC6 XML is wrong, it should be positive. It is recommended to consult the book《Airplane Flight Dynamics and Automatic Flight Controls part1》，It contains a large number of other aircraft ixz data。Of course, many flight dynamics books contain a large number of aircraft ixz data, many of which are actual values. If the ixz error of dhc6 is regarded as correct, it is not difficult to understand the error of inertia matrix. Is the ixz data source of dhc6 reliable? Or is it calculated using a non-standard structural coordinate system? As an aviation engineer, I think it is very likely to be wrong.

bcoconni:
Following the discussion we had in #125, I'd also suggest that we take opportunity to allow supplying the inertia tensor in either the structural or the body frame.

@likangbei you could modify this PR to allow JSBSim:
1. Reading the inertia components `Ixy`, `Ixz` and `Iyz` with the same sign convention than classical flight dynamics textbooks.
2. Reading the inertia tensor in either the body or the structural frame.

This could be achieved by adding options to `<mass_balance>`:
* `<mass_balance frame="BODY|STRUCTURAL">` to supply the inertia tensor in the body or structural frame.
* `<mass_balance sign_convention="textbooks">` to read the non diagonal components of the inertia tensor
* Any combination thereof.

Or any other option that you might think of.

seanmcleod:
@bcoconni while I was refreshing my memory in terms of products of inertia I came across the Wikipedia page - https://en.wikipedia.org/wiki/Moment_of_inertia#Inertia_tensor

Which in particular mentions both conventions.

> There are some CAD and CAE applications such as SolidWorks, Unigraphics NX/Siemens NX and MSC Adams that use an alternate convention for the products of inertia. According to this convention, the minus sign is removed from the product of inertia formulas and instead inserted in the inertia matrix: 

But in the end the sign is either in the definition of the product of inertia or in the inertia matrix.

aarondewindt:
@bcoconni Yea, using the name `textbook` is probably not a good idea, I have the feeling I've come both across in textbooks.

I've seen people refer to JSBSim's current convention as `negated`, since the product of inertia is already negated. So I would suggest `<mass_balance negated_inertia="true|false">`? I'll have a look at my old notes tomorrow to see if I find some other more relevant names. It's not a reference frame thing, it really is only the sign convention.

In the meantime I think this page from Mathworks explains the issue quite nicely. JSBSim and Simscape use the same convention, so adding a similar (or same) explanation to JSBSim's docs with a link in the FAQ would help people with the same issue. 
https://www.mathworks.com/help/physmod/sm/ug/specify-custom-inertia.html

likangbei:
@bcoconni 
At present, in the field of flight dynamics, the standard ixz definition is as follows
image
not
image

Many people with basic knowledge of flight dynamics cannot realize this problem when using jsbsim. They often use jsbsim in professional research fields, which often require accurate simulation. Of course, if you mistake the sign of ixz, you naturally don't care much about such simulation errors for flight games such as FlightGear

Adding additional tags is not a good idea. The best way is to be compatible with the standard definitions in the field of flight dynamics. After all, jsbsim is a professional flight simulation software

In some current aircraft XML files, for example, ixz of B737 is defined according to the flight dynamics standard. Obviously, the author of XML does not realize that ixz in jsbsim is different from the standard definition

Adding options to <mass_balance> is Not a good idea . If it is not changed to the default standard definition, at least remind all users clearly

aarondewindt:
@likangbei Yea, I get what you're saying and the same happened to me a while back. However JSBSim has been under development for a long time, so changing convention will most likely cause more issues than solve anything.

Adding the documentation will at least give a explanation of the issue and the new option should draw the attention of those filling the xml to it.

aarondewindt:
@bcoconni another way to solve the issue is to allow users to fill in the inertia matrix directly simmilar to how tabels are filled in, there no ambiguity on the sign convention there.

This could be the new recommended way and the current method can be depreciated and left there for backwards compatibility.

bcoconni:
> However, I still recommend explicit definitions so that the default can be changed in the future. Reduce the possibility of unconscious errors. In addition, it is suggested to distinguish the versions of XML models to reduce the burden of maintaining compatibility in the process of software upgrade.

I agree. I guess the JSBSim project should lead as an example meaning we should update all the models in the `aircraft\` folder with `<mass_balance negated_inertia="true">` ?

likangbei:
> I agree. I guess the JSBSim project should lead as an example meaning we should update all the models in the `aircraft\` folder with `<mass_balance negated_inertia="true">` ?

Yes, I've thought about it too. I think it's necessary to modify all aircraft configuration files. Many people will learn or modify based on the existing aircraft configuration file. Adding "negative_crossproduct_inertia" will help remind users to pay attention to the sign of ixz, which may be more effective than FAQ to reduce errors.

bcoconni:
> Yes, I've thought about it too. I think it's necessary to modify all aircraft configuration files. Many people will learn or modify based on the existing aircraft configuration file. Adding "negative_crossproduct_inertia" will help remind users to pay attention to the sign of ixz, which may be more effective than FAQ to reduce errors.

Indeed and this change can easily be done via a script.

bcoconni:
>> I agree. I guess the JSBSim project should lead as an example meaning we should update all the models in the `aircraft\` folder with `<mass_balance negated_inertia="true">` ?
>
>Yes, I've thought about it too. I think it's necessary to modify all aircraft configuration files. Many people will learn or modify based on the existing aircraft configuration file. Adding "negative_crossproduct_inertia" will help remind users to pay attention to the sign of ixz, which may be more effective than FAQ to reduce errors.

This is now done in the commit 9a1b7db40cef6dd4c5c3e334d75c56e80356d6b8. I also pushed to the repo the Python script [`admin/XML_mass_update.py`](https://github.com/JSBSim-Team/jsbsim/blob/master/admin/XML_mass_update.py) that I used to automatically update the aircraft files. It might be useful someday.

PR #508, opened 2021-10-17
Fixed property "aero/h_b-mac-ft" algorithm error

likangbei:
The algorithm of property "aero/h_b-mac-ft" in JSBSim is wrong. This error will lead to an error in the calculation of aircraft ground effect. The code in FGAuxiliary.cpp is：
```
  hoverbcg = in.DistanceAGL / in.Wingspan;

  FGColumnVector3 vMac = in.Tb2l * in.RPBody;
  hoverbmac = (in.DistanceAGL + vMac(3)) / in.Wingspan;
```
In fact, the Z axis of the **body frame** and **NED frame** are downward. Assuming that the aircraft stops on the ground, when MAC is higher than the CG, vMac(3) is negative, resulting in hoverbmac less than hoverbcg. In fact, hoverbmac should be greater than hoverbcg. Therefore, vMac(3) should be inverted，The correct code should be:
```
  hoverbcg = in.DistanceAGL / in.Wingspan;

  FGColumnVector3 vMac = in.Tb2l * in.RPBody;
  hoverbmac = (in.DistanceAGL - vMac(3)) / in.Wingspan;
```
Please confirm and check whether the algorithm is correct.

### DIFF
--- a/src/models/FGAuxiliary.cpp
+++ b/src/models/FGAuxiliary.cpp
@@ -223,7 +223,7 @@ bool FGAuxiliary::Run(bool Holding)
   hoverbcg = in.DistanceAGL / in.Wingspan;
 
   FGColumnVector3 vMac = in.Tb2l * in.RPBody;
-  hoverbmac = (in.DistanceAGL + vMac(3)) / in.Wingspan;
+  hoverbmac = (in.DistanceAGL - vMac(3)) / in.Wingspan;
 
   return false;
 }


seanmcleod:
Yes this does look like a bug. The `AeroRP` (which actually doesn't need to be the ac, is a separate issue) is translated from the body frame of reference to the local (NED) frame of reference, via `Tb2l`, which also has it's Down axis as positive pointing down.

I double-checked the 737 model in terms of it's `AeroRP` position (z = +24in) relative to the `cg` (z = -35.1in) in the structural frame and set a break-point to check it's z-coordinate in the local frame when sitting on the ground.

I didn't notice this issue when I did some ground-effect testing a while ago - https://seanmcleod.github.io/2018/02/landing-ground-effect-flare-jsbsim/

Given the table used in the 737 model you don't end up seeing a negative factor.

```xml
        <function name="aero/function/kCLge">
            <description>Change_in_lift_due_to_ground_effect</description>
            <product>
                <table>
                    <independentVar>aero/h_b-mac-ft</independentVar>
                    <tableData>
                      0             1.203       
                      0.1           1.127       
                      0.15          1.09        
                      0.2           1.073       
                      0.3           1.046       
                      0.4           1.055       
                      0.5           1.019       
                      0.6           1.013       
                      0.7           1.008       
                      0.8           1.006       
                      0.9           1.003       
                      1             1.002       
                      1.1           1           
                    </tableData>
                </table>
            </product>
        </function>
```
What's interesting is that in some Boeing 737 simulator documentation I came across they actually use the height of the bottom of the main gear tire above the ground.

image

likangbei:
@seanmcleod 
In the code：
```
FGColumnVector3 vMac = in.Tb2l * in.RPBody;
hoverbmac = (in.DistanceAGL - vMac(3)) / in.Wingspan;
```
`in.RPBody` is defined in the body frame,  Z-axis is directed downward
`vMac(3)` is defined in the NED frame，Z-axis is also directed downward
But `in.DistanceAGL` and `hoverbmac` are directed upwards.
Therefore, `vMac(3)` is different from others and is opposite. We need to reverse it.
You may have forgotten that in the NED frame, The height is equal to a negative Z value

likangbei:
>What's interesting is that in some Boeing 737 simulator documentation I came across they actually use the height of the bottom of the main gear tire above the ground.

Use the height of the bottom of the main gear tire above the ground， is because the radio altimeter measures the ground clearance according to the lowest point of the aircraft structure (usually the main landing gear during normal take-off and landing)

PR #559, opened 2022-01-06
Remove engine code that reads in position and orientation 

seanmcleod:
As mentioned in https://github.com/JSBSim-Team/jsbsim/discussions/547#discussioncomment-1863391 the engine position and orientation data is read in but isn't used, it's only the thruster's location and orientation that matter, which causes confusion.

### DIFF
--- a/src/models/propulsion/FGEngine.cpp
+++ b/src/models/propulsion/FGEngine.cpp
@@ -56,8 +56,6 @@ FGEngine::FGEngine(int engine_number, struct Inputs& input)
   : in(input), EngineNumber(engine_number)
 {
   Type = etUnknown;
-  X = Y = Z = 0.0;
-  EnginePitch = EngineYaw = 0.0;
   SLFuelFlowMax = 0.0;
   FuelExpended = 0.0;
   MaxThrottle = 1.0;
@@ -109,18 +107,6 @@ unsigned int FGEngine::GetSourceTank(unsigned int i) const
   } else {
     throw("No such source tank is available for this engine");
   }
-}
-
-//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-
-void FGEngine::SetPlacement(const FGColumnVector3& location,
-                            const FGColumnVector3& orientation)
-{
-  X = location(eX);
-  Y = location(eY);
-  Z = location(eZ);
-  EnginePitch = orientation(ePitch);
-  EngineYaw = orientation (eYaw);
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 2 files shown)


src/models/propulsion/FGEngine.cpp
```diff
@@ -200,19 +186,13 @@ bool FGEngine::Load(FGFDMExec *exec, Element *engine_element)
...
-// Jon: The engine location is not important - the nozzle location is.
-
-  local_element = parent_element->FindElement("orient");
-  if (local_element)  orientation = local_element->FindElementTripletConvertTo("RAD");
-//  else          cerr << "No engine orientation found for this engine." << endl;
-// Jon: The engine orientation has a default and is not normally used.
+  // If engine location and/or orientation is supplied issue a warning since they
+  // are ignored. What counts is the location and orientation of the thruster.
+  if (parent_element->FindElement("location"))
+    cerr << "Engine location ignored, only thruster location is used." << endl;
```
bcoconni, line 192:
```suggestion
  local_element = parent_element->FindElement("location");
  if (local_element)
    cerr << local_element->ReadFrom()
         << "Engine location ignored, only thruster location is used." << endl;
```
It should display the error location (file name and line number) above the error message to help the user locating the cause of the warning.

I know BTW that `::ReadFrom` is an ugly name for that method, at some point we should find a better, more explicit name.

src/models/propulsion/FGEngine.cpp
```diff
@@ -200,19 +186,13 @@ bool FGEngine::Load(FGFDMExec *exec, Element *engine_element)
...
 
-  SetPlacement(location, orientation);
+  if (parent_element->FindElement("orient"))
+    cerr << "Engine orientation ignored, only thruster orientation is used." << endl;
```
bcoconni, line 195:
```suggestion
  local_element = parent_element->FindElement("orient");
  if (local_element)
    cerr << local_element->ReadFrom()
         << "Engine orientation ignored, only thruster orientation is used." << endl;
```
Same as above.

bcoconni (changes requested):
Thanks for the PR.
I'd suggest using the method `Element::ReadFrom` in the warning messages to help the user locating the origin of the warnings.

bcoconni:
Also the test `CheckScripts` displays 10 times the warning message `Engine location ignored, only thruster location is used.` and 10 times the warning message `Engine orientation ignored, only thruster orientation is used.`. I guess we should remove these elements from our aircraft definitions in order to avoid confusion when users review our aircraft models as examples.

The changes I have suggested above for the C++ code of the PR should help locating these now useless elements `<location>` and `<orientation>`.

src/models/propulsion/FGEngine.cpp
seanmcleod: Pushed a commit to uses local_element->ReadFrom()

src/models/propulsion/FGEngine.cpp
seanmcleod: Pushed a commit to uses local_element->ReadFrom()

seanmcleod:
> I guess we should remove these elements from our aircraft definitions in order to avoid confusion

Correct, I was planning on doing that next once this pull request is approved and merged in.

bcoconni:
> > I guess we should remove these elements from our aircraft definitions in order to avoid confusion
> 
> Correct, I was planning on doing that next once this pull request is approved and merged in.

All right.

PR #560, opened 2022-01-07
BLDC engine added

pbecchi:
This pull request is relative  to  #333 and propose a new BrushLess DC motor to be incorporated as alternative to basic electric motor. BLDC motor code is based on basic "3 constant motor equations" and allow a good phisical model of motor performances.

It require 3 basic motor properties that are normally available from manufactures or easy to measure:
- Kv speed motor constant
- Rm internal coil resistance
- I0   no load current

In addition a braking time constant has been added to allow change of braking torque:

 ```<velocityconstant> 960 <velocityconstant>
  <torqueconstant> 6.1217 </torqueconstant>
  <coilresistance> 0.117 </coilresistance>
  <noloadcurrent> 0.45 </noloadcurrent>
  <maxvolts> 14.63 </maxvolts>
  <decelerationTime> 2.0 </decelerationTime>
````
Torque constant is just a unit conversion factor and maxvolts is battery voltage available. This voltage is nominal battery voltage that can be factored to consider internal and wires resistance. Next step could be to add a module that represent batteries (and wires) and that will output voltage as function of nominal battery data and SOC (state of charge).

This pull request include the source code files that have been modified as well as **several** examples.
Few examples have been tested and run OK. Some other are just an heritage from repositories from where have been taken @mvacanti and @rega0051 forks.  May be they are unnecessary and it will be cleaner to remove it. I am not expert of PR procedures, so let me know what shoud be included.

### CODE REVIEW

**Comment by seanmcleod** [comment]
I'd recommend only committing the core BLDC engine code in the initial pull request. Will make it easier for others to see/find the relevant code when looking at the commits without having to wade through tons of additional files related to aircraft specific examples etc.

Rather submit the aircraft example making use of the BLDC engine as a separate commit/pull request.

Also take a look at your CR-LF settings, I see a lot of the diffs are showing control-M characters.

Lastly also match the coding standards in terms of the number of spaces for indentation.

**Comment by rega0051** [comment]
I wouldn't want to push the F450 into the JSBSim repo. I still haven't validated it against flight data. And there are a bunch of commented out blocks, which I think is just more confusing. I also have Aero.xml, Gear.xml, ... as separate files because of how I auto generate config files and select amongst options. A clean version of a monolithic F450.xml would be appropriate in the future.

Also, this has a U of MN UAV Lab image, and I'm pretty sure the .ac file for FlightGear is actually an UltraStick25e...

**Comment by bcoconni** [comment]
Thanks for the PR @pbecchi :+1: 
As @seanmcleod and @rega0051 mentioned it needs to be split into several PRs to avoid reviewing a huge commit.

There is a bit of cleanup involved but you should start with a PR containing only the C++ code with the corresponding update to the build files (`CMakeLists.txt` and Visual Studio project files).

**Comment by pbecchi** [comment]
@seanmcleod 
in the last 2 commit I have cleaned up as requested 
 and adjusted the c++ code indentation and line returns (I am not sure if it is what you expect...)
@bcoconni 
It is OK to delete all examples , but without a user manual or input description , it is hard for future users to use BLDC motor.
@rega0051 
dji f450 is a very good example that I have tested using PX4_sitl and , I think, perform very well.
As far as i know, it  is the only multicopter PX4_sitl example that use jsbsim and represent motors as real electric motors.
If you agree , I can reformat the all thing and remove all references to your original files.
I think  a good example is very usefull.

**Comment by bcoconni** [comment]
> It is OK to delete all examples , but without a user manual or input description , it is hard for future users to use BLDC motor.

Sorry if I did not make myself clear but I meant that the example files should be merged in a separate PR for the only purpose of making this PR smaller and easier to review. I agree with you that including examples and documentation will be needed ...but in a separate PR.

File: src/models/propulsion/FGBldc.cpp [hunk 1]
```diff
@@ -0,0 +1,224 @@
...
+
+    // this property is not necessary since is computed using other properties
+    // if (el->FindElement("maxcurrent"))
+    // MaxCurrent= el->FindElementValueAsNumber("maxcurrent");
+
+      if (el->FindElement("maxvolts"))
+        MaxVolts= el->FindElementValueAsNumber("maxvolts");
+
+      if (el->FindElement("velocityconstant"))
+        VelocityConstant= el->FindElementValueAsNumber("velocityconstant");
```
bcoconni, line 62:
```suggestion
      if (el->FindElement("velocityconstant"))
        VelocityConstant= el->FindElementValueAsNumberConvertTo("velocityconstant", ***);
```
Given that `velocityconstant` seems to be of the same dimension than a current, is `velocityconstant` a proper name for that parameter ?
Please request a unit by using `FindElementValueAsNumberConvertTo` instead of `FindElementValueAsNumber`.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+      exec->GetPropertyManager()->Tie(base_property_name + "/power-hp", &HP);
+
+      exec->GetPropertyManager()->Tie(base_property_name + "/current-a", &CurrentRequired);
+
+      Debug(0); // Call Debug() routine from constructor if needed
+}
+
+//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+FGBldc::~FGBldc()
```
bcoconni, line 89:
```suggestion
FGBrushlessDCMotor::~FGBrushlessDCMotor()
```
Please rename the class with a more explicit name.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+    {
+      Debug(1); // Call Debug() routine from constructor if needed
+}
+
+//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+void FGBldc::Calculate(void)
```
bcoconni, line 96:
```suggestion
void FGBrushlessDCMotor::Calculate(void)
```
Explicit class name.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+    V = MaxVolts * in.ThrottlePos[EngineNumber];
+  
+//  Commanded RPM = (input voltage - currentRequired * coil resistance) * velocity costant
+    CommandedRPM = (V - CurrentRequired * coilResistance) * VelocityConstant;
+    DeltaRPM = round((CommandedRPM - RPM));
+
+//  Torque is MaxTorque (stall torque) at 0 RPM and linearly go to 0 at max RPM (MaxVolts*VelocityCostant)
+//  MaxTorque = MaxCurrent*torqueconstant/velocityconstant*(1-RPM/maxRPM)
+
+    MaxTorque = MaxCurrent / VelocityConstant * TorqueConstant * (1 - RPM / (MaxVolts* VelocityConstant));
```
bcoconni, line 122: Is `RPM / (MaxVolts* VelocityConstant)` dimensionless ?

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    TorqueAvailable = MaxTorque - TorqueRequired;
+    DeltaTorque = (((DeltaRPM/60)*(2.0 * M_PI))/(max(0.00001, in.TotalDeltaT))) * ((FGPropeller*)Thruster)->GetIxx();
```
bcoconni, line 125:
```suggestion
    DeltaTorque = ((DeltaRPM* M_PI / 30.))/(max(0.00001, in.TotalDeltaT))) * ((FGPropeller*)Thruster)->GetIxx();
```
Please simplify the constants.
What is the purpose of `max(0.00001, in.TotalDeltaT)` ?

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+    // MaxCurrent= el->FindElementValueAsNumber("maxcurrent");
+
+      if (el->FindElement("maxvolts"))
+        MaxVolts= el->FindElementValueAsNumber("maxvolts");
+
+      if (el->FindElement("velocityconstant"))
+        VelocityConstant= el->FindElementValueAsNumber("velocityconstant");
+
+      if (el->FindElement("torqueconstant"))
+        TorqueConstant= el->FindElementValueAsNumber("torqueconstant");
```
bcoconni, line 65: Same as above: please specify a unit using `FindElementValueAsNumberConvertTo`.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+  
+      if (el->FindElement("coilresistance"))
+          coilResistance = el->FindElementValueAsNumber("coilresistance");
```
bcoconni, line 68: Same as above: please specify a unit using `FindElementValueAsNumberConvertTo`.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+      if (el->FindElement("noloadcurrent"))
+          noLoadCurrent = el->FindElementValueAsNumber("noloadcurrent");
```
bcoconni, line 70: Same as above: please specify a unit using `FindElementValueAsNumberConvertTo`.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+      if (el->FindElement("decelerationTime"))
+          deceleration_time = el->FindElementValueAsNumber("decelerationTime");
```
bcoconni, line 72: Same as above: please specify a unit using `FindElementValueAsNumberConvertTo`.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+//  Acceleration is due to the max delta torque available and is limited to the inertial forces
+
+    if (DeltaRPM >= 0) {
+      TargetTorque = min(DeltaTorque, TorqueAvailable) + TorqueRequired;
+    } else {
+//  Deceleration is due to braking force given by the ESC and set by parameter deceleration_time 
+      TargetTorque = TorqueRequired  - min(abs(DeltaTorque)/(max(deceleration_time,0.01)*30),RPM*TorqueConstant/VelocityConstant/VelocityConstant/coilResistance);
+    }
+
+    EnginePower = ((2 * M_PI) * max(RPM, 0.0001) * TargetTorque) / 60;
```
bcoconni, line 137: What is the purpose of `max(RPM, 0.0001)` ? What's wrong if `EnginePower` gets null ?

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+    HP = EnginePower / 550;
```
bcoconni, line 138:
```suggestion
    HP = EnginePower / hptowatt;
```
EnginePower is in Watt, isn't it ? In which case, the conversion should use `hptowatt`.

src/models/propulsion/FGBldc.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    RunPostFunctions();
+}
+
+//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+double FGBldc::CalcFuelNeed(void)
+{
+  return 0;
+}
```
bcoconni, line 150:
```suggestion
```
Trivial methods must be inlined in the header.

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
...
+public:
+  /// Constructor
+  FGBldc(FGFDMExec* exec, Element *el, int engine_number, FGEngine::Inputs& input);
+  /// Destructor
+  ~FGBldc();
+
+  void Calculate(void);
+  double GetPowerAvailable(void) {return (HP * hptoftlbssec);}
+  double GetCurrentRequired(void) {return CurrentRequired;}
+  double getRPM(void) {return RPM;}
```
bcoconni, line 64:
```suggestion
  double getRPM(void) {return RPM;}
  double CalcFuelNeed(void) { return 0; }
```
The trivial method `CalcFuelNeed` shall be inlined (even though virtual methods cannot be inlined by the linker).

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
+/*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+ Header:       FGBldc.h
+ Author:       Matt Vacanti
+ Date started: 11/08/2020
+
+ ----- Copyright (C) 2020  Matt Vacanti (mdvacanti@gmail.com) --------------
+
+UPDATE LICENSE
```
bcoconni, line 9:
```suggestion
 --------- Copyright (C) 2020-2022  Matt Vacanti / Paolo Becchi-------------

 This program is free software; you can redistribute it and/or modify it under
 the terms of the GNU Lesser General Public License as published by the Free
 Software Foundation; either version 2 of the License, or (at your option) any
 later version.

 This program is distributed in the hope that it will be useful, but WITHOUT
 ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
 FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
 details.

 You should have received a copy of the GNU Lesser General Public License along
 with this program; if not, write to the Free Software Foundation, Inc., 59
 Temple Place - Suite 330, Boston, MA 02111-1307, USA.

 Further information about the GNU Lesser General Public License can also be
 found on the world wide web at http://www.gnu.org.
```
The license shall be LGPL 2.0+. Note that you'll need an agreement from @mvacanti for that.

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
...
+
+/** Models an electric brushless DC motor or more appropriately permanent magnet
+    synchronous motor.
+    FGElectric models an electric motor based on the configuration file
+    \<power> parameter.  The throttle controls motor output linearly from
+    zero to \<power>.  This power value (converted internally to horsepower)
+    is then used by FGPropeller to apply torque to the propeller.  At present
+    there is no battery model available, so this motor does not consume any
+    energy.  There is no internal friction.
+    @author David Culp
```
bcoconni, line 46: Please update this documentation (or remove it).

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
+/*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+ Header:       FGBldc.h
```
bcoconni, line 3:
```suggestion
 Module:       FGBrushlessDCMotor.h
```
Please use a more explicit name for the file such as `FGBrushlessDCMotor.h`

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
...
+CLASS DECLARATION
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
+
+class FGBldc : public FGEngine
+{
+public:
+  /// Constructor
+  FGBldc(FGFDMExec* exec, Element *el, int engine_number, FGEngine::Inputs& input);
+  /// Destructor
+  ~FGBldc();
```
bcoconni, line 59: Please use a more explicit name for the class.
```suggestion
class FGBrushlessDCMotor : public FGEngine
{
public:
  /// Constructor
  FGBrushlessDCMotor(FGFDMExec* exec, Element *el, int engine_number, FGEngine::Inputs& input);
  /// Destructor
  ~FGBrushlessDCMotor();
```
Please use a more explicit name for the class.

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
...
+  void Calculate(void);
+  double GetPowerAvailable(void) {return (HP * hptoftlbssec);}
+  double GetCurrentRequired(void) {return CurrentRequired;}
+  double getRPM(void) {return RPM;}
+  std::string GetEngineLabels(const std::string& delimiter);
+  std::string GetEngineValues(const std::string& delimiter);
+
+private:
+
+  double CalcFuelNeed(void);
```
bcoconni, line 70:
```suggestion
```
The method `CalcFuelNeed` is defined as a `public` method in `FGEngine`. Please do not modify that.

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
...
+  double CommandedRPM;       // desired RPM set by voltage
+  double MaxTorque;          // maximum available torque from motor
+  double DeltaRPM;
+  double TorqueAvailable;
+  double TargetTorque;
+  double TorqueRequired;
+  double CurrentRequired;
+  double EnginePower;
+  double DeltaTorque;
+  double  deceleration_time = 0.5;
```
bcoconni, line 95:
```suggestion
  double DecelerationTime = 0.5;
```
JSBSim class members are using [CamelCase naming](https://en.wikipedia.org/wiki/Camel_case) please match the convention.

src/models/propulsion/FGBldc.h
```diff
@@ -0,0 +1,100 @@
...
+private:
+
+  double CalcFuelNeed(void);
+
+  //double BrakeHorsePower;
+
+  // constants
+  double hptowatts;
+  double noLoadCurrent;
+  double coilResistance;
```
bcoconni, line 77:
```suggestion
  double NoLoadCurrent;
  double CoilResistance;
```
JSBSim class members are using [CamelCase naming](https://en.wikipedia.org/wiki/Camel_case) please match the convention.

bcoconni (changes requested):
Thanks again for the PR @pbecchi 

There is quite a lot of changes that are requested:
1. Cosmetic changes
* Do not use cryptic acronyms such `bldc` for naming. That's OK for a proof of concept but not for production code. So please rename the files `FGBlbc.*` to `FGBrushlessDCMotor.*` and the class `FGBldc` to `FGBrushlessDCMotor`.
* Please match JSBSim code formatting conventions:
  * No tabs.
  * 2 spaces indentations.
  * [Camel case naming](https://en.wikipedia.org/wiki/Camel_case).
  * No trailing spaces.
  * Opening brackets of methods shall be on a new line and the leftmost column.
* Remove outdated/useless comments.
2. Legal changes
* The C++ code shall be released under the [LGPL2.0+ license](https://www.gnu.org/licenses/old-licenses/lgpl-2.0.en.html). You will need Matt Vacanti's agreement for that.
3. Implementation
* I am concerned by the consistency of the units used in your code. The units of `VelocityConstant` are unclear and I am quite skeptical that `RPM / (MaxVolts* VelocityConstant)` is dimensionless (but I'd be glad to be proven wrong :smiley:).
  * Please check the units consistency.
  * Please specify the units using `FindElementValueAsNumberConvertTo` (note that you might need to update the container `convert` in `src/input_output/FGXMLElement.cpp` with new units).
* Are real life brushless DC motor include an RPM controller ? I am surprised to find that the motor torque is computed from a target RPM `commandedRPM`...

bcoconni:
Actually, the units can be consistent provided that:
* `MaxVolts` is supplied in Volts
* `VelocityConstant` is supplied in A/W/min (Amperes per Watt per minute !?!)
* `TorqueConstant` is dimensionless
* `NoLoadCurrent` is supplied in Amperes
* `DecelerationTime` is dimensionless

pbecchi:
@bcoconni 
All format changes will require a little time to be completed.

Regarding Units:
MaxVolts is in [Volts]
VelocityConstant is  RPM/volts so units are [1/(min*volts)]
CoilResistance is in [Ohm]
TorqueConstant is dimensionless and is a fixed factor (may be should be taken out from input parameters)
NoLoadCurrent is in [Amperes]
DecelerationTime is dimensionless since is a factor (I should call it DecelerationFactor)

All those units are somehow "uniques", since there is no alternatives ....
Nevertheless I will try to find an example on how to code Unit conversions.

Regarding name of variables most are heritage from Vacanti code (commanded_rpm in reality is delta_rpm ) .
I will do a full renaming to make code more easy to understand. 
At this point the new .cpp will be completely different from the M.Vacanti one.

another question : there is a standard copyright note I can copy from other files??

bcoconni:
> @bcoconni All format changes will require a little time to be completed.

No problem. There is no rush.
> 
> Regarding Units: MaxVolts is in [Volts] VelocityConstant is RPM/volts so units are [1/(min*volts)] CoilResistance is in [Ohm] TorqueConstant is dimensionless and is a fixed factor (may be should be taken out from input parameters) NoLoadCurrent is in [Amperes] DecelerationTime is dimensionless since is a factor (I should call it DecelerationFactor)
> 
> All those units are somehow "uniques", since there is no alternatives .... Nevertheless I will try to find an example on how to code Unit conversions.

You'll most likely need to add new lines to `src/input_output/FGXMLElement.cpp`:
https://github.com/JSBSim-Team/jsbsim/blob/2bd60cfb9fd4b66a27c0d2eff00d509870cf7123/src/input_output/FGXMLElement.cpp#L119-L121

New lines will be for instance:
```c++
  convert["VOLTS"]["VOLTS"] = 1.0
  convert["OHMS"]["OHMS"] = 1.0
```
You don't need to add unit conversion for a start, just add the units you need just as the example above.
> 
> Regarding name of variables most are heritage from Vacanti code (commanded_rpm in reality is delta_rpm ) . I will do a full renaming to make code more easy to understand.

Yes please.
> At this point the new .cpp will be completely different from the M.Vacanti one.

No problem.
> 
> another question : there is a standard copyright note I can copy from other files??

Yes, see the example below:

https://github.com/JSBSim-Team/jsbsim/blob/2bd60cfb9fd4b66a27c0d2eff00d509870cf7123/src/FGFDMExec.cpp#L2-L29

pbecchi:
@bcoconni 
I made all changes in a single commit working locally with visualstudio.
Only Makefile.txt and measurement units are missing.
Could you check if this commit is ok? Thanks.

bcoconni:
@pbecchi The build fails because you've forgotten to update the `CMakeLists.txt` file.

Please take this opportunity to clean the tabs that you have inserted in `CMakeLists.txt`.
I'll review your code once the build will succeed.

Thanks.

pbecchi:
Done...wait for your comments!

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  Further information about the GNU Lesser General Public License can also be
+  found on the world wide web at http://www.gnu.org
+
+FUNCTIONAL DESCRIPTION
+--------------------------------------------------------------------------------
+Following code represent a new BrushLess DC motor to be used as alternative 
+to basic electric motor. 
+BLDC motor code is based on basic "3 constant motor equations"
+It require 3 basic phisical motor properties:
```
seanmcleod, line 32: phisical - physical

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,242 @@
...
+  Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+
+  Further information about the GNU Lesser General Public License can also be
+  found on the world wide web at http://www.gnu.org
+
+FUNCTIONAL DESCRIPTION
+--------------------------------------------------------------------------------
+Following code represent a new BrushLess DC motor to be used as alternative 
+to basic electric motor. 
+BLDC motor code is based on basic "3 constant motor equations"
```
seanmcleod, line 31: Do you have an online link/reference for the "3 constant motor equations"?

pbecchi: not realy.....

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,242 @@
+/*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+ Module:       FGBrushLessDCMotor.cpp
+ Autor         Paolo Becchi
+ 1st release   1/1/2022
+ Purpose:      This module models an BLDC electric motor
+
+  ------------- Copyright (C) 1999  Paolo Becchi (pbecchi@aerobusinees.it) -------------
```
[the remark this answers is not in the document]
pbecchi: yes 2022

src/models/propulsion/FGBrushLessDCMotor.h
```diff
@@ -0,0 +1,111 @@
...
+  with this program; if not, write to the Free Software Foundation, Inc., 59
+  Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+
+  Further information about the GNU Lesser General Public License can also be
+  found on the world wide web at http://www.gnu.org
+
+
+HISTORY
+--------------------------------------------------------------------------------
+1/01/2022 MDV  Created
```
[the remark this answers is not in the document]
pbecchi: i dont know something copied from another history. To be removed

src/models/FGPropulsion.cpp
```diff
@@ -392,12 +394,17 @@ bool FGPropulsion::Load(Element* el)
...
       } else if (engine_element->FindElement("rocket_engine")) {
-        Element *element = engine_element->FindElement("rocket_engine");
-        Engines.push_back(make_shared<FGRocket>(FDMExec, element, numEngines, in));
+        Element *element = engine_element->FindElement("rocket_engine");        Engines.push_back(make_shared<FGRocket>(FDMExec, element, numEngines, in));
       } else if (engine_element->FindElement("electric_engine")) {
         Element *element = engine_element->FindElement("electric_engine");
         Engines.push_back(make_shared<FGElectric>(FDMExec, element, numEngines, in));
-      } else {
+      } else if (engine_element->FindElement("bldc_engine")) {
+        //  HaveBldcEngine = true;
```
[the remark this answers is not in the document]
pbecchi: already removed!!

src/models/propulsion/CMakeLists.txt
```diff
@@ -1,34 +1,36 @@
-set(SOURCES FGElectric.cpp
```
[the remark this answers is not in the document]
pbecchi: there is no tab......

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,242 @@
...
+to basic electric motor. 
+BLDC motor code is based on basic "3 constant motor equations"
+It require 3 basic phisical motor properties:
+Kv speed motor constant
+Rm internal coil resistance
+I0 no load current
+
+HISTORY
+--------------------------------------------------------------------------------
+1/01/2022  MDV  Created
```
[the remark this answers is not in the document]
pbecchi: remove

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,242 @@
...
+  if (el->FindElement("velocityconstant"))
+    VelocityConstant = el->FindElementValueAsNumber("velocityconstant");
+
+//  if (el->FindElement("torqueconstant"))
+//    TorqueConstant = el->FindElementValueAsNumber("torqueconstant");
+
+  if (el->FindElement("coilresistance"))
+    coilResistance = el->FindElementValueAsNumber("coilresistance");
+  if (el->FindElement("noloadcurrent"))
+    noLoadCurrent = el->FindElementValueAsNumber("noloadcurrent");
```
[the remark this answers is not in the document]
pbecchi: OK

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,242 @@
...
+ 
+  //  compute acceleration and deceleration phases:
+  //  Acceleration is due to the max delta torque available and is limited to the inertial forces
+
+  if (DeltaRPM >= 0) {
+      TargetTorque = min(InertiaTorque, TorqueAvailable) + TorqueRequired;
+  } else {
+  //  Deceleration is due to braking force given by the ESC and set by parameter deceleration_time 
+      TargetTorque = TorqueRequired - min(abs(InertiaTorque)/(max(DecelerationFactor,0.01)*30),RPM*TorqueConstant/VelocityConstant/VelocityConstant/coilResistance);
+  }
```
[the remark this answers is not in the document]
pbecchi: This wait the factor is aproximately  the time delay in seconds

src/models/propulsion/CMakeLists.txt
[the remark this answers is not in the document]
pbecchi: I dont know what to change!!

src/models/FGPropulsion.cpp
```diff
@@ -392,12 +394,17 @@ bool FGPropulsion::Load(Element* el)
         Element *element = engine_element->FindElement("turboprop_engine");
         Engines.push_back(make_shared<FGTurboProp>(FDMExec, element, numEngines, in));
       } else if (engine_element->FindElement("rocket_engine")) {
-        Element *element = engine_element->FindElement("rocket_engine");
-        Engines.push_back(make_shared<FGRocket>(FDMExec, element, numEngines, in));
+        Element *element = engine_element->FindElement("rocket_engine");        Engines.push_back(make_shared<FGRocket>(FDMExec, element, numEngines, in));
```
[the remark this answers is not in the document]
pbecchi: OK

pbecchi:
I made all corrections possible.....

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,236 @@
...
+  InertiaTorque = (((DeltaRPM/60)*(2.0 * M_PI))/(max(0.00001, in.TotalDeltaT))) * ((FGPropeller*)Thruster)->GetIxx();
+ 
+  //  compute acceleration and deceleration phases:
+  //  Acceleration is due to the max delta torque available and is limited to the inertial forces
+
+  if (DeltaRPM >= 0) {
+    TargetTorque = min(InertiaTorque, TorqueAvailable) + TorqueRequired;
+  } else {
+  //  Deceleration is due to braking force given by the ESC and set by parameter deceleration_time 
+    TargetTorque = TorqueRequired - min(abs(InertiaTorque)/(max(DecelerationFactor,0.01)*30),RPM*TorqueConstant/VelocityConstant/VelocityConstant/CoilResistance);
```
seanmcleod, line 146: > This wait the factor is approximately the time delay in seconds

Rather make this a constant with an appropriate name and comment so that anyone else looking at the code knows what it is.

src/models/propulsion/FGBrushLessDCMotor.h
```diff
@@ -0,0 +1,111 @@
...
+
+  // constants
+  double hptowatts;
+  double NoLoadCurrent;      // 0 torque current [A]
+  double CoilResistance;     // internal resistance [Ohm]
+  double PowerWatts;         // maximum engine power
+  double MaxCurrent;         // maximum current [A]
+  double MaxVolts;           // max voltage available from battery [V]
+  double VelocityConstant;   //.speed constant of brusless DC motors [RPM/V]
+  double TorqueConstant=6.1217;
```
seanmcleod, line 94: What is `TorqueConstant`?

src/models/propulsion/FGBrushLessDCMotor.h
```diff
@@ -0,0 +1,111 @@
...
+
+/** Models an electric brushless DC motor or more appropriately permanent magnet
+    synchronous motor.
+    FGElectric models an electric motor based on the configuration file
+    \<power> parameter.  The throttle controls motor output linearly from
+    zero to \<power>.  This power value (converted internally to horsepower)
+    is then used by FGPropeller to apply torque to the propeller.  At present
+    there is no battery model available, so this motor does not consume any
+    energy.  There is no internal friction.
+    @author David Culp
```
bcoconni, line 61: I'm pretty sure this documentation has been copied-pasted from `FGElectric`. Please update it or remove it.

In addition I'd expect that you document the parameters that `<brushless_dc_motor>` is expecting such as:
```xml
@code
  <brushless_dc_motor>
    <maxvolts units="VOLTS">        {number} </maxvolts>
    <velocityconstant>              {number} </velocityconstant>
    <coilresistance units="OHMS">   {number} </coilresistance>
    <noloadcurrent units="AMPERES"> {number} </noloadcurrent>
    <decelerationfactor>            {number} </decelerationfactor>
  </brushless_dc_motor>
@endcode
```
And explain their meaning - see `src/models/propulsion/FGEngine.h` for example:
https://github.com/JSBSim-Team/jsbsim/blob/db4d967b8e441ce0406b0c5baaad3201294386a9/src/models/propulsion/FGEngine.h#L67-L97

src/models/propulsion/FGBrushLessDCMotor.h
bcoconni:
```suggestion
  constexpr double TorqueConstant = 6.1217;
```
If it's kept as is then it should be defined as `constexpr`.

src/models/propulsion/FGBrushLessDCMotor.h
```diff
@@ -0,0 +1,111 @@
...
+  double GetCurrentRequired(void) {return CurrentRequired;}
+  double getRPM(void) {return RPM;}
+  double CalcFuelNeed(void);
+  std::string GetEngineLabels(const std::string& delimiter);
+  std::string GetEngineValues(const std::string& delimiter);
+
+private:
+
+  // constants
+  double hptowatts;
```
bcoconni, line 87:
```suggestion
  constexpr double hptowatts = 745.7;
```
I'm pretty sure it would be better if this variable is defined as a `constexpr` (in which case you'll need to remove its initialization in the constructor otherwise the compiler will complain.

src/models/propulsion/FGBrushLessDCMotor.h
```diff
@@ -0,0 +1,111 @@
...
+  /// Constructor
+  FGBrushLessDCMotor(FGFDMExec* exec, Element *el, int engine_number, FGEngine::Inputs& input);
+  /// Destructor
+  ~FGBrushLessDCMotor();
+
+  void Calculate(void);
+  double GetPowerAvailable(void) {return (HP * hptoftlbssec);}
+  double GetCurrentRequired(void) {return CurrentRequired;}
+  double getRPM(void) {return RPM;}
+  double CalcFuelNeed(void);
```
bcoconni, line 80:
```suggestion
  double CalcFuelNeed(void) { return 0.; }
```
Since this method is trivial then it should better be defined in the header (just as `getRPM()` above).

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,236 @@
...
+  Type = etElectric;
+  PowerWatts = 745.7;
+  hptowatts = 745.7;
+
+
+  if (el->FindElement("maxvolts"))
+    MaxVolts = el->FindElementValueAsNumber("maxvolts");
+
+  if (el->FindElement("velocityconstant"))
+    VelocityConstant = el->FindElementValueAsNumber("velocityconstant");
```
bcoconni, line 77:
```suggestion
  if (el->FindElement("velocityconstant"))
    VelocityConstant = el->FindElementValueAsNumber("velocityconstant");
  else {
    cerr << el->ReadFrom()
         << "<velocityconstant> is a mandatory parameter" << endl;
    throw JSBBaseException("Missing parameter");
  }
```
There is no default value for `VelocityConstant` so you must throw if the user does not specify it.

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  if (el->FindElement("coilresistance"))
+    CoilResistance = el->FindElementValueAsNumber("coilresistance");
+  if (el->FindElement("noloadcurrent"))
+    NoLoadCurrent = el->FindElementValueAsNumber("noloadcurrent");
+  if (el->FindElement("deceleration_factor"))
+    DecelerationFactor = el->FindElementValueAsNumber("deceleration_factor");
```
bcoconni, line 84:
```suggestion
  if (el->FindElement("deceleration_factor"))
    DecelerationFactor = el->FindElementValueAsNumber("deceleration_factor");
  else {
    cout << el->ReadFrom()
         << "Using default value " << DecelerationFactor << " for <deceleration_factor>" << endl;
  }
```
This one is different, there is a default value `1.0` so you must warn the user that JSBSim is using a default value and continue processing (no throw this time :smile:)

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  MaxCurrent = MaxVolts / CoilResistance + NoLoadCurrent;
+
+  PowerWatts = MaxCurrent * MaxVolts;
+
+  string base_property_name = CreateIndexedPropertyName("propulsion/engine",
+    EngineNumber);
+  exec->GetPropertyManager()->Tie(base_property_name + "/power-hp", &HP);
+
+  exec->GetPropertyManager()->Tie(base_property_name + "/current-a", &CurrentRequired);
```
bcoconni, line 94:
```suggestion
  exec->GetPropertyManager()->Tie(base_property_name + "/current-a", &CurrentRequired);
```
According to your comment, wouldn't this property be better named `current-required_amperes` or something similar ?

bcoconni (changes requested):
We're almost getting there @pbecchi. A few additional comments (sorry !) mostly to polish the details.

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  if (DeltaRPM >= 0) {
+    TargetTorque = min(InertiaTorque, TorqueAvailable) + TorqueRequired;
+  } else {
+  //  Deceleration is due to braking force given by the ESC and set by parameter deceleration_time 
+    TargetTorque = TorqueRequired - min(abs(InertiaTorque)/(max(DecelerationFactor,0.01)*30),RPM*TorqueConstant/VelocityConstant/VelocityConstant/CoilResistance);
+  }
+
+  EnginePower = ((2 * M_PI) * max(RPM, 0.0001) * TargetTorque) / 60;   //units [#*ft/s]
+  HP = EnginePower / 550; // units[HP]
```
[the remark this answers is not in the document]
pbecchi: engine power is in  pounds*ft/s since torque is in ft*pounds. So conversion factor is not 745.7. but 745/1.36.
I added the comments with [units] to make all this understandable......... 
I'd rather prefer to use only metric units...

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+FUNCTIONAL DESCRIPTION
+--------------------------------------------------------------------------------
+Following code represent a new BrushLess DC motor to be used as alternative 
+to basic electric motor. 
+BLDC motor code is based on basic "3 constant motor equations"
+It require 3 basic physical motor properties:
+Kv speed motor constant
+Rm internal coil resistance
+I0 no load current
```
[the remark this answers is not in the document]
pbecchi: I could find this https://electronics.stackexchange.com/questions/33315/understanding-motor-constants-kt-and-kemf-for-comparing-brushless-dc-motors that give some explanation on Motor Constants and their meaning.
KV [RPM/V] is well known property of a Brushless DC motor. It is related to Kt Torque constant by the relation Kt=1/KV[r/s/V] where torque is computed in NM.  To make things more clear I will modify .cpp and .h to define conversion factor in explicit way as you suggest .

pbecchi:
I have added units to motor speed constant Kv, i think it is necessary to avoid confusion.
Do you want me to add it also to FGXMLElement.cpp?
The last commits should close all remaining comments......at least i hope.....time is getting very short!

src/models/propulsion/FGBrushLessDCMotor.cpp
bcoconni: `RPM*TorqueConstant/VelocityConstant/VelocityConstant/CoilResistance)` describes the energy dissipated by the Joule effect, doesn't it ?

bcoconni (approved):
OK now there has been a major improvement regarding units. The code looks much slicker and, with the exception of `DecelerationFactor`, it is no longer using magic factors.

And just for clarification: the formula `RPM*TorqueConstant/VelocityConstant/VelocityConstant/CoilResistance)` models the energy dissipated by the Joule effect, no ? But in that case, why isn't the Joule effect modeled during acceleration as well ? After all, the dissipation takes place as long as the current and the voltage are non-zero.

I'd suggest to keep these topics open for a later improvement.

PR #562, opened 2022-01-08
Remove unused engine location and orientation from aircraft files

seanmcleod:


bcoconni (approved):
Wow ! That's a big change :+1: Thanks !

seanmcleod:
Mostly laborious, I was in 2 minds whether to write a once of script along the lines of the one you wrote for the moment of inertia update or just to edit them all manually 😉

bcoconni:
Well that was exactly the point of including the script in the repo: so that it could be re-used when similar changes need to be made. Personally, I'd have hacked the Python script :wink:

PR #563, opened 2022-01-08
Remove engine location and orientation elements

seanmcleod:
I think this is the final bit in terms of updating the code, example aircraft files and the schema definition.


bcoconni (approved):
At the moment, it's for the sake of completeness because this XSD schema is not yet used in the CI.

PR #574, opened 2022-01-21
DJI F450 quadcopter alpha1 release

pbecchi:
This PR contain an  example for testing brushless DC motor. The example is a DJI F450 quadcopter FDM.


seanmcleod:
Thanks @pbecchi  this is much easier to to review.

Very initial comment, it looks like roughly half the files have `^M` at the end of each line and half don't, looks like some mixture of `LF` versus `LF + CR`?

pbecchi:
Yes it could be since I work on Windows and Visual Studio , I will try to correct all those at first commit.....
I see there is also some automatic check errors, but I dont understand what they are about.....

seanmcleod:
I also work with Visual Studio on Windows. Visual Studio will typically handle files with either `CRLF` or just `LF` and keep it that way, and prompt you when you open a file which uses a mixture asking if you want to normalize them. But there is then also your git settings in terms of how it handles line endings, potentially converting on the fly when pushing to the repo and pulling etc.

However I think I possibly saw these `^M` in @rega0051's repo originally.

seanmcleod:
In terms of the build failures I noticed that `TestInitialConditions.py` is failing, with a unit lookup failing. I seem to remember that the commit for the BLDC included some new unit conversions for electrical properties, maybe one of the existing unit conversions was accidently deleted?

```sh
ERROR: test_initial_conditions_v1 (__main__.TestInitialConditions)
  File "D:/a/jsbsim/jsbsim/tests/TestInitialConditions.py", line 120, in test_initial_conditions_v1
    self.CheckICValues(vars, 'script %s' % (f,), fdm, IC_root)
  File "D:/a/jsbsim/jsbsim/tests/TestInitialConditions.py", line 168, in CheckICValues
    conv = var['unit'][var_tag.attrib['unit']]
KeyError: 'M/S'
```

engine/DJI_E305.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<!--bldc_engine name="DJI E305"-->
+  <brushless_dc_motor name="DJI E305">
+  <!-- E305 Motor is a DJI 2312E - 960 -->
+  <power unit="WATTS"> 320 </power>
+  <velocityconstant> 960 </velocityconstant>
+  <torqueconstant> 6.1217 </torqueconstant>
```
seanmcleod, line 7: Bit confused about `<torqueconstant>`. Is this supposed to be a user supplied value for `Kq` as per - http://web.mit.edu/drela/Public/web/qprop/motor1_theory.pdf

Although this element isn't read in by `FGBrushLessDCMotor::FGBrushLessDCMotor()`, instead we have the following in the header file.

```c++
constexpr double TorqueConstant = 60 / (2 * M_PI * NMtoftpound);
```

So what is the relationship between them?

pbecchi:
Let me answer your questions:
I will fix all line ending in next commit
I forgot to delete TorqueCostant from motor.xlm : you are correct it is not needed anymore.
Regarding unit let me ask you a stupid question( since I still have to learn......a lot ):
 Since we added unit definitions on C++ code do we need to add the unit on the properties defined in motor.xml? So  for example, I need  to write:
`<maxvolts unit="VOLTS"> 14.6 </maxvolts>`??

seanmcleod:
> I forgot to delete TorqueCostant from motor.xlm : you are correct it is not needed anymore.

Is this because you're working on the assumption that `Kq == Kv`? 

From [First-Order DC Electric Motor Model](http://web.mit.edu/drela/Public/web/qprop/motor1_theory.pdf)

> The torque constant can be simply assumed to be the same as Kv
> Alternatively, it can be obtained from motor torque data if this is available.

If so, why not allow the user to specify it optionally, and if it isn't supplied then assume `Kq == Kv`?

> Since we added unit definitions on C++ code do we need to add the unit on the properties defined in motor.xml?

Currently you only have a single set of units for each electrical type, so at the moment it's really optional.

engine/DJI_E305.xml
pbecchi: According to theory Kq (torque constant) can be assumed to be 1/Kv (speed constant ) if we define speed  in rad/sec torque in Newton*Meter. 
Therefore since torque is in ft*pound a unit conversion factor is necessary (the conversion factor is `TorqueConstant`)

seanmcleod: The naming is confusing, nomenclature from the motor theory document:

image

However your variable, `TorqueConstant` is a constant conversion factor. Whereas when I was initially reading the source code and comparing it to the motor model document I assumed your `TorqueConstant` was referring to Kq.

image

Also given this:

image

Shouldn't we allow the user to optionally specify the torque constant Kq if they want to model the efficiency more accurately?

pbecchi: Yes i agree....

pbecchi:
Yes i agreeOn Jan 23, 2022 5:42 PM, Sean McLeod ***@***.***> wrote:
@seanmcleod commented on this pull request.



In engine/DJI_E305.xml:
> @@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+
+  <brushless_dc_motor name="DJI E305">
+  
+  <power unit="WATTS"> 320 </power>
+  <velocityconstant> 960 </velocityconstant>
+  <torqueconstant> 6.1217 </torqueconstant>

The naming is confusing, nomenclature from the motor theory document:

However your variable, TorqueConstant is a constant conversion factor. Whereas when I was initially reading the source code and comparing it to the motor model document I assumed your TorqueConstant was referring to Kq.

Also given this:

Shouldn't we allow the user to optionally specify the torque constant Kq if they want to model the efficiency more accurately?

bcoconni:
I have fixed the conflicts by merging `master` to the PR. Now it builds successfully.

aircraft/F450/F450-set.xml
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0"?>
+
```
bcoconni, line 2: This file `F450-set.xml` should be removed as it is only meaningful to FlightGear.

engine/DJI_E305.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<!--bldc_engine name="DJI E305"-->
+  <brushless_dc_motor name="DJI E305">
+  <power unit="WATTS"> 320 </power>
```
bcoconni, line 4: Unles I'm mistaken `power` is not used by `FGBrushlessDCMotor`.

engine/DJI_E305.xml
```diff
@@ -0,0 +1,10 @@
...
+  <velocityconstant> 960 </velocityconstant>
+  <coilresistance>0.117</coilresistance>
+  <noloadcurrent>0.45</noloadcurrent>
+  <maxvolts> 14.63 </maxvolts>
+  <deceleration_factor> 5</deceleration_factor>
```
bcoconni, line 9: Unless I'm mistaken `deceleration_factor` is not used by `FGBrushlessDCMotor`.

bcoconni (changes requested):
Thanks for the PR. This will give a means to test the brushless DC motor model.
A bit of cleanup is needed before merging:
* Please remove the file `aircraft/F450/F450-set.xml` as it is not used by JSBSim.
* Please remove unused parameters from `engine/DJI_E305.xml` (i.e. the parameters `power` and `deceleration_factor`).

engine/DJI_E305.xml
pbecchi: Resolved with commit Corrections

pbecchi: Resolved with commit Corrections

aircraft/F450/F450-set.xml
pbecchi: resolved with commit corrections

engine/DJI_E305.xml
pbecchi: resolved with commit corrections

bcoconni (approved):

bcoconni:
PR merged. Thanks for your contribution.

Issue #494, opened 2021-08-13
FGTurbine don't start windmilling until powered up at least once


Octal450:
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**

The engine stays at 0% N1/N2 and does not windmill unless it is started at least one time

**What is the current behavior?**

To reproduce start a simulation with FGTurbine but do not start one of the engines (example, in a twinjet, start only 1 engine), then takeoff and accelerate. Notice the engine stays at 0% N1 and N2. Then if you engage the starter and then shut down to get slightly rotation, then it seeks the windmill value properly

**What is the expected behavior?**

Engine should always windmill when off

**What is the motivation / use case for changing the behavior?**
So that he will work realistic

**Please tell us about your environment:**

seanmcleod:
Taking an initial look at the code in FGTurbine, it looks like windmilling is only implemented as part of the engine shutting down, i.e. 

https://github.com/JSBSim-Team/jsbsim/blob/08d2f857b407b5501fe8bba5c245c1051c606c60/src/models/propulsion/FGTurbine.cpp#L172-L183

That's the only reference to windmilling in FGTurbine excluding the lookup to see if windmilling is disabled.

https://github.com/JSBSim-Team/jsbsim/blob/08d2f857b407b5501fe8bba5c245c1051c606c60/src/models/propulsion/FGTurbine.cpp#L477-L478

bcoconni:
Yes, but the variable `phase` is initialized to `tpOff`:
https://github.com/JSBSim-Team/jsbsim/blob/08d2f857b407b5501fe8bba5c245c1051c606c60/src/models/propulsion/FGTurbine.cpp#L98
so `FGTurbine::Off` should be called even if the engine has not yet been started:
https://github.com/JSBSim-Team/jsbsim/blob/08d2f857b407b5501fe8bba5c245c1051c606c60/src/models/propulsion/FGTurbine.cpp#L154-L162
However `phase` is not initialized in the constructor so that might be the cause of the problem.

@Octal450, does the issue vanish if you initialize `phase = tpOff` in the `FGTurbine` constructor ?

Octal450:
I'll try my best to test this tonight or tomorrow. A little low on time. 

Thanks Bertrand!
Josh

Octal450:
Hi @bcoconni, no difference. 
[...]

Issue #233, opened 2019-11-11
Turbine: ignitionn2


NikolaiVChr:
Hey, I have a turbine in FG 2019.1

When I set ignitionn2 N2 still goes to 25.18, but N1 goes to what ignitionn2 is set it to, has this been fixed in later JSBSim versions?

I just read the code:
if (el->FindElement("ignitionn2"))
    IgnitionN1 = el->FindElementValueAsNumber("ignitionn2");

legoboyvdlp:
oops...

Issue #196, opened 2019-07-08
Turbine fuel consumption


mike402:
Hi,

I found the following problems with turbine fuel flow.

1. `fuel-flow-rate-pps` considerably lags behind the change of `thrust-lbs` -- it takes almost half a minute to settle after thrust has. But as far as I've heard, thrust and fuel flow are roughly proportional in each given regime, and flight engineers used flow gauges to match thrust of individual engines in flight?

2. There is no way to make variable TSFC, as `<tsfc>` tag is just a number and is not exposed anywhere in property tree -- so if you have a table of TSFC per flight regime, there is no way to put it to use, and you will have wrong range for anything but cruise.

Kind regards,
Mike

bcoconni:
> `fuel-flow-rate-pps` considerably lags behind the change of thrust-lbs -- it takes almost half a minute to settle after thrust has. But as far as I've heard, thrust and fuel flow are roughly proportional in each given regime, and flight engineers used flow gauges to match thrust of individual engines in flight?

`FGTurbine` uses millions of magic numbers and the latency for the fuel flow is determined by some of them.
https://github.com/JSBSim-Team/jsbsim/blob/c0b8e62113492570f211f6510fd075eb65396319/src/models/propulsion/FGTurbine.cpp#L190
https://github.com/JSBSim-Team/jsbsim/blob/c0b8e62113492570f211f6510fd075eb65396319/src/models/propulsion/FGTurbine.cpp#L239
https://github.com/JSBSim-Team/jsbsim/blob/c0b8e62113492570f211f6510fd075eb65396319/src/models/propulsion/FGTurbine.cpp#L253
https://github.com/JSBSim-Team/jsbsim/blob/c0b8e62113492570f211f6510fd075eb65396319/src/models/propulsion/FGTurbine.cpp#L262
These numbers need to be tweaked to obtain the result you are seeking.
> There is no way to make variable TSFC, as `<tsfc>` tag is just a number and is not exposed anywhere in property tree -- so if you have a table of TSFC per flight regime, there is no way to put it to use, and you will have wrong range for anything but cruise.

It is feasible to make the TSFC available as a property however `FGTurbine` uses a corrected version of TSFC that depends on the ambient temperature, `N2` and some more magic numbers.
https://github.com/JSBSim-Team/jsbsim/blob/c0b8e62113492570f211f6510fd075eb65396319/src/models/propulsion/FGTurbine.cpp#L238
[...]

Issue #195, opened 2019-07-07
Preposterous Windmilling Behavior


Octal450:
![image](https://user-images.githubusercontent.com/25598604/60763306-89651e00-a03f-11e9-893c-ac6ea76d9e6b.png)
^ IDG A32X
Hi guys -- even at a high speed, this makes no sense. At low speeds too, the windmilling is too high. This can be see here:
![image](https://user-images.githubusercontent.com/25598604/60763314-a13ca200-a03f-11e9-82c0-2608e23e96c3.png)
^ IDG MD-11

I will get the values of what it should be about windmilling at. N2 seems plausible, but N1 is just nonsensical.
Any thoughts?

Kind Regards,
Josh

bcoconni:
This may be unrelated but as I mentioned in the issue #178
> `N1` and `N2` are hard coded to be respectively equal to `qbar/10` and `qbar/15` when the engine is off. The mass airflow would more relevant than the dynamic pressure for that matter.

During windmilling it can be assumed that the airflow is proportional to the ram airflow which is proportional to `in.Density` (air density) times the aircraft velocity forward velocity `in.aeroUVW(eU)`. The file `src/models/propulsion/FGTurbine.cpp` could be modified as shown below:

```diff
@@ -190,8 +190,9 @@ double FGTurbine::Off(void)
   FuelFlow_pph = Seek(&FuelFlow_pph, 0, 1000.0, 10000.0);
   // some engines have inlets that close when they are off. So, if a flag is true disable windmilling
   if (disableWindmill == false) {
-    N1 = Seek(&N1, in.qbar/10.0, N1/2.0, N1/N1_spindown);
-    N2 = Seek(&N2, in.qbar/15.0, N2/2.0, N2/N2_spindown);
+    double AirFlow = in.Density * in.AeroUVW(eU);
+    N1 = Seek(&N1, AirFlow/10.0, N1/2.0, N1/N1_spindown);
+    N2 = Seek(&N2, AirFlow/15.0, N2/2.0, N2/N2_spindown);
   } else {
     N1 = Seek(&N1, 0, N1/2.0, N1/N1_spindown);
     N2 = Seek(&N2, 0, N2/2.0, N2/N2_spindown);
```
and the constants `10.0` and `15.0` might need to be updated to reach the expected rotation speeds `N1` and `N2`.

If you can compile FlightGear could you please check the above and let me know if that fixes the problem ?

Octal450:
[...]

PR #577, opened 2022-01-23
Corrected deceleration formula

pbecchi:
I was not satisfied from deceleration formula on FGBrushLessDCMotor.cpp.
I spend all weekend testing with different deceleration factors and integration delta time. 
I ended up simplifying and modifing the formula so that results do not change changing integration delta time.
I think results are much better now :
-best integration interval is smaller than 0.004 seconds
-best deceleration_factor is 5 , it may be adjusted changing model size and inertias.


bcoconni:
I must admit that I am not convinced at all that the changes that you are submitting in this PR are relevant to the code of `FGBrushlessDCMotor`. The purpose of JSBSim engine models is to simulate (or at least emulate) the physics of the engine and I am really struggling to see why BLDC motors should physically have some hysteresis in their accelerations (i.e. why it should decelerate faster - or slower - than it is accelerating). Especially with a [physical model which is basically linear](http://web.mit.edu/drela/Public/web/qprop/motor1_theory.pdf). This `DecelerationFactor` looks very much like a fudge factor to me (kind of a [Deux ex machina](https://en.wikipedia.org/wiki/Deus_ex_machina) if you will).

While I was reading @seanmcleod comments to your other PR #574 about the relationship between `Kq` and `Kv`, it occurred to me that commercially available drones such as DJI's use very sophisticated software for their GNC. Hence it is almost certain that there is a decent layer of software between the user inputs and the voltage at the quadcopter motors. For instance, it is perfectly plausible that the DJI control software applies a negative voltage to the motors to decelerate abruptly the propellers as soon as the user releases the throttle. Conversely their control software can maintain a certain amount of voltage to the motors after the user has released the throttle in order to achieve a smooth deceleration. Said otherwise, you can't assume that there is a linear relationship between the user throttle input and the voltage at the motors.

As a consequence, the complex behaviours that are observed on the real drone should typically be coded in the control software of your DJI450 rather than in the class `FGBrushLessDCMotor` where it would be hardcoded.

pbecchi:
Let my breefly answer your concerns.
For what I know about multicopters , I  can tell you this: propulsion is always based on this chain of components:
**propeller-brushless motor-ESC (speed controller)-battery**. 
All these are standard components and  each has his own function.
There is no direct control of Voltage or motor RPM from any flight controllers I know. The FCS simply send a throttle command to the ESC that has the function of modulating the voltage and supply a 3 phase alternate current to brushless motor.
The code of FGBrushLessDCMotor emulate or simulate if you prefer the functions of Brushless Motor + ESC.
There maxvolts represents the input voltage  to the ESC , this voltage is the output from battery. The 3 constant equations represents the stationary electromechanical  behaviour of the motor  . 
Transient accelerations and decelerations are fully controlled by the ESC within the fisical performances available from the motor : max positive and negative torque (where acceleration torque >>deceleration  torque).
The way i have computed acceleration is simple: it  is is based on  max torque available.
Braking is a different story:  maximum torque can only be reversed if you assume that ESC will generate an inverse current to charge batteries or to be dissipated  in an external resistor.
This is not normally done from any multirotor ESC. Most ESC use an opensource control software named BLHeli32 . This SW give the possibility to the user to change the braking behaviour of the motor .This deceleration factor is not controlled by the FCS but is fixed and chosen when SW is uploaded in the ESC.  Therefore it seams correct to me to leave the possibility to adjust  braking torque. The best way i have found is to factor the inertia torque when  braking is more or less powerfull.
Than i propose to leave the possibility to adjust braking torque and to use the proposed set of equations.

It will be ideal to have the possibility to test some battery, ESC, Motor and propeller  to mach behaviour of the FGBrushLessDCMotor.  But the test setup is complex and this is not feasible in the short time . If you still are doubtfull  ask some multicopter, UAV or FCS expert.
This is for me a good ( approximate) emulation of the behaviour of a multicopter brushless motor and speed controller.

pbecchi:
This is a screenshot of the BHeli32 ESC SW:
image

"non damped mode" is parameter used to change braking force .
 "brake on stop" is only to brake propeller at 0 = RPM ,only used for airplanes.

bcoconni:
> The code of FGBrushLessDCMotor emulate or simulate if you prefer the functions of Brushless Motor + ESC.

I'll address later the other topics you mentioned but I'd like to react to this statement because I think this is where the misunderstanding is coming from. The code architecture of JSBSim is designed to separate the physical simulation of the engines (basically the C++ code located in `src/models/propulsion`) from the control software (i.e. the XML files in `systems`).

There is a good reason for that architecture: the physics of an engine are very unlikely to change but the same cannot be told for the FCS or the ECS. As you mentioned these are pieces of software that can be easily modified/tuned and as popular as BLHeli32 might be, some other users might select another software, including a customized version of BLHeli32.

If the logic of BLHeli32 is hardcoded in the C++ code of `FGBrushlessDCMotor` then other users have no longer the option of using/coding another (maybe obscure/less popular) ESC than BLHeli32. Likewise students might want to code their own ESC for the only purpose of learning how an ESC is coded. There again, if an ESC logic is already hardcoded in `FGBrushlessDCMotor` they would no longer have this option. You might have noticed that there are no hardcoded C++ autopilots in JSBSim and that is for the very same reason than I mentioned above: users might want to write their own autopilot either for their specific usage of JSBSim or for educational purpose. The current architecture of JSBSim allows much freedom regarding FCS programming or, for that matter, ECS programming.

So due to the design of JSBSim, the ESC *must not* be located in the class `FGBrushlessDCMotor`. Instead  `FGBrushlessDCMotor` should contain the equations from the linear model of an electric motor, no more no less. Everything that pertains to the ESC must be removed from  `FGBrushlessDCMotor` and coded with the XML control features in the same way than FCS are currently coded.

pbecchi:
Sorry but i çant follow what you are saying.
The ESC is a brushless motor controller no bldc motor exist without, it is like a carburator for a reciprocating engine, it is like the brushes on a brushed motor.
It is not in any way something that can be considered part of the FCS or that can be modeled like a flight controller.
They normally do a very high frequency switching in order to sinusoidaly modulate voltage on motor windings.
 How this can be emulated with an xml file?
What is the output that will drive the motor?
The function of a BLdc motor how can be represented without an Esc? 
I think that what you are asking it is not feasible (at least for me).
Sorry

pbecchi:
Let me try to explain myself better....

First of all a little background:
I am not an expert on electronics or electric motors. I am an old aeronautical engineer that struggle to design  a new unmanned eVTOL .
I had to learn about multicopters , BLDC motors ,ESC, batteries , all things  were not existing at the the time I complete my studies.
The basic knowledge i have put in the code is coming from recent info i got from a company that will provide electric motors for the UAV. So I am not an expert but the equations in the code are the one I use for evaluation of performances of the aircraft.

Now , talking about the ESC..... It is to be considered , as i said yesterday, a part of the motor itself.  When we represent physical behaviour  we need to define blocks that can be aproximated by formulas . 
In stationary condition the ESC linearly convert throttle to alternate voltage to the motor, there is negligeable lag and there is an internal resistance  that need to be added to the one of the coils.
In dynamic conditions during the transitions, what it generate the lag is not the ESC itself but the complex electric and magnetic hysteresis  phenomena inside the windings of the motor. To represent these effect is very complex and , i think, out of scope.
It is a very   complex behaviuor  with frequencies of 1/revolution multiplied by number of poles  , very difficult to study and represent with control blocks. I am not saying it is impossible , but it is something I dont realy know how to do.
  Sorry, again.
Anyhow from my first tests I can say that the FGBrushlessDCMotor represent well enought a motor to allow simulations of multicopters and eVTOLs. I will keep working with it. 
I am allways available , if you wish, to complete this work.... more need to be done !!
For example you will need a module that represent the battery : his output will be the maxvolts input to the motors and it will have a function  similar to the tank for a thermic motor.

bcoconni:
@pbecchi My previous comments may have been misleading: I am not trying to dismiss the relevance of your work by any means. Your contributions are very much appreciated and they are adding a feature that the JSBSim community is long-awaiting as is illustrated by the discussions in issue #333. Thank you for stepping up and addressing this topic :+1:.

Regarding this PR, I am not saying that the ESC is not needing any tuning nor that the modifications you have made are irrelevant. As I tried to explain, initially I was under the assumption that `FGBrushlessDCMotor` was containing nothing more than the [First-Order DC Electric Motor Model from Mark Drela](http://web.mit.edu/drela/Public/web/qprop/motor1_theory.pdf). As I could not make a connection between Drela's model and the logic based on `DecelerationFactor`, I was challenging your PR. Now that I understand where the confusion is coming from, I am suggesting a different design for your code. I can provide you with some guidance to address the changes I am suggesting, the modifications are most likely not as significant as you may currently be anticipating.

And there are some other less technical but still very important factors that must be accounted for. Open Source projects are understaffed and JSBSim is no exception. A fact of life is that most contributions to the project are "one shot" and the vast majority of the contributors are no longer being heard of after their contribution has been integrated. I am not complaining; this statement is just the result of my experience and unfortunately JSBSim is no exception in that regard among the realm of Open Source projects. The implied consequence is that the project maintenance and support is made by a very few number of people (@seanmcleod and myself at the moment). For that purpose, we need to have a reasonable understanding of the changes that are submitted by occasional contributors since, at the end of the day and in all likelihood, it is @seanmcleod and myself that will have to answer questions from users and fix bugs; even though we are no experts in BLDC motors (which is the very reason why @seanmcleod asked you some links to some documentation for your model). From that perspective, contributions must follow some design principles that are the result of our experience regarding users questions and requests.

My reluctance to hardcoded controls is coming from the class `FGTurbine` which in my opinion is the perfect example of a badly designed engine model. Almost everything regarding the engine behaviour has been hardcoded in `FGTurbine` and we are now addressing a never ending list of requests to allow for more versatility in the `FGTurbine` (issues #178, #195, #196, #233, #494, etc.)

Hopefully you now understand my motivations 😄 and I will address the technical side of our discussion later today.

bcoconni:
### Current code design
The current code in `FGBrushlessDCMotor` is a feedback control loop that drives the propeller rotation speed `RPM` from the motor input voltage. It is executed once every time step and it is (sort of) using the time derivative of `RPM` to assess the increase/decrease in torque `InertialTorque`:
https://github.com/JSBSim-Team/jsbsim/blob/03842c7742994120f6cdd7a5adcc3cd94b4b55b9/src/models/propulsion/FGBrushLessDCMotor.cpp#L163

Said otherwise, this is a PID loop with P=0, I=0 and D=1/Ixx. By the way the D constant is a first order contributor to the speed at which the target `RPM` will be reached, yet it is hardcoded in the current code ?

As a side note and strictly speaking, `DeltaRPM` is not the time derivative of `RPM`. Instead it is the acceleration needed to reach the target RPM in one time step. But this distinction does make much difference for the matter in discussion.

So the input of this feedback loop is the motor input voltage
https://github.com/JSBSim-Team/jsbsim/blob/03842c7742994120f6cdd7a5adcc3cd94b4b55b9/src/models/propulsion/FGBrushLessDCMotor.cpp#L152
and the target is the propeller rotation speed which is proportional to the throttle:
`TargetRPM = VelocityConstant * MaxVolts * in.Throttle[EngineNumber]`.

### New design proposal
The PID that is currently in the C++ code can be replaced by an XML function in the control system (this would be the ECS):
```xml
<!-- compute the difference between the target RPM and the current RPM -->
<function name="propulsion/engine[#]/delta-rpm">
  <difference>
    <product>
      <p> fcs/throttle-cmd-norm[#] </p>
      <p> propulsion/engine[#]/max-volts </p>
      <p> propulsion/engine[#]/Kv-rpm_volts </p>
    </product>
    <p> propulsion/engine[#]/engine-rpm </p>
  </product>
</function>

<!-- feedback loop to drive the engine RPM to the target -->
<pid>
  <input> propulsion/engine[#]/delta-rpm </input>
  <output> fcs/throttle-pos-norm[#] </output>
  <kd> 0.1 </kd> <!-- Or whatever value that fits the need -->
</pid>
```
Note that the target RPM is based on the throttle **command** `fcs/throttle-cmd-norm` and the input voltage is driven by the throttle **position** `fcs/throttle-pos-norm`. For that function to work, we need to expose the properties `propulsion/engine[#]/max-volts` and `propulsion/engine[#]/Kv-rpm_volts` but that's trivial.

The code in `FGBrushlessDCMotor::Calculate` would then be much simpler:
```c++
void FGBrushLessDCMotor::Calculate(void)
{
  RunPreFunctions();

  if (Thruster->GetType() == FGThruster::ttPropeller) {
    ((FGPropeller*)Thruster)->SetAdvance(in.PropAdvance[EngineNumber]);
    ((FGPropeller*)Thruster)->SetFeather(in.PropFeather[EngineNumber]);
  }

  double RPM = Thruster->GetRPM();
  double V = MaxVolts * in.ThrottlePos[EngineNumber];
  double Current = (V - RPM/VelocityConstant)/CoilResistance; // Equation (4) from Drela's document
  double TargetTorque = (Current - NoLoadCurrent)/VelocityConstant; // Assuming Kv == Kq

  EnginePower = ((2 * M_PI) * RPM * TargetTorque) / 60;   //units [#*ft/s]
  HP = EnginePower /hptowatts*NMtoftpound;   // units[HP]
  LoadThrusterInputs();
  Thruster->Calculate(EnginePower);

  RunPostFunctions();
}
```

seanmcleod:
@pbecchi I'd like to echo what @bcoconni mentioned above, i.e. thanks for the pull requests and taking this on, as he mentioned I opened the original issue/discussion quite a while ago.

But I'd also like to echo what he said about us and me in particular not being experts in terms of BLDC motors, hence my reason for asking for some sort of reference, and the reason my initial comments regarding your pull request only really dealt with coding style issues etc., because I didn't have a 'textbook' reference model to compare your code against.

In your subsequent commit you did add a reference to the motor model described by Mark Drela and hence the reason I started to ask questions, e.g. Kv versus Kq etc.

So ideally the code would be fairly easy to match up with the equations in the 'textbook', similar to textbook references we have in other parts of the code. Making it much easier for someone, whether that's us or others in the future when digging into the BLDC code without trying to figure out 'magic' numbers etc. that don't appear to be referenced in the textbook explanation etc.

It looks like @bcoconni has helped a lot in that regards with his latest comment above. I'll take a more detailed look tomorrow.

pbecchi:
I dont know..... sorry bue I know nothing about transfer functions and control loops.
What i can tell you is that the motor equations that tie RPM, Torque and Volts are ok. They are linear but propeller curve are not  .
I am not sure  if you can split RPM from deltaRPM the way you are propose : for me deltaRPM is not only function of commandedVolts but also of Torque (or RPM).  Without knowing current torque , you cannot derive RPM only from Voltage. 
I believe you have to do something more complex that the simple product of throttle , volts and Kv....... You have to use the same equations you are putting in the c code, this way it will work.
But then why to split things this way? 

The differential equation that represent motor+propeller should be:

**`RPMdot  * Inertia + f( RPM^2)=f(RPM, Volts)`**
where` f(RPM^2)`  is the propeller torque
and   ` f(RPM,Volts)`  is the torque generated by the motor.

I really dont know if it can be solved  the way you propose........
Sorry
We have different backgrounds , therefore we see problem solutions in different way.

pbecchi:
I have found this : https://ctms.engin.umich.edu/CTMS/index.php?example=MotorSpeed&section=SystemModeling

Could this be the best set of equation to dynamical model a brushlessDC motor?

Could all this be represented as a xml control block? In this case, the `FGBrushlessDCMotor.cpp` is still necessary?

The only doubt is how to get additional  input necessary like the inductance that is very hard to get form motor datasheet.
Regarding damping coeficient , I think it will be negligeable compared to propeller torque.

bcoconni:
> I am not sure if you can split RPM from deltaRPM the way you are propose : for me deltaRPM is not only function of commandedVolts but also of Torque (or RPM). Without knowing current torque , you cannot derive RPM only from Voltage.
I believe you have to do something more complex that the simple product of throttle , volts and Kv....... You have to use the same equations you are putting in the c code, this way it will work.

I agree: my computation of `DeltaRPM` is incorrect. Indeed I forgot to subtract the voltage generated by the resistance but this can easily be fixed in the XML control function.

> But then why to split things this way?

As I explained previously, the ECS can easily be modified once it is coded in an XML control file: most users do not wish to compile JSBSim themselves.

> The differential equation that represent motor+propeller should be:
>
> **`RPMdot  * Inertia + f( RPM^2)=f(RPM, Volts)`**
> where` f(RPM^2)`  is the propeller torque
> and   ` f(RPM,Volts)`  is the torque generated by the motor.
>
> I really dont know if it can be solved  the way you propose........

The differential equation that you are mentioning is actually solved in `src/models/propulsion/FGPropeller.cpp`:
https://github.com/JSBSim-Team/jsbsim/blob/03842c7742994120f6cdd7a5adcc3cd94b4b55b9/src/models/propulsion/FGPropeller.cpp#L289

so it is not being removed. And to be clear, the code in `FGBrushlessDCMotor` that computes `InertialTorque` does not solve a differential equation; it computes a derivative (it is using the inverse of the time step):
https://github.com/JSBSim-Team/jsbsim/blob/03842c7742994120f6cdd7a5adcc3cd94b4b55b9/src/models/propulsion/FGBrushLessDCMotor.cpp#L163

bcoconni:
> I have found this : https://ctms.engin.umich.edu/CTMS/index.php?example=MotorSpeed&section=SystemModeling
> 
> Could this be the best set of equation to dynamical model a brushlessDC motor?

We might consider later improving the code in `FGBrushlessDCMotor` with a more complex and higher fidelity model but for now I'd suggest we keep focused on Drela's first order model.

pbecchi:
I have modified  FGBrushlessDCMotor as you suggested. Without any external feedback it works.
This show test results:
image
If you agree I will correct this way code in the PR.

bcoconni:
> I have modified FGBrushlessDCMotor as you suggested. Without any external feedback it works. [...] If you agree I will correct this way code in the PR.

Yes, by all means. Please do.

pbecchi:
New commit: "Simplify&RemoveDeltaRPM uploaded.

pbecchi:
I am considering to add effect of inductance: it will reduce the sharp torque pick and is physically more correct .....

I do it here or in a new PR??

bcoconni:
> I am considering to add effect of inductance: it will reduce the sharp torque pick and is physically more correct .....
> 
> I do it here or in a new PR??

A pull request should focus on one goal only. So yes, please address the inductance feature in a separate PR.

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -142,42 +145,27 @@ void FGBrushLessDCMotor::Calculate(void)
...
-  HP = EnginePower /hptowatts*NMtoftpound;                             // units[HP]
+  double Current = (V - RPM / VelocityConstant) / CoilResistance;        // Equation (4) from Drela's document
+
+  if (Current >= 0)
+    TargetTorque = (Current - NoLoadCurrent) / VelocityConstant * TorqueConstant;    // Assuming Kv == Kq
+  else
+    TargetTorque = (Current + NoLoadCurrent) / VelocityConstant * TorqueConstant; // Assuming Kv == Kq
+
+
+  EnginePower = ((2 * M_PI) * max(RPM, 0.0001) * TargetTorque) / 60;              //units [#*ft/s]
```
bcoconni, line 162:
```suggestion
  EnginePower = ((2 * M_PI) * RPM * TargetTorque) / 60;              //units [#*ft/s]
```
I still don't catch why `EnginePower` should not be zero ?

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -142,42 +145,27 @@ void FGBrushLessDCMotor::Calculate(void)
   }
 
   RPM = Thruster->GetRPM();
-  TorqueRequired = abs(((FGPropeller*)Thruster)->GetTorque()); //units [#*ft]
 
-  CurrentRequired = (TorqueRequired * VelocityConstant) / TorqueConstant;
+  TorqueRequired = abs(((FGPropeller*)Thruster)->GetTorque());           //units [#*ft]
+
```
bcoconni, line 150:
```suggestion
```
Unless I am mistaken `TorqueRequired` is no longer used.

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -142,42 +145,27 @@ void FGBrushLessDCMotor::Calculate(void)
...
-  }
 
-  EnginePower = ((2 * M_PI) * RPM * TargetTorque) / 60;   //units [#*ft/s]
-  HP = EnginePower /hptowatts*NMtoftpound;                             // units[HP]
+  double Current = (V - RPM / VelocityConstant) / CoilResistance;        // Equation (4) from Drela's document
+
+  if (Current >= 0)
+    TargetTorque = (Current - NoLoadCurrent) / VelocityConstant * TorqueConstant;    // Assuming Kv == Kq
+  else
+    TargetTorque = (Current + NoLoadCurrent) / VelocityConstant * TorqueConstant; // Assuming Kv == Kq
```
bcoconni, line 159: I don't see why the sign of `NoLoadCurrent` should be kept the opposite of `Current` ? Its definition is not influenced by the sign of `i` i.e. of `Current`.
Capture d’écran du 2022-01-30 14-13-09

bcoconni (changes requested):
A number of minor details needs to be fixed.

However I'm struggling about the reason behind the condition:
```c++
  if (Current >= 0)
    TargetTorque = (Current - NoLoadCurrent) / VelocityConstant * TorqueConstant;    // Assuming Kv == Kq
  else
    TargetTorque = (Current + NoLoadCurrent) / VelocityConstant * TorqueConstant; // Assuming Kv == Kq
```
This means the `NoLoadCurrent` should always have an opposite sign to `Current`. However according to the definition of `NoLoadCurrent` (aka `i0` in Drela's document), I don't see why this should be the case ?

src/models/propulsion/FGBrushLessDCMotor.cpp
pbecchi: This is necessary otherwise EnginePower will always be 0, since at startup RPM=0.

src/models/propulsion/FGBrushLessDCMotor.cpp
pbecchi: No load current is by definition the current necessary for the motor to overcome internal friction : therefore it is always against applied torque  and consequently to current.

pbecchi: Anyhow it should be 
`TargetTorque=0;
if(Current>=NoLoadCurrent) =(Current-NoloadCurrent)/VelocityConstant * TorqueConstant;
if(Current<-NoLoadCurrent) =(Current+NoloadCurrent)/VelocityConstant * TorqueConstant;

`
It is correct that when abs(current)<noloadcurrent then torque=0.

pbecchi: Or if you prefer I can leave it without checking current direction , the effect is negligeable.......

bcoconni: > No load current is by definition the current necessary for the motor to overcome internal friction : therefore it is always against applied torque and consequently to current.

OK. It makes sense but it certainly deserve a comment to explain just what you wrote above.

src/models/propulsion/FGBrushLessDCMotor.cpp
```diff
@@ -142,42 +142,25 @@ void FGBrushLessDCMotor::Calculate(void)
...
```
bcoconni, line 154:
```suggestion
  CurrentRequired = (V - RPM / VelocityConstant) / CoilResistance;        // Equation (4) from Drela's document

  if (CurrentRequired >= 0)
    TargetTorque = (CurrentRequired - NoLoadCurrent) / VelocityConstant * TorqueConstant;    // Assuming Kv == Kq
  else
    TargetTorque = (CurrentRequired + NoLoadCurrent) / VelocityConstant * TorqueConstant; // Assuming Kv == Kq
```
It occurred to me that we should use the member `CurrentRequired` here otherwise the property `propulsion/engine[#]/current-amperes` will no longer be updated.

bcoconni (changes requested):
The code should use the member `CurrentRequired` instead of the local variable `Current` otherwise the property `propulsion/engine[#]/current-amperes` won't be updated. I introduced this bug while suggesting an alternative way of implementing the code, sorry.

bcoconni:
By the way the members `DeltaRPM`, `MaxTorque`, `TorqueAvailable`, `TorqueRequired`, `InertialTorque` and `DecelerationFactor` are no longer needed in `src/models/propulsion/FGBrushlessDCMotor.h`.

src/models/propulsion/FGBrushLessDCMotor.cpp
pbecchi: Ok I will correct the way i have proposed above...

pbecchi: Resolved with last 2 commits

src/models/propulsion/FGBrushLessDCMotor.cpp
pbecchi: resolved with last 2 commits

seanmcleod:
@pbecchi it looks like your latest commit https://github.com/JSBSim-Team/jsbsim/pull/577/commits/d345d71d24161a0c31e6651eb17b4413006be79f has gone an added a bunch of irrelevant files.

seanmcleod:
`TorqueConstant`, since we're specifically referencing Drela's motor model document let's not use `TorqueConstant` unless it refers to `Kq`. What you currently have is a unit conversion constant.

In terms of consistency with Drela's motor model document I suggest we replace `VelocityConstant` with `SpeedConstant` for `Kv`. 

Why `Required` in the variable `CurrentRequired`? Surely a better description is simply `Current`, since it's simply the current given by equation 4, given the motor parameters `Kv` and `R` and then the current value of `RPM` and the currently supplied `V`. Required implies someone/some component is requesting or needs the current to be this value.

Similar question with regards to `TargetTorque`, surely it's simply the torque that the motor generates again given the motor constants and the current voltage supplied etc.?

Now in terms of this current going negative, i.e. when the supplied voltage `V` is less than the back-EMF, i.e. `RPM/Kv` should the torque not go negative? In other words a braking torque and current is used for regenerative charging?

bcoconni:
@seanmcleod I have merged this PR, I'll include your comments in a cleanup commit that I'll push shortly.

pbecchi:
Yes when you decelate torque become negative , you can see it from the plot few messages above...
What concern me is the high peak in torque at the start of an input step, to fix this we need to add Inductance.

Sorry for variable names....we can always improve ....

bcoconni:
@seanmcleod, I tried to include as many of your comments as I could in commit 60041d7ee2e30ee961bd9f491e8a92be24e7de60. I guess we still need to rename `<velocityconstant>` to `<Kv>` and `<noloadcurrent>` to `<zerotorquecurrent>` to match Drela's paper but I am waiting for PR #574 to merged first.

seanmcleod:
> Yes when you decelate torque become negative , you can see it from the plot

@pbecchi ah, while glancing at the diff at the time I didn't notice that you were implementing a deadband, so I was a bit confused about what you were trying to do with that piece of code.

pbecchi:
This is the correct way to code the noloadcurrent formula.
Otherwise without the `if` at 0 amperes the torque will be negative and the motor will turn backward!!!!

seanmcleod:
> Otherwise without the if at 0 amperes the torque will be negative...

But it does go negative in the graph at time 30 right? Basically when the back-emf is large enough as the input voltage is dropped suddenly resulting in current < -i0 (no load current).

pbecchi:
Yes, correct! With simplified model braking generate current .

pbecchi:
Did you test the example of PR 574?

PR #617, opened 2022-04-29
Added JSBSim for Unreal Engine 5, and UE Reference Application

AlbanBERGERET-Epic:


bcoconni:
Hi @AlbanBERGERET-Epic,

Thank you very much for your contribution. This is a remarquable new application for JSBSim.

I quickly reviewed your PR and I have a few questions to ask:

Could you please remove the file `systems/fg-fglue.xml` from your PR as it fails the XML validation test (see https://github.com/JSBSim-Team/jsbsim/runs/6225004337?check_suite_focus=true#step:5:25) ?

Also is there a way not to include the multimegabytes "assets" files in the PR as it will significantly increase the size of our repo ?

Thanks for your help.

AlbanBERGERET-Epic:
Hi @bcoconni 

I reduced the assets file and provided a download link to get access to HD Content optionally. 
The total size of the UnrealEngine folder is under 50MB (47MB)

And the bas XML file has gone. 

Thanks!!

seanmcleod:
@AlbanBERGERET-Epic this link the readme doesn't appear to be valid - https://dev.epicgames.com/community/learning/tutorials/mmL/a-diy-flight-simulator-tutorial

ICanDoThings2:
Is this working with 5.01? Or only for 5.1? How would I compile this, I see no solution file in the example project folder for UE5.

HeadClot:
> Is this working with 5.01? Or only for 5.1? How would I compile this, I see no solution file in the example project folder for UE5.

Hey @ICanDoThings2 - You need to generate the solution file by right clicking the .uproject file and click "Generate Visual Studio Project files". It should also work with Unreal 5.01.

ICanDoThings2:
> > Is this working with 5.01? Or only for 5.1? How would I compile this, I see no solution file in the example project folder for UE5.
> 
> Hey @ICanDoThings2 - You need to generate the solution file by right clicking the .uproject file and click "Generate Visual Studio Project files". It should also work with Unreal 5.01.

I found it now I think, thanks and sorry! I think I see now. I thought LWC wasn't supported in UE5 yet though.

AlbanBERGERET-Epic:
@ICanDoThings2, yes, it works for UE 5.0.1
You'll find in the Unreal Engine folder a md file[](https://github.com/JSBSim-Team/jsbsim/blob/master/UnrealEngine/JSBSimForUnreal.md)  with all instructions for the build. 

UE will indeed generate the sln file for you. Have a look at the tutorial (link in the md) to learn more. LWC is here in 5.0 but should should enable large world in the worldsettings to enable it entirely.

ICanDoThings2:
I followed the instructions from the JSBSimForUnreal.md file, but it is telling me includes are missing on Option 2 step. Specifically;  cannot open include file FGFDMExec.h
image

Should I reinstall the .JSB project or something?

ICanDoThings2:
After we compile the project with the plugin in UE5, can we move the plugin to other projects and outside the folder it's compiled in? Or must it stay there and/or paired to the example project forever?

AlbanBERGERET-Epic:
@ICanDoThings2 

Sorry for the answer delay, but we are live on wats2022 event, with lots of attendance on our booth, so I can't be as present as needed...

Did you build the jsbsimforunreal.sln file at the root folder of jsbsim first? This solution will build jsbsim as static libs and copy all the needed includes and resources in the plugin folder. This is needed by UE plugin system. 
So make sure you have the includes copied during the post-build event. They should be located in the Unreal engine/plugins/jsbsimflightdynamicmodel/source/3rdparty/jsbsim (or something like that, from the top of my head).

Once you build the plugin for the first time (and actually, even once you build the root jsbsimforunreal.sln for the first time), you can just copy/paste the whole plugin folder to your own application. This is even recommended to do so because the reference application is here as an example, and as a way to do non-regression tests.

AlbanBERGERET-Epic:
Please also make sure that there is never any space in your folder hierarchy.

ICanDoThings2:
> @ICanDoThings2
> 
> Sorry for the answer delay, but we are live on wats2022 event, with lots of attendance on our booth, so I can't be as present as needed...
> 
> Did you build the jsbsimforunreal.sln file at the root folder of jsbsim first? This solution will build jsbsim as static libs and copy all the needed includes and resources in the plugin folder. This is needed by UE plugin system. So make sure you have the includes copied during the post-build event. They should be located in the Unreal engine/plugins/jsbsimflightdynamicmodel/source/3rdparty/jsbsim (or something like that, from the top of my head).
> 
> Once you build the plugin for the first time (and actually, even once you build the root jsbsimforunreal.sln for the first time), you can just copy/paste the whole plugin folder to your own application. This is even recommended to do so because the reference application is here as an example, and as a way to do non-regression tests.

I got it now, so sorry about that. Is there any sort of documentation for implementing new planes with the plugin and demo? And again, when we compile the example project, can we take the plugin out and put it in other projects, or will it only work with the example?

ICanDoThings2:
Do you think for a future release, we might have support for creating and editing models in the editor instead of having to write them strictly in an .xml file? It would make tweaking models much easier in my opinion.

ICanDoThings2:
Also, maybe a way to toggle flight dynamics calculations on/off for when an aircraft is landed too or something so it's not doing math for nothing, or in cases when we want to animate it kinematic wise temporarily I think would be invaluable.

thorcx:
@AlbanBERGERET-Epic Hi, how do i report bugs about the JSBSim for unreal engine5 plugin?

PR #632, opened 2022-05-28
Metric output

ermarch:
AeromatiC++:
* Overhaul of the propeller code, the produced tables now matches published data quite close.
* Create a configuration file in metric units when the parameters are entered in metric units.
* Add the option to save the aerodynamics section and the external-force section into separate files.


utils/aeromatic++/Systems/Controls.cpp
```diff
@@ -152,75 +152,82 @@ void CableControls::set(const float* cg_loc)
...
+    }
+
+    // Fuselage
+    float Df = _aircraft->get_fuselage_diameter();
+    float Lf = _aircraft->_length;
+    float fr = Lf/Df; // fuselage finess ratio
+
+    float Qf = 1.0f;
+    float FFf = 1.0f + 60.0f/powf(fr, 3.0f) + fr/400.0f;
+    float Swet_f = PI*Df*Lf*pow(1.0-2.0/fr, 0.667f)*(1.0 + 1.0/(fr*fr));
```
bcoconni, line 195: GitHub CodeQL raises an error on that line because `PI` is (I think) a `double` while the rest of the expression is made of `float` variable. Would you mind convert `PI` to a float in this expression ?
```suggestion
    float Swet_f = ((float)PI)*Df*Lf*pow(1.0-2.0/fr, 0.667f)*(1.0 + 1.0/(fr*fr));
```

bcoconni (changes requested):
Please fix the line 195 of `Systems/Control.cpp` with regard to the mix of `double` and `float` in the same expression.

utils/aeromatic++/Systems/Controls.cpp
ermarch: Odd, PI is defined float explicitly:
types.h:#define PI			3.14159265358979323846f

Maybe it's too exact for a float and turned to a double automatically?

bcoconni: May be or may be it's the `pow()` function that needs to be changed to `powf()` to return a `float` as well as the constants that should be appended with a `f` ?

ermarch: Shoot, I noticed forgot about the pow function. Hold on.

bcoconni (approved):

bcoconni:
Does this fix the issue #173 ?

ermarch:
No, this does not fix issue #173 That would require to create an incompatible configuration file format.
But you could easily edit the configuration file to update the number of blades in the generated file itself.

PR #650, opened 2022-06-08
Using JSBSim as a DLL to allow for LGPL License + Plugin License

AlbanBERGERET-Epic:
The JSBSim LGPL license can't be used in an Unreal-based application unless we use a DLL instead of a static link. 
This PR resolves this issue by using JSBSim as a DLL

We also added an MIT License to the plugin-specific code. It's free to use and to edit for any kind of use.


AlbanBERGERET-Epic:
@bcoconni 

I have checked with Epic Legals. 

- It's totally fine to use JSBSim in any commercial application as long as we link dynamically, so we use the DLL build. 
- We don't need our contributions to be subject to the LGPL license. (I don't know if there are github options that would exclude my PRs from the LGPL implicit terms). 
- We provide the UE Reference App and the JSBSimFlightDynamic plugin as different packages, each of them as MIT license, so free to use and modify in any context. I created two separate license files for that very purpose. 

Last note : It should be useful to add a dedicated text in the main README.md in the legal notice to restate that the Unreal Engine connection is not using the LGPL, but the MIT.

bcoconni:
> > Last note : It should be useful to add a dedicated text in the main README.md in the legal notice to restate that the Unreal Engine connection is not using the LGPL, but the MIT.
> 
> Good point. I will updated the "general" README.

This is now done following the commit c18873e2f072282855e56e2485d02f39684f5475.

PR #651, opened 2022-06-08
Providing more flexible functionality to the MATLAB S-Function library

noah-curran:
Hi all,
A few weeks ago I dove into the MATLAB S-Function for JSBSim in order to improve it as an experimental tool.
For those who didn't follow, there was a discussion at #627. In addition to some changes I planned to make, I also incorporated some changes based on the feedback from the conversations listed there.

To summarize what is being added here:
1. **Build instructions provided for MacOS**: previously there was just a build instruction for Windows. The instructions for MacOS may also work for a Linux machine (e.g., Ubuntu), but I didn't get around to setting up an environment to test this on.
2. **Flexible configuration of input/output data**: previously, the S-Function was built statically such that it would need to be reprogrammed and recompiled in order to support additional input/output. Now, the S-Function takes an XML file as input in the following way:
```xml
<?xml version="1.0"?>
<s_function_config>
  <input>
    <!-- Note: Properties here must be readable. -->
    <property> path/to/input/property-1 </property>
    <!-- Example:
    <property> fcs/elevator-cmd-norm </property>    
     -->
    <property> path/to/input/property-n </property>
  </input>
  <outputs>
    <!-- Note: name attribute is not necessary for function. Simply for organizing. -->
    <output name="Output 1">
        <!-- Note: Properties here must be writable. -->
        <property> path/to/input/property-1 </property>
        <!-- Example:
        <property> velocities/vc-kts  </property>    
         -->
        <property> path/to/input/property-n </property>
    </output>
    <output name="Output 2">
      <!-- ... -->
    </output>
    <!-- You can have any number of output elements as you want. -->
  </outputs>
</s_function_config>
```
The input is taken as a file path in the S-Function parameters. If this path is changed after the S-Function is run, it will change the ports automatically to reflect the new configuration.
3. **Optionally, you can configure the XML to include weather input**: previously, weather input was not supported. Now, you can include it by having the optional `<weather>` element. Functionally, it works similarly to the `<input>` element, except it only accepts writable properties from the `atmosphere` directory.
4. **Some general error handling is provided in order to help inform user of issues**: If there are issues in `JSBSimInterface.cpp`, the methods return false. Error handling is provided in `JSBSim_SFunction.cpp` through `ssSetErrorStatus`. It then returns in order to terminate execution to provide further issues (such as MATLAB fully crashing).
5. **The parameters to the S-Function have been changed**: Now, they are `aircraft_name, sim_time_step, use_script, script_file, initial_conds_file, io_config_file`. The `script_file` is used only if `use_script` is provided. The `initial_conds_file` is used otherwise. The rationale for this is because `script_file` will include initial conditions as part of one of its elements. We can consider merging these to be a single parameter that's use depends on the `use_script` parameter.
6. **General code cleanup**: there was a lot of unused/messy code. In this iteration, I cut a lot of it. Previous discussion (https://github.com/JSBSim-Team/jsbsim/discussions/498#discussioncomment-1376062) mentions that some of it was for leaning into a flexible functionality feature set. I found this code to be mostly unnecessary in pursuing this  goal (as it would have slowed down the system due mentioned in other discussion: https://github.com/JSBSim-Team/jsbsim/discussions/530#discussioncomment-1790716) so I took the liberty to cut it out.

Finally, I experienced some issues when compiling on a Windows machine (which I figure is probably something wrong with the version of my C++ compiler considering some of the code was existing in the current version in this repo). In order to fix them we had to do two things:
1. In `JSBSimInterface::JSBSimInferface` we needed to remove the `.get()` from each of the shared pointers. For some reason, the compiler we were using optimized this so that the shared pointer that was returned just provided the raw pointer...
2. In `mdlUpdate` we needed to change the values in `ctrlVec` and `weatherVec` initialization to static values (e.g., 100). In new versions of the C++ compiling, they can be initialized dynamically, but for some reason the Windows compiler we used required these to still be static.


seanmcleod:
> probably something wrong with the version of my C++ compiler

Which compiler are you using on Windows?

> In order to fix them we had to do two things:

Have you only implemented the fixes/work-arounds locally? Didn't see an obvious commit based on the commit messages, but haven't checked the code to see if you included these in the pull request.

noah-curran:
> Which compiler are you using on Windows?
Microsoft Visual C++ 2022. I obtained it from MS Visual Studio. I don't ever dev on Windows, so apologies if this is the wrong one.
> Have you only implemented the fixes/work-arounds locally? Didn't see an obvious commit based on the commit messages, but haven't checked the code to see if you included these in the pull request.
Yes, only locally. Because I feel that this is something wrong with my choice of compiler and because one of the changes was in the original version of the S-Function, I kept these local changes out of this PR in order to avoid confusion.

seanmcleod:
> Microsoft Visual C++ 2022.
> In new versions of the C++ compiling, they can be initialized dynamically, but for some reason the Windows compiler we used required these to still be static.

Hmm, well 2022 is the latest, so I'd be surprised if an earlier version, i.e. 2019 would support the dynamic sizing and not the latest version.

Although having said that both 2019 and 2022 allow you to specify which C++ language standard to adhere to. So maybe in your 2022 setup you had set or defaulted to an older standard? This is a screen snippet from VS2019.

image

I'll approve the workflow run for this pull request, and let's see what the compile results are.

noah-curran:
Sounds good, but there is a solid chance the tests fail when it tries to verify the test model since I don't believe that I committed an updated test that makes use of the new features.

seanmcleod:
Yep, we don't want to rely on a `gcc` specific extension. We've standardized on C++14 for now.

https://github.com/JSBSim-Team/jsbsim/blob/76a4d4fb37e351714664119c61f2853ad23b7a15/src/CMakeLists.txt#L7

https://github.com/JSBSim-Team/jsbsim/blob/76a4d4fb37e351714664119c61f2853ad23b7a15/matlab/CMakeLists.txt#L11

I double-checked and we do build the Matlab S-Function for Windows, actually for Windows, MacOS and Linux.

noah-curran:
> I double-checked and we do build the Matlab S-Function for Windows, actually for Windows, MacOS and Linux.

Sorry, what I meant by not updating a new test is that it will try to run the `ex737cruise`, but this PR does not actually change anything in that file, so it will still assume the old build. When I fix the Windows compile bug I will also make changes to that file so that the step for testing functionality doesn't fail (seen in the S-Function build checks for Linux and MacOS, which both compile successfully unlike the Windows build check).

seanmcleod:
Yep, I noticed, they took a while to appear, that the Windows build failed as expected and the Linux and MacOS runs failed due to the test script failing.

bcoconni:
The compilation on Windows fails because the code is using variables to declare an array size.

```c++
int inputSize;
int weatherInputsize;

double ctrlVec[inputSize];  // Legal in g++ (with extensions), illegal in proper C++
double weatherVec[weatherInputSize];  // ditto
```

As is [explained in this StackOverflow question](https://stackoverflow.com/questions/15013077/arrayn-vs-array10-initializing-array-with-variable-vs-numeric-literal) this is legal in C but illegal in C++. `gcc` allows this as an extension but MSVC does not. So strictly speaking, the Windows build is correct in reporting an error while Linux and Mac OSX compilers are permissive.

A better, safer way is to use C++ `std::vector<>`

```c++
int inputSize;
int weatherInputsize;

vector<double> ctrlVec;
ctrlVec.reserve(inputSize);  // Declare a vector which size is at least equal to inputSize


vector<double> weatherVec;  // ditto
weatherVec.reserve(weatherInputSize);
```

matlab/JSBSim_SFunction.cpp
```diff
@@ -500,90 +553,52 @@ static void mdlOutputs(SimStruct *S, int_T tid)
...
-  {
-	  real_T *dx = ssGetdX(S);
-	  double *w2 =  (double *) ssGetDWork(S,2);
-	  for (int i = 0; i < ssGetDWorkWidth(S,2); i++)
-	 {
-		dx[i] = w2[i]; // outputs are the flight control outputs 
-	 }
+    InputRealPtrsType ctrlCmdInput = ssGetInputPortRealSignalPtrs(S, 0);
+    double* dWorkCtrlCmdIn = (double*) ssGetDWork(S, 0);
+    double ctrlVec[inputSize];
```
bcoconni, line 566: Declare arrays with a size that is unknown at compilation is illegal in C++. Better use `vector<>`:
```suggestion
    vector<double> ctrlVec;
    ctrlVec.reserve(inputSize);
```

matlab/JSBSim_SFunction.cpp
```diff
@@ -500,90 +553,52 @@ static void mdlOutputs(SimStruct *S, int_T tid)
...
+        ssSetErrorStatus(S, "Issue copying control inputs to JSBSim.\n");
+        return;
+    }
 
-  }
-#endif /* MDL_DERIVATIVES */
+    if (useWeather) {
+        InputRealPtrsType weatherInput = ssGetInputPortRealSignalPtrs(S, 1);
+        double* dWorkWeatherIn = (double*) ssGetDWork(S, 1);
+        double weatherVec[weatherInputSize];
```
bcoconni, line 581: Same as above, use `std::vector<>` instead of an illegal C++ syntax that is accepted as an extension by some compilers:
```suggestion
        vector<double> weatherVec;
        weatherVec.reserve(weatherInputSize);
```

bcoconni (changes requested):
Don't declare arrays with a variable size, use `std::vector<>` instead.

noah-curran:
> The compilation on Windows fails because the code is using variables to declare an array size.

Yep, knew this, but thank you for the suggestion of using vectors. I was going to do something like:
```C++
double* val = new double[variable];
```
I have tested it on my Windows (and MacOS) machine and this solution also works.
The vector way of doing it is probably the better way, but I was intentionally trying to avoid vectors if possible.

noah-curran:
Sorry for the delay on getting this fix committed. I was putting together a new test Simulink file to commit, and came across an interesting bug related to indexed properties. For instance: `propulsion/engine[0]/set-running` or `fcs/throttle-pos-norm[0]`. As in the original Simulink example, I'm using the Boeing 737. I guess I assumed I could just call these properties using:
https://github.com/JSBSim-Team/jsbsim/blob/4f8d48f33452578df766297cc4b8f54ec0c65788/src/input_output/FGPropertyManager.h#L386-L387

But I get this error:
> FGPropertyManager::GetNode() No node found for fcs/throttle-cmd-norm[0]

I noticed that I could index, so I thought perhaps this was the way to go:
https://github.com/JSBSim-Team/jsbsim/blob/4f8d48f33452578df766297cc4b8f54ec0c65788/src/input_output/FGPropertyManager.h#L388-L389
Again, same error... I'm wondering if there is something specifically related to these properties that might be causing the nodes to not be found? They definitely should be found, so maybe I'm just not understanding how to index properly.

seanmcleod:
Hmm, works fine in my testing.

```c++
  JSBSim::FGPropertyNode* throttle0 = FDMExec->GetPropertyManager()->GetNode("fcs/throttle-pos-norm[0]");
  JSBSim::FGPropertyNode* throttle1 = FDMExec->GetPropertyManager()->GetNode("fcs/throttle-pos-norm[1]");

  throttle0->setDoubleValue(0.25);
  throttle1->setDoubleValue(0.5);

  double t0 = throttle0->getDoubleValue();
  double t1 = throttle1->getDoubleValue();
```

I'm wondering, has the aircraft model been loaded when you call `GetNode()`?

noah-curran:
Ugh... That must be it. The aircraft model is not loaded beforehand because when a script is used the code just jumps right into it... There should be a callback function I can make use of, so I'll take a crack at this tonight.

This is very helpful and insightful! Thanks @seanmcleod for the tip!

seanmcleod:
> There should be a callback function I can make use of

Do you need a callback function? Can't you just delay initializing the properties until after you've called `OpenScript()`?

noah-curran:
Yeah, my thought is that that should work. I'm not sure why I was thinking I might have to do a callback; I feel like I remember there being an issue somewhere else, but it might be completely unrelated in this case. Only way to know is to try. I'll report back here tonight once I do.

bcoconni:
> 

> > The compilation on Windows fails because the code is using variables to declare an array size.
> 
> Yep, knew this, but thank you for the suggestion of using vectors. I was going to do something like:
> 
> ```c++
> double* val = new double[variable];
> ```
> 

Using the C++ `new` operator is a bad practice in 2022. Because `new` means `delete[]` which means memory leaks waiting to happen.

> I have tested it on my Windows (and MacOS) machine and this solution also works. The vector way of doing it is probably the better way, but I was intentionally trying to avoid vectors if possible.

If for some reason you dislike vectors then please use `std::array` but please do not use `new`.

agodemar:
@noah-curran great work so far!

noah-curran:
> If for some reason you dislike vectors then please use std::array but please do not use new.

This is not possible. `std::array` only allows a static input to its template argument. I'd rather not use `std::vector` and just delete allocated memory, but it's not my codebase, so I'll go with your choice of `std::vector`.

noah-curran:
Okay folks, these commits should address the issues we discussed above:
- I fixed the dynamic allocation of the arrays in order to be compliant with the base version of C++.
- I also added a new Simulink model (and an example XML file) that runs correctly with this code.
- While fixing these, I also noticed a couple of other bugs that went unnoticed, and cleaned them up quickly.

Looking forward to hearing what you all think!

bcoconni (approved):

bcoconni:
The PR has been merged.
Thanks for your contribution @noah-curran

noah-curran:
Thanks @bcoconni. I'm the future, if there are any issues with this code that others bring up, feel free to tag me.

seanmcleod:
@noah-curran did you see my comments about there only being inputs for one of the engines in the example. Plus the unnecessary mixture input. I'm guessing the original setup used something like the C-172 model and when it was replaced with the 737 the inputs weren't changed. You're definitely not going to demonstrate regular cruise flight with only 1 engine, zero rudder and aileron input etc. :wink:

noah-curran:
@seanmcleod I didn't but you're absolutely right! I just kept it conformed to the original example. Obviously you would need to use both engines to get this example to truly work. Perhaps a different PR can put together a better example (or several examples to demonstrate different flights/simulation possibilities), and corresponding documentation that can provide tips for expanding on the use of the S-Function. This example also doesn't demonstrate the weather inputs, so that's another area of improvement.

PR #667, opened 2022-06-23
Reduced Unreal project size -reduced 3d model size and removed unused assets.

gallonmate:


gallonmate:
wrong pull, wasn't finished yet.

PR #670, opened 2022-06-24
#666 - remove static var to support run in multi-thread env

### LINKED ISSUE #666: Support for run JSBSim instance in multi-thread environment
**Feature request**

**Describe the issue**
  I'm a developer, and used JSBSim as a C++ static library for a quiet long time. Everything works like a charms, until i try to run mutilple JSBSim instance conurrently. It will crash sometimes. After some debugging, i found that the message queue in FGJSBBase in a static variable. The thread try to pop a empty queue which clear by other thread lead to this issue.

```
 static std::queue <Message> Messages;
```

So i'd like to request a feature to support run JSBSim in multi-thread env.

**Environment:**
 - OS: macOS/Windows
 - JSBSim v1.1.11
 - C++

**Other information**
same type issue as #201

kemen209:
Temperary solution to fix #666


bcoconni:
Hi @kemen209, thanks for the PR but that's far too much complicated a change.
I'd prefer this being discussed in issue #666, if you don't mind.

Your PR is rejected.

Issue #684, opened 2022-07-25
Prefer PROJECT_SOURCE_DIR over CMAKE_SOURCE_DIR


Aaron-SP:
When bundling JSBSIM as an external project using CMAKE FetchContent, JSBSIM is not configured because it uses CMAKE_SOURCE_DIR instead of PROJECT_SOURCE_DIR.

[CMAKE Issue Explained](https://stackoverflow.com/questions/32028667/are-cmake-source-dir-and-project-source-dir-the-same-in-cmake)

I've made a fork with this change at [repo](https://github.com/Aaron-SP/jsbsim).

Does it make sense to make this change to expand packaging capabilities of JSBSim with CMAKE?

bcoconni:
Thanks for reporting this issue.
Would you mind submitting a pull request from your repo ?

Aaron-SP:
I have opened a pull request [here](https://github.com/JSBSim-Team/jsbsim/pull/689). This fixes my issue but I did not check building the extra modules such as Python etc. We might want to double check that any files that code generate CMAKE are also updated, not sure.

bcoconni:
Your pull request has been merged.
Thanks for your contribution 👍

PR #689, opened 2022-07-28
Prefer PROJECT_SOURCE_DIR over CMAKE_SOURCE_DIR

Aaron-SP:
This PR fixes issues with embedding JSBSim into a higher level repository where JSBSim is not at the highest CMAKE level. This supports use cases such as CMAKE FetchContent.

Linking original issue [here](https://github.com/JSBSim-Team/jsbsim/issues/684#issue-1317159634)


bcoconni:
> This fixes my issue but I did not check building the extra modules such as Python etc. We might want to double check that any files that code generate CMAKE are also updated, not sure.

[All the checks have passed](https://github.com/JSBSim-Team/jsbsim/actions/runs/2757069041).

bcoconni (approved):
Looks good to me.

Issue #698, opened 2022-08-07
Input sockets are broken on Windows


bcoconni:
### Discussed in https://github.com/JSBSim-Team/jsbsim/discussions/693

<div type='discussions-op-text'>

<sup>Originally posted by **achodoglugil** August  4, 2022</sup>
I set the input parameters like this in c1723.xm
![input](https://user-images.githubusercontent.com/74592595/182802102-e75f7864-0e2b-4f87-b3ce-059358bf8066.PNG)

I run the command `.\JSBSim --script=scripts/c1723.xml --realtime --suspend` and then from another terminal i run the command `telnet localhost 1137` and doesnt get the message `connected to jsbsim server`. I just get a blank terminal and cant type anything</div>

bcoconni:
As reported by @seanmcleod in https://github.com/JSBSim-Team/jsbsim/discussions/693#discussioncomment-3336344, the issue is due to an improper use of the `FGfdmSocket` members `sckt` and `sckt_in` as variables of the type `SOCKET`on Windows.

seanmcleod:
So looking at `socket()` and `accept()` documentation for Linux and presumably for other Unix based OS, it looks like -1 is returned for both to indicate failure.

Linux socket() - https://www.man7.org/linux/man-pages/man2/socket.2.html
Linux accept() - https://www.man7.org/linux/man-pages/man2/accept.2.html

So one option to potentially avoid a bunch of extra `#ifdefs` around every call to `socket()`, `accept()` etc. is to introduce:

```C++
#ifndef WINDOWS
#define INVALID_SOCKET -1 
#endif
``` 

So resulting in the following sorts of changes:

```c++
// Currently
sckt = sckt_in = 0;
// To
sckt = sckt_in = INVALID_SOCKET;

// Currently
  if (sckt) shutdown(sckt,2);
  if (sckt_in) shutdown(sckt_in,2);
// To
  if (sckt != INVALID_SOCKET) shutdown(sckt,2);
  if (sckt_in != INVALID_SOCKET) shutdown(sckt_in,2);

// Currently 
if (sckt_in > 0) {
// To
if (sckt_in != INVALID_SOCKET 0) {

// Currently
  if (sckt_in <= 0 && Protocol == ptTCP) {
// To
  if (sckt_in != INVALID_SOCKET && Protocol == ptTCP) {

// Currently
sckt_in = -1;
// To
sckt_in = INVALID_SOCKET;
```

[...]

PR #700, opened 2022-08-08
Fix the telnet connection on Windows

bcoconni:
This pull request should fix the bug reported in issue #698 and discussion #693:

* Error codes returned by `socket()` and `append()` should now be correctly handled on Windows (the `SOCKET` type is a pointer type).
* JSBSim no longer expects `telnet` to be in line mode. Actually, the [default `telnet` on Windows cannot be set to line mode](https://superuser.com/questions/1251991/is-it-possible-to-run-windows-telnet-in-linemode) so `recv()` sends one character at a time instead of a string. This PR modifies `FGInputSocket` to append the data received from the socket in a buffer. The buffer is parsed once it contains a CARRIAGE RETURN `\r` or a LINE FEED `\n`.

As per @seanmcleod's suggestion, dummy flags `INVALID_SOCKET` and `SOCKET_ERROR` are defined for non-Windows platforms to use the same syntax for all platforms and save a lot of `#ifdef`.

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -382,7 +382,7 @@ jobs:
         run: ctest -R fpectl --build-config RelWithDebInfo -V
       - name: Test JSBSim
         working-directory: build
-        run: ctest -j2 -E TestInputSocket --build-config RelWithDebInfo --output-on-failure
+        run: ctest -j2 --build-config RelWithDebInfo --output-on-failure
 
     # On failure, upload logs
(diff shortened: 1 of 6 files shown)


bcoconni:
I took the opportunity of this PR to fix the test `TestInputSocket` so that it should now work on Windows and MacOSX.

I removed the threading code altogether as this was not needed and the test is now taking place in a single thread. The resulting code is much more straightforward and achieves the goal of testing whether the input socket is working or not on the 3 platforms that we are supporting (instead of Linux alone).

PR #725, opened 2022-09-17
Simplify `FGCondition`

bcoconni:
The class `FGCondition` is currently using a `std::map` to convert strings such as `<`, `==`, `GT`, etc. to an `enum eComparison` that is later used to perform the comparison:

https://github.com/JSBSim-Team/jsbsim/blob/7b1549cf3428ad3cbb70e03d1cca7585df02e7ac/src/math/FGCondition.cpp#L200-L226

The problem is that a `std::map` instance `mComparison` is created by the method `FGCondition::InitializeConditionals()` for every single instance of `FGCondition`: there are as many copies of the map object `mComparison` than there are instances of `FGCondition`. This is suboptimal.

This PR uses a `static constexpr` object that is created once and only once during compilation. This PR is therefore a proof of concept of a replacement of a `static std::map<>` member by a `static constexpr std::array<std::pair<const char*, T>,N>` member which does not need to be populated by a special method `InitializeXYZ()` before being used, avoiding race conditions issues in the process.

The opportunity of this PR is also taken to replace the usage of `new` and `delete` by modern C++14 memory management (i.e. `std::shared_ptr<>` and `std::make_shared<>`).

The constructor `FGCondition(const std::string& test, std::shared_ptr<FGPropertyManager> PropertyManager, Element* el)` is also made `public` again in the process so that it can be used by `std::make_shared<>`. As a consequence, the unit test `FGConditionTest` is extended to test the new `public` constructor.

### DIFF
--- a/src/math/FGCondition.cpp
+++ b/src/math/FGCondition.cpp
@@ -38,6 +38,8 @@ INCLUDES
 #include <cstdlib>
 #include <stdexcept>
 #include <assert.h>
+#include <array>
+#include <utility>
 
 #include "FGCondition.h"
 #include "FGPropertyValue.h"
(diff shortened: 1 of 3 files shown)


agodemar:
Hi @bcoconni it looks like no conflicts have been spotted. You can merge your PR

bcoconni:
> Hi @bcoconni it looks like no conflicts have been spotted. You can merge your PR

Thanks. Done !

PR #820, opened 2023-01-28
Improvements to the Python examples

### LINKED ISSUE #798: The python example has fdm.FGFDMExec twice, with different explanations
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**

The `examples/python/PlotAoaVsCAS.py` script contains the following contradictory code:

```python
# Path to JSBSim files
PATH_TO_JSBSIM_FILES="../../"

```

and

```python
    # Points to jsbsim executable and make an object
```

Either `jsbsim.FGFDMExec` needs the path for the jsbsim directory, in which case the second code block is incorrect, or it needs the path for the jsbsim executable, in which case the "constant" in the first block is wrong.

**What is the current behavior?**

We can't run this example =)

**What is the motivation / use case for changing the behavior?**

Projects live or die based on having good docs and working examples =)

**Other information**

It's also curious that `jsbsim.FGFDMExec` gets run three times, does it need to be recreated every iteration? One would assume that once `fdm` has been set up, one simply keeps using that.

bcoconni:
A couple of changes that should make the Python examples even friendlier:
- The aircraft name is a global variable in `PlotAoaVsCAS.py`
- Log messages are disabled to avoid flooding the console.

This is following the discussion that took place in issue #798.

Fixes #798.

### DIFF
--- a/examples/python/TrimEnvelope.py
+++ b/examples/python/TrimEnvelope.py
@@ -1,7 +1,7 @@
 # TrimEnvelope.py
 #
 # Calculate the set of trim points for an aircraft over a range of airspeeds
-# and range of flight path angles (gamma). The required thrust and AoA is 
+# and range of flight path angles (gamma). The required thrust and AoA is
 # indicated via a colormap for each trim point.
 #
 # Copyright (c) 2023 Sean McLeod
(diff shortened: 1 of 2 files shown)


examples/python/TrimEnvelope.py
```diff
@@ -25,12 +25,15 @@
...
+PATH_TO_JSBSIM_FILES="../.."
 
-fdm = jsbsim.FGFDMExec(PATH_TO_JSBSIM_FILES)  
+# Avoid flooding the console with log messages
+jsbsim.FGJSBBase().debug_lvl = 0
+
+fdm = jsbsim.FGFDMExec(PATH_TO_JSBSIM_FILES)
 
 # Load the 737 aircraft model
-fdm.load_model('737')
```
seanmcleod, line 33: Don't you want to remove my hard-coded 737 and make it a global variable `AIRCRAFT_NAME` like you did in the `PlotAoAVsCAS.py` example?

bcoconni: I was assuming that, since there is a single occurrence of the aircraft name in `TrimEnvelope.py`, the user could figure out what to do to change the aircraft name. But for the sake of consistency, it might be better ti use `AIRCRAFT_NAME` in all cases indeed.

bcoconni: I'll fix that.

bcoconni: This is now fixed.

PR #829, opened 2023-02-06
C130.xml invalid location value fixed.

Zaryob:
I worked with various things. When I looked C130.xml realize that, main problem is not related with t56 motor. It is related with location of engine 2.

And also the other planes have an extra feed value. Now C130 works.

### DIFF
--- a/aircraft/C130/C130.xml
+++ b/aircraft/C130/C130.xml
@@ -112,7 +112,13 @@
     </ground_reactions>
     <propulsion>
         <engine file="t56">
+            <location unit="IN">
+                <x> 586.5 </x>
+                <y> -265 </y>
+                <z> -40 </z>
+            </location>
             <feed>0</feed>
+            <feed>4</feed>
             <thruster file="t56_prop">
                 <location unit="IN">
                     <x> 586.5 </x>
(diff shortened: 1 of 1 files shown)


aircraft/C130/C130.xml
```diff
@@ -112,7 +112,13 @@
     </ground_reactions>
     <propulsion>
         <engine file="t56">
+            <location unit="IN">
+                <x> 586.5 </x>
+                <y> -265 </y>
+                <z> -40 </z>
+            </location>
```
bcoconni, line 119: This code is useless and has been removed by the PR #562  (commit 150805eae056b37a169d75aa6971133f01b9747f).
Please remove that change from your PR.

aircraft/C130/C130.xml
```diff
@@ -128,10 +134,11 @@
         </engine>
         <engine file="t56">
             <feed>1</feed>
+            <feed>4</feed>
             <thruster file="t56_prop">
                 <location unit="IN">
                     <x> 586.5 </x>
-                    <y> -265 </y>
+                    <y> 265 </y>
```
bcoconni, line 141: With this change 3 engines will be located on the right hand side and only one on the left hand side. I am quite skeptical that this change is meaningful.

Granted however that the current version of c130 is erroneous since both right engines are located at the same `Y` coordinate which should not be the case (the same applies to the left engines except the sign of their `Y` coordinates are opposite).

bcoconni (changes requested):
It's unlikely that this PR is the correct fix for your problem since :
- It introduces non operational code (the `<location>` tag below an `<engine>` element).
- It sets 3 engines on the right and the last remaining of the 4 engines on the left.

Such a model will suffer from a huge torque to the left when the 4 engines will be at full power.

PR #830, opened 2023-02-06
Fix actuator printout

heshpdx:
We were printing out the pointer value for `lag` instead of the property value. Simple fix.


bcoconni (approved):
That's a 17 years old error 😄
Good catch !

PR #860, opened 2023-03-22
Aeromatic fix inputs

### LINKED ISSUE #787: Negative pointmass is calculated
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**
A negative pointmass Payload is calculated with a comment (in output file):
- in web aeromatic: "- xxx LBS + full (35202 LBS) fuel should bring model up to entered max weight",
- in command line aeromatic: "- xxx LBS should bring model up to entered max weight".

This should not happen, because a negative mass is:
- not allowed by schema, and
- physically impossible.

**What is the current behavior?**
Relevant input data in my case was:
MTOW: 34473 kg,
EmptyW: 23288 kg.
Question about max. allowed fuel mass is not asked.
As a result negative pointmass is calculated (which is bad).

**What is the expected behavior?**
I would expect to be able to enter maximum allowed fuel capacity (in LBS or kg). 
As it is now, aeromatic makes an assumption, which produces bad result.

**What is the motivation / use case for changing the behavior?**
Better user experience :)

**Please tell us about your environment:**

- Linux (Ubuntu 22.04)
- AeromatiC++ version 3.3.20
- https://jsbsim-team.github.io/aeromatic/ (as currently available)

**Other information**

If so requested I can provide full input data used, to help replicate this issue.

ermarch:
Harden input handling and report warnings and alerts if something seems wrong.
This might fix #787. If not it could always be opened again.


utils/aeromatic++/Aircraft.cpp
```diff
@@ -249,6 +249,21 @@ bool Aeromatic::fdm()
...
+        _max_weight = _empty_weight;
+        _empty_weight = tmp;
+        _warnings.push_back("Empty weight is set larger than maximum weigth, swapping.");
+    }
+
+    if (_max_weight == 0)
+    {
+        _alerts.push_back("Maximum weigth is set to zero. Guessing.");
+        _max_weight = 10000.0f;
+    }
```
bcoconni, line 265: I'd move this block before the test `if (_empty_weight > _max_weight)` above because if one defines an empty weigth of 11000 with no max weight then they would end up with an empty weight of 11000 and a max weight of 10000 and this would trigger the bug reported in #787

ermarch: With that input combination empty weight and max weight would get swapped making empty weight == 0 and max weight == 11000 which triggers the test further down the code:
```
    // estimate empty weight, based on max weight
    if (_empty_weight == 0) {
        _empty_weight = _max_weight * aircraft->get_empty_weight();
    }
```

Testing it for real results in:
```
    Maximum takeoff weight             : 11000.00
    Empty weight                       : 9240.00
```

bcoconni: Indeed. Thanks for your answer.

Issue #140, opened 2018-12-22
FlightGear: useless warnings


ghost:
Hi, I had posted this on FlightGear mailing list before I realised these warnings are actually in JSBSim source. I hope my claims are valid and I'm not misusing the systems too badly?

https://sourceforge.net/p/flightgear/mailman/message/36499706/
> When I start FG with some aircraft which has a lot of JSBSim code, for
> example my Tu-144, I see a lot of warnings that are just noise that
> drowns out really important things like the new warnings about missing
> <function> args, they just get no chance of being seen unless you look
> for them on purpose. When I get crash because of a missing property I
> also have to grep through the logs for possible offenders, because
> it's not always mentioned in the last few lines.
> 
> 1. "Missing" properties
> 
> FGPropertyManager::GetNode() No node found for fcs/canard/pos-norm-input
> 
> presumably caused by properties that are a filter with name="" that
> another filter created before the given one tries to read, but fails
> the first time because it's not there yet
> 
> This warning normally serves a good purpose everywhere else, but
> JSBSim will just exit if it finds anything that is "really" missing.
> 
> It might be that replacing name="" with <output> would fix it, but I
> can not imagine the amount of editing in many files of many aircraft,
> because I might not be the only one to have been using name="" for too
> long before finding out it causes this.
> 
> 2.
> 
> In file /usr/home/mike/src/git/ac/Tu-144/Tu-144-jsbsim/control.xml: line 180
>       Property fcs/canard/pos-norm is already defined.
> 
> These are of course about the properties that I have defined in the
> beginning of each JSBSim system as <property> that are replaced by
> filters with name="" on the same property name, in this case
> "fcs/canard/pos-norm".
> 
> I think this practice of defining <property> I came up with is a good
> one (and hope it catches on if it's not terrible misuse) and should
> not be discouraged by such warnings because:
> - this, interspersed with some comments, is a good way to clearly
>   document the input interface of the system
> - it allows you to remove for debugging some system, without editing
>   a single line in all the systems depending on it (otherwise you have
>   to either add every property used by the other system as <property>
>   or comment out everything that uses it -- which could be hundreds of
>   lines in dozens of places)
> - another good use is to define initial values for properties there
>   (such as panel switches) or dummy value for isolated testing,
>   because it's self-contained. The only other option is maintaining a
>   copy in -set.xml, which would be a huge pain to keep in sync.

Cheers,
Mike

bcoconni:
[...]

Issue #108, opened 2018-08-07
Start a rocket motor from JSBSim code without NASAL in Flightgear


abassign:
Hi, I would like to start some JATO (solid fuel) engines I made for the G91-R1B in Flightgear. If I start the engines by setting the parameter in a NASAL program:

setprop("controls/engines/engine[1]/throttle", 1);

Is all ok, the engine start and end the propulsion when finish the solid propellant.

But i I try by this command in JSBSim code:

```
        <switch name="fcs/throttle-cmd-norm[1]">
            <default value="0"/>
            <test logic = "AND" value = "1">
                systems/jato/ignition-on == 1
                systems/jato/rocket_number_1 == 1
            </test>
        </switch>
```

Where: 
`systems/jato/ignition-on == 1`
This is the flag that activates the ignition of the engine from the panel in the cabin. This flag lasts a couple of seconds because, as in reality, the solid fuel engine is started by means of an electro-activated pyrotechnic ignition device.
The:
`systems/jato/rocket_number_1 == 1`
is the flag that guarantees to start only that specific motor (in G91 there were 4 JATO of 1000 lib)

Honestly I do not understand where I'm wrong, can you help me?
Thank for the collaboration :)

abassign:
I solved the problem with the help of the friends of Discord, the problem is due to the fact that there is a binding between the Flightgear application and the JSBSim implementation. This means that the variable **fcs/throttle-cmd-norm[n]** (_where n is the engine number_) is continuously updated by Flightgear and therefore can not be changed directly. Unfortunately, the implementation of JSBSim for Flightgear does not currently have the ability to silence the **fcs/ throttle-cmd-norm[n]** variable synchronization function with the corresponding Flightgear **variable /controls/engines/engine[n]/throttle** So it's better to play cunning and edit the Flightgear assign **/controls/engines/engine[n]/throttle** directly by assigning the value that we would like to assume **fcs/throttle-cmd-norm[n]** at this point everything works perfectly.
For example, to activate a solid fuel rocket engine it is sufficient to set (even for a moment) the variable: **/controls/engines/engine[n]/throttle** to 1 via a switch method or a function fcs_function this is the code I used:

```
        <switch name="fcs/throttle-cmd-norm[1]">
            <default value="0"/>
            <test logic = "AND" value = "1">
                systems/jato/ignition-on == 1
                systems/jato/rocket_number_1 == 1
            </test>
            <output>/controls/engines/engine[1]/throttle</output>
[...]

PR #864, opened 2023-03-24
Avoid misleading messages about properties being already defined

bcoconni:
JSBSim issues warning messages `"Property x/y/z is already defined"` when encountering a property definition in an definition XML file while the property exists.

This warning message puzzles users as it is not obvious to understand what this message exactly means. This has been reported several times:
* Discussion #854
* Issue #140 
* Issue https://github.com/JSBSim-Team/jsbsim/issues/108#issuecomment-412089961

The code that issues this warning message is located in
https://github.com/JSBSim-Team/jsbsim/blob/e1af7cea6ce6dfc35cad99a005df6cce66a0b328/src/input_output/FGPropertyReader.cpp#L86-L110

It is a feature that allows to specify the initial value of a property:
```xml
<property value="1.0"> system/variable </property>
```
This statement is very similar to a variable declaration in C++ such as `double x=1.0;`.

When JSBSim meets this statement, it creates the property `system/variable` and sets its value to `1.0`.

However if the property `system/variable` already exists then JSBSim considers that the property has already been defined (and possibly initialized) elsewhere. So JSBSim ignores this XML statement and issues a warning message `Property system/variable is already defined` to avoid **overriding** the property initialization (possibly with an inconsistent value).

Alternatively, JSBSim allows declaring a property without the `value` attribute:
```xml
<property> system/variable </property>
```
This variant of the statement is similar to a C++ variable declaration without an initialization such as `double x;`. It is meant to tell JSBSim to create a property if it does not already exist. This is useful because some XML statements might need a property to exist at the moment of their creation while the property is actually defined at some other place in the XML aircraft definition files. Once again, this is very similar to the declaration of a C++ global variable or function.

If the property `system/variable` already exists then JSBSim issues the warning message `Property system/variable is already defined.`

The point of this PR is that it is useless to issue this warning message when the XML statement does not specify the `value`. As nothing wrong can derive from ignoring the XML statement `<property> system/variable </property>` when the property `system/variable` already exists.

This would save a number of warning messages that are issued when loading some FDM, including some of the example models that are distributed with JSBSim.

---

I also took opportunity of this PR to rename the variable `override` to `override_props` since, as @ermarch mentioned in the discussion #854, `override` is a C++ identifier since C++11. [Strictly speaking `override` is not a reserved keyword](https://en.cppreference.com/w/cpp/language/override) and compilers do not issue warning messages when using `override` as a variable name (neither gcc, nor MSVC or clang). However it is most likely not good practice to do so as it might bring confusion. What about
```c++
void override(bool override) override;
```
(which is legal in C++) :wink: ?

### DIFF
--- a/src/input_output/FGPropertyReader.cpp
+++ b/src/input_output/FGPropertyReader.cpp
@@ -64,12 +64,12 @@ bool FGPropertyReader::ResetToIC(void)
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-void FGPropertyReader::Load(Element* el, FGPropertyManager* PM, bool override)
+void FGPropertyReader::Load(Element* el, FGPropertyManager* PM, bool override_props)
 {
   Element *property_element = el->FindElement("property");
   if (property_element && FGJSBBase::debug_lvl > 0) {
     cout << endl << "    ";
-    if (override)
+    if (override_props)
       cout << "Overriding";
     else
       cout << "Declared";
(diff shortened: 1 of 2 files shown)


seanmcleod:
Looks good.

PR #867, opened 2023-03-25
Rudder kick example in python and JSBSim XML script

seanmcleod:
Newcomers to JSBSim sometimes ask how to provide some scripted control inputs to get an aircraft to perform some maneuver. Currently none of the existing python examples in the repo demonstrate controlling control surfaces, both of them run through generating various trim solutions.

So this rudder kick example demonstrates how to go about inputting a rudder and aileron deflection for a particular flight test maneuver. An equivalent version based on JSBSim XML scripting is also provided so a user can compare and decide whether to go with a python based approach or an XML script approach.

### DIFF
--- a/examples/python/RudderKick.py
+++ b/examples/python/RudderKick.py
@@ -84,11 +84,9 @@ fdm.run_ic()
 try:
     fdm['simulation/do_simple_trim'] = 1
 
-except RuntimeError as e:
-    # The trim cannot succeed. Just make sure that the raised exception
-    # is due to the trim failure otherwise rethrow.
-    if e.args[0] != 'Trim Failed':
-        raise
+except jsbsim.TrimFailureError:
+    print("Trim failed, continuing rudder kick in an untrimmed state.")
+    pass  # Ignore trim failure
 
 # Time to run for in seconds
 run_period = 20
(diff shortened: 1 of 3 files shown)


examples/python/RudderKick.py
```diff
@@ -0,0 +1,133 @@
...
+
+# Trim
+try:
+    fdm['simulation/do_simple_trim'] = 1
+
+except RuntimeError as e:
+    # The trim cannot succeed. Just make sure that the raised exception
+    # is due to the trim failure otherwise rethrow.
+    if e.args[0] != 'Trim Failed':
+        raise
```
bcoconni, line 91: You could use the `TrimFailureError` exception to catach trim failure as that would be more explicit than the legacy code that was used before the `TrimFailureError` exception was introduced by the commit a5484eeea23cb15492b9d73f7ef1e833e025a99c
```suggestion
except jsbsim.TrimFailureError:
    pass  # Ignore trim failures
```

bcoconni (commented):
Just a tiny comment about using the `TrimFailureError` exception.

examples/python/RudderKick.py
seanmcleod: Have done. As a separate pull request I'll also update the existing 2 python examples to make use of this trim specific exception.

PR #874, opened 2023-03-30
Removed unused code that prevents Macos/Linux compilation of the Unreal Engine plugin.

demonixis:
This PR removes the `enum_traits` in `props.hxx` because it's not used and break the compilation of the Unreal Engine plugin on Macos and Linux.

I've compiled the JSBSim and JSBSimUnreal solution without any issue.

### DIFF
--- a/src/simgear/props/props.hxx
+++ b/src/simgear/props/props.hxx
@@ -1878,30 +1878,6 @@ inline std::string getValue<std::string>(const SGPropertyNode* node)
     return node->getStringValue();
 }
 
-/** Extract enum from SGPropertyNode */
-template<typename T>
-#if PROPS_STANDALONE
-inline T
-#else
-inline typename boost::enable_if<boost::is_enum<T>, T>::type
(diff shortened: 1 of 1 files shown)


seanmcleod:
I've approved the workflow to be run so that we can confirm there are no build issues across all the build types we perform and no errors from any of the tests we have.

demonixis:
I see many errors, have you a doc about all these targets? A workarround could be to add an specific define in case we're compiling with Unreal.

seanmcleod:
> @demonixis, actually you can safely delete the code lines 1881-1929 in src/simgear/props/props.hxx, these are unused by JSBSim.

But @demonixis you didn't follow @bcoconni's instructions carefully enough, take note of the line numbers he mentioned versus what you deleted.

demonixis:
@seanmcleod you're right! It's fixed, you can restart the tests

seanmcleod:
Just kicked them off.

Issue #786, opened 2022-12-16
Unreal Engine 5.1 Airplane starts underground


johndcollins:
**I'm submitting a ...**
  - [x ] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**

When starting the game the aircraft starts halfway underground

**What is the current behavior?**

**What is the expected behavior?**

Aircraft starts above ground.

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

**Other information**

![Screenshot 2022-12-16 090439](https://user-images.githubusercontent.com/1684130/208151221-7e4a4b5f-072b-4da8-845c-c954a27afe4a.png)

gallonmate:
Can you confirm if you edited or changed any settings in the default project or settings?
For example did you load a different flight model, other than the 787?
 
If you can attach the UE5 log or past the relevant snippet of it. As well as the flight debug info, press the D key to bring up the debug info. (I haven't updated my own UE5 to 5.1 yet, I will confirm later if once I update to 5.1 if is find any obvious bugs or not.)

johndcollins:
Nothing was edited or changed.

I'm not sure what log entries you're looking for:
LogJSBSim: Display: UJSBSimMovementComponent::LoadAircraft 787-8
LogJSBSim: Display: Initializing JSBSimFlightDynamicsModel using Data in 'F:/Games/Projects/jsbsim-master/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Resources/JSBSim'
LogJSBSim: Display: Model 787-8 Loaded successfully !
LogJSBSim: Display: PrepareJSBSim - Setting Initial Conditiond and computing initial state
LogJSBSim: Display: Trim Complete

![Screenshot 2022-12-17 154238](https://user-images.githubusercontent.com/1684130/208270032-7f99e7c0-3c9c-4d54-83ea-c0a13cb2be4b.png)

gallonmate:
I updated to UE5.1 and can confirm this bug exists in the default reference project. It's likely some issues with the Geo plugin or UE world coordinates, as UE5.1 includes updates such as:
> Large World Coordinates World Max Increase: New for this release, we raised the default world extents from 22 km to 88,000,000 km, allowing for much larger worlds to be created. 

[...]

PR #877, opened 2023-04-03
Fixed Bug #786

AlbanBERGERET-Epic:
Added a property to control the AGL raycast vertical offset, otherwise the aircraft can sink in ground at Trim time


bcoconni:
PR merged. Thanks @AlbanBERGERET-Epic.

PR #881, opened 2023-04-08
Move the Mach/VCAS conversion methods to `FGAtmosphere`

bcoconni:
Following the on-going process of removing `static` variables and methods as motivated by the issue #666, this PR moves the conversion methods between the Mach number and the calibrated airspeed (VCAS) to the `FGAtmosphere` class. This allows `StdDaySLsoundspeed` no longer being a static member and removes 4 static methods from `FGJSBBase`.

In addition the conversion methods `VcalibratedFromMach` and `MachFromVcalibrated` now use the altitude rather than the ambient pressure as an input parameter. This change makes the methods consistent with other methods from `FGAtmosphere`. In particular this avoids interrogating the `FGAtmosphere` class for the ambient pressure before calling a static method from `FGJSBBase` which simplifies the code in `FGInitialConditions`.

"Magic" constants have also been replaced by their expression using the specific heat ratio $\gamma$ as found in classical textbooks. Most of these computations use `constexpr` meaning that the constant is computed at compile time. These changes have therefore no performance impact in most cases (as the compiled code use a constant) but this improve the code legibility and allows to document how the constants are computed.

Finally the unit tests for these conversion methods have been rewritten and now check that the computation made by the conversion methods are consistent with thermodynamics:
* Mass conservation $\rho_1u_1=\rho_2u_2$
* Momentum conservation $p_1+\rho_1u_1^2=p_2+\rho_2u_2^2$
* Energy conservation $C_pt_1+\frac{1}{2}u_1^2=C_pt_2+\frac{1}{2}u_2^2$

where $p$ and $t$ are static pressure and static temperature. This allows removing the usage of hardcoded values and therefore testing a greater range of values and scenarios.


seanmcleod:
Glancing at the diffs it looks good.

bcoconni:
> Glancing at the diffs it looks good.

Thanks ! The PR is merged.

PR #883, opened 2023-04-08
User

ermarch:
Allow setting the surface bump height by the calling program.

This makes it possible to have a user specific way of handling ground displacement for the likes of rocky terrain or even wave following.

### DIFF
--- a/src/models/FGSurface.cpp
+++ b/src/models/FGSurface.cpp
@@ -52,7 +52,6 @@ CLASS IMPLEMENTATION
 FGSurface::FGSurface(FGFDMExec* fdmex, int number) :
    contactNumber(number)
 {
-  _PropertyManager = fdmex->GetPropertyManager();
   resetValues();
 }
 
@@ -78,10 +77,8 @@ void FGSurface::resetValues(void)
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-void FGSurface::bind(void)
+void FGSurface::bind(FGPropertyManager* PropertyManager)
 {
-  if (!_PropertyManager) return;
-
   string base_property_name;
   string property_name;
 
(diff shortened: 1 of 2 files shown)


bcoconni:
Bump height, ground normals and velocities should be specified via `FGGroundCallback`. `FGSurface` is a duplication of `FGGroundCallback` and will be removed at some point so this PR will not be merged.

ermarch:
You're right, both the patch and the GetBumpHeight function are unnecessary. And it was an easy for too for FligthGear.

But 90% of the FGSurface class handles surface properties which are not handled anywhere else in the code. It, for example, allows different breaking conditions between dry runway and frozen lakes in FlightGear.

bcoconni:
> But 90% of the FGSurface class handles surface properties which are not handled anywhere else in the code. It, for example, allows different breaking conditions between dry runway and frozen lakes in FlightGear.

Well, the code is essentially a collection of trivial setters and getters
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGSurface.h#L79-L113
* The method `GetBumpHeight` should be moved to `FGGroundCallback` inherited classes as is exemplified by the current PR so it can be removed.
* The method `FGSurface::_CreateIndexedPropertyName` is a duplication (i.e. a strict copy) of `FGJSBBase::CreateIndexedPropertyName` so it can be removed.
* `FGSurface` duplicates the "*gear number*" with a private member `contactNumber` which is a strict copy of `FGLGear` member `GearNumber`. This is useless as `FGLGear` inherits from `FGSurface`:
  * `contactNumber` definition in `FGSurface.h`
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGSurface.h#L130-L131
  * `contactNumber` initialization in `FGSurface.cpp`
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGSurface.cpp#L52-L54
  * `FGLGear` inherits from `FGSurface`
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGLGear.h#L192
  * The same value `number` is passed to the `FGSurface` constructor and to the `FGLGear` member `GearNumber`
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGLGear.cpp#L72-L76
* `FGSurface` member `ContactType` duplicates (and [is shadowed by](https://en.wikipedia.org/wiki/Variable_shadowing)) the `FGLGear` contact type (`ContactType`)
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGSurface.h#L66
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGLGear.h#L221
* On the other hand, the members `staticFCoeff` and `dynamicFCoeff` have been moved to `FGSurface` for no reason as they are only accessed by `FGLGear`.
* `FGSurface` adds the members `staticFFactor`, `rollingFFactor`, `maximumForce`, and `isSolid` to the `FGLGear` class but these members are **always** overwritten by the values coming from `FGGroundReactions` (which also inherits from `FGSurface`)
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGLGear.cpp#L304-L312
In the code above, the variable `surface` is **never** null as `FGLGear::GetBodyForces` is only called from `FGGroundReactions::Run` which passes `this` as an argument to `FGLGear::GetBodyForces` and of course `this` can never be null:
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/src/models/FGGroundReactions.cpp#L103
As as a consequence, we always have `surface == static_cast<FGSurface*>(GroundReactions)` and `(*surface).GetXXX()` is just a convoluted way of calling `GroundReactions->GetXXX()`. So passing `this` to `FGLGear::GetBodyForces()` is useless.

All this means that `FGLGear` inherits from `FGSurface` for no reason: either the inherited members are duplicated or they are systematically overwritten before being used. Even the properties created when calling `FGSurface::bind()` from `FGLGear` are nothing more than aliases from `FGGroundReactions` properties created by `FGSurface::bind()`. So at the very minimum `FGLGear` does not need to inherit from `FGSurface`.

bcoconni:
I've opened a new PR #884 that addresses the topic of `FGSurface`.
I suggest that the discussion should be continued in PR #884 and this PR will be closed as @ermarch confirmed that the patch and the GetBumpHeight function are unnecessary.

PR #891, opened 2023-04-16
Unreal: Added a way to load aircraft outside of the plugin

demonixis:
Hello,
This PR adds the possibility to load aircraft from another location than the JSBSim Unreal Plugin.
This is useful when working on custom aircrafts that are not shipped with the plugin. It also makes the update process easy because custom aircraft are stored elsewhere.


UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Public/JSBSimMovementComponent.h
```diff
@@ -197,6 +197,11 @@ class JSBSIMFLIGHTDYNAMICSMODEL_API UJSBSimMovementComponent : public UActorComp
 	UPROPERTY(Transient, BlueprintReadOnly, VisibleAnywhere, Category = "State")
 	FAircraftState AircraftState;
 
+	UPROPERTY(Transient, BlueprintReadOnly, VisibleAnywhere, Category = "Settings")
+	bool bUseExternalAircraftPath;
```
bcoconni, line 201:
```suggestion
	bool UseExternalAircraftPath;
```
JSBSim does not use [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) nor does the Unreal plugin AFAICT.
Could you please stick to the naming convention of the plugin ?

demonixis: This is the Unreal Engine naming convention which is used in all the plugin. Everything is pascal case and boolean starts with b.

bcoconni: Could not find any `bool` in the plugin nor any occurrence of this notation. Could you please point out one example to support your claim ?

demonixis: The plugin uses the Unreal naming convention, but I can change this variable name if it's not correct for the repository.

bcoconni: > This is the Unreal Engine naming convention which is used in all the plugin.

> The plugin uses the Unreal naming convention

Hmm, I think I got the idea the first time you wrote it.

Could you please provide an example in the plugin ? Just one line in one file of this plugin and I set this conservation as resolved and it saves you the trouble of changing the variable name.

bcoconni: Found it for you
https://github.com/JSBSim-Team/jsbsim/blob/92fa1eaa29322dc3f6ab55587161e55a6b24d36d/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp#L206

bcoconni: The topic is resolved.

bcoconni:
Looks good to me.
@AlbanBERGERET-Epic, @gallonmate any comments ?

AlbanBERGERET-Epic:
Hmmm, it looks like the external path will be treated as an absolute path, which is dangerous in case you package and deploy the software to another computer. 

Imagine you point to a D:\MyAircrafts and you deploy to a computer with only a C drive. 

I would recommend adding a JSBSim config class with an ExternalRepositoryFolder property (and a useExternalRepo bool on the component)
https://docs.unrealengine.com/5.1/en-US/configuration-files-in-unreal-engine/

That way, you could potentially set it from the final application. 

And I thing the ContentDir specifier for a FDirectoryPath UProperty will offer a folder picker to select it.

demonixis:
That's a good idea, I'll take some time to update this PR using config files

PR #892, opened 2023-04-16
Unreal: Added an event triggered when the aircraft crashed

demonixis:
Hello,
This PR adds an event which is triggered when the aircraft crashed, it's useful to reload the simulation or doing anything else.


seanmcleod:
How reliable is the crash detection, I realise you didn't implement the original crash detection logic, but aren't there going to be a lot of crashes that don't end up with the cg 10ft underground?

```c++
	if (AircraftState.AltitudeAGLFt < -10.0 || AircraftState.AltitudeASLFt < -10.0) {
		Exec->SuspendIntegration();
		AircraftState.Crashed = true;
	        AircraftCrashed.Broadcast();
	}
```

demonixis:
This is the original code, I just added an event to be able to react when it's happen. However we can improve the detection in a separate pull request for sure.

seanmcleod:
Yep, there is a whole discussion on crash detection in JSBSim, see - https://github.com/JSBSim-Team/jsbsim/discussions/819

bcoconni:
Looks good to me.
@AlbanBERGERET-Epic, @gallonmate, any comments ?

AlbanBERGERET-Epic:
Good to me. 

Actually the current isCrashed algorithm is very simple when checking the ground level... IDK if JSBSim can return other events when for instance a wing hit the ground, which is bad in general. It could be a further improvement.

bcoconni:
Thanks for the PR @demonixis
Thanks for the feedback @AlbanBERGERET-Epic and @gallonmate.
The PR is now merged.

PR #898, opened 2023-04-25
Move the Mach/VCAS conversion methods to `FGAuxiliary`

bcoconni:
This PR is a counter proposal to the PRs #881 and #885 that I previously submitted. The motivations remain the same than the ones exposed in #881 while avoiding the problems that the PR #885 is awkwardly trying to fix. In the current PR, the Mach/VCAS conversion methods are moved to `FGAuxiliary`. As such this PR undoes #881 and supersedes #885.

Basically the idea is to move the Mach/VCAS conversion routines to `FGAuxiliary` rather than to `FGAtmosphere` as the PR #881 did. For that purpose, `FGAuxiliary` only needs an access to `FGAtmosphere` constants such as the specific heat ratio and the standard day pressure, temperature and speed of sound. As these are `constexpr` members, `FGAuxiliary` can safely access them using C++ qualified identifiers such as `FGAtmosphere::StdDaySLpressure` which is much simpler than #881 and #885 combined. Even though the standard day speed of sound needs a special treatment as the function `sqrt` cannot be used for `constexpr` members. So we need to fall back to a good old `const` variable for `StdDaySLsoundspeed`.

IMHO, with this PR, all the pieces are finally falling into place and I'd suggest to merge this PR and cancel #885.


seanmcleod:
Looks good.

PR #900, opened 2023-04-27
Correct VEAS to use standard day SL density

seanmcleod:
Slightly messy layout in FGAtmosphere.h in terms of ideally wanting to add `StdDaySLdensity` to the existing group of public constants.

```c++
  static constexpr double StdDaySLtemperature = 518.67;
  static constexpr double StdDaySLpressure = 2116.228;
  const double StdDaySLsoundspeed;
  static constexpr double SHRatio = 1.4;
```

But require `Reng0` to be defined first, but it's currently lower down in a `protected` section.

So couple of options in terms, not sure if the one I've chosen is the best.

### DIFF
--- a/src/FGFDMExec.cpp
+++ b/src/FGFDMExec.cpp
@@ -464,7 +464,6 @@ void FGFDMExec::LoadInputs(unsigned int idx)
   case eAuxiliary:
     Auxiliary->in.Pressure     = Atmosphere->GetPressure();
     Auxiliary->in.Density      = Atmosphere->GetDensity();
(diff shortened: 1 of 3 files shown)


bcoconni (approved):
Looks good to me.

bcoconni:
> But require `Reng0` to be defined first, but it's currently lower down in a `protected` section.

Not sure there is much reason for this constant to be hidden behind a `protected` wall. At some point maybe it should be renamed to `R_dry_air` or any name that is more explicit than `Reng0` ?
> So couple of options in terms, not sure if the one I've chosen is the best.

You've chosen the most conservative which is fine as far as I am concerned 😄.

PR #902, opened 2023-05-02
Unbind `FGAtmosphere` properties

bcoconni:
As I mentioned previously in https://github.com/JSBSim-Team/jsbsim/pull/885#issuecomment-1513821802, I'm currently working on adding the [NRLMSIS2.0 atmospheric model](https://ccmc.gsfc.nasa.gov/models/NRLMSIS~v2.0/) to JSBSim which is kind of a revival of the existing `src/models/atmosphere/FGMSIS.cpp` code (which was based on the previous version 1.0 of NRLMSIS)[^1].

In order for this feature to work, we will need to modify the atmosphere model that is currently created by default in `FGFDMExec`:
https://github.com/JSBSim-Team/jsbsim/blob/39e70585021752fe9a68061b556d496e3587b22f/src/FGFDMExec.cpp#L229

Basically, the idea is to call at some point the following code:
```c++
if (some_XML_instructions_are_found) {
  // Delete the FGStandardAtmosphere instance and replace it by an FGMSIS instance
  Models[eAtmosphere] = std::make_shared<FGMSIS>(this);
  // ERROR messages : Failed to tie property XXX to object methods.
}
```
The trick is that if no special care is taken, the properties `atmosphere/T-R`, `atmosphere/P-Psf`, etc. will still be bound to the old instance of `FGStandardAtmosphere` even after the above code has been executed. With the current code, `FGMSIS` would fail to bind itself to the aforementioned properties as they would already be tied to `FGStandardAtmosphere` and JSBSim would complain with the infamous error message `Failed to tie property atmosphere/T-R to object methods`. Even worse, if one would try to access `atmosphere/T-R` and the likes, it would cause a SEGFAULT as the property would try to call the method `FGAtmosphere::GetTemperature()` on a deleted `FGStandardAtmopshere` instance.

To avoid this issue, we need to tell JSBSim to unbind `FGStandardAtmosphere` before an instance of `FGMSIS` is created. Currently there is no generic code in JSBSim that allows to unbind a single `FGModel` instance (you'd need to untie the properties one by one which is less than ideal from a maintenance standpoint).

To address this issue the current PR adds a new method `FGPropertyManager::Unbind(void* instance)` to unbind the properties of a given class instance. With this new method, the code above will become something like:
```c++
if (some_XML_instructions_are_found) {
  PropertyManager->Unbind(Models[eAtmosphere]);
  Models[eAtmosphere] = std::make_shared<FGMSIS>(this);
  // SUCCESS! No "Failed to tie..." error messages.
}
```

The new `Unbind` method also restores the read/write attributes of each properties as the newly bound instance might want to set them differently.

Finally, the unit test `FGAtmosphereTest` is amended by the PR to test the new feature and checks that the properties such `atmosphere/T-R` are tied to the correct instance (see the method `testRun()` in `FGAmosphereTest`).

[^1]: For the record the code in `FGMSIS.cpp` is dead code as there is currently no means to execute this code from JSBSim.

### DIFF
--- a/src/input_output/FGPropertyManager.cpp
+++ b/src/input_output/FGPropertyManager.cpp
@@ -53,10 +53,25 @@ namespace JSBSim {
 
 void FGPropertyManager::Unbind(void)
 {
-  for(auto& prop: tied_properties)
-    prop->untie();
+  for(auto& property: tied_properties)
+    property.untie();
 
   tied_properties.clear();
+}
+
+//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+void FGPropertyManager::Unbind(void* instance)
+{
+  auto it = tied_properties.begin();
+
+  while(it != tied_properties.end()) {
+    auto property = it++;
+    if (property->BindingInstance == instance) {
+      property->untie();
+      tied_properties.erase(property);
+    }
+  }
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -305,8 +320,8 @@ void FGPropertyManager::Untie(SGPropertyNode *property)
   assert(property->isTied());
 
   for (auto it = tied_properties.begin(); it != tied_properties.end(); ++it) {
-    if (*it == property) {
-      property->untie();
+    if (it->node.ptr() == property) {
+      it->untie();
       tied_properties.erase(it);
       if (FGJSBBase::debug_lvl & 0x20) cout << "Untied " << name << endl;
       return;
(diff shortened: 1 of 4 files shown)


seanmcleod:
Assuming we're not planning on allowing a user to switch atmospheric models part-way through a simulation, i.e. a single atmospheric model is chosen/set at startup then instead of having to handle, binding, unbinding, binding again can't we simply create the relevant instance of the atmosphere model at:

https://github.com/JSBSim-Team/jsbsim/blob/39e70585021752fe9a68061b556d496e3587b22f/src/FGFDMExec.cpp#L229

bcoconni:
> Assuming we're not planning on allowing a user to switch atmospheric models part-way through a simulation,

I confirm there is no such plan.
> i.e. a single atmospheric model is chosen/set at startup then instead of having to handle, binding, unbinding, binding again can't we simply create the relevant instance of the atmosphere model at:
> 
> https://github.com/JSBSim-Team/jsbsim/blob/39e70585021752fe9a68061b556d496e3587b22f/src/FGFDMExec.cpp#L229

Unfortunately we can't because the line you mentioned is called from the `FGFDMExec` constructor but it's only when `FGFDMExec::LoadModel()` parses the XML files that you discover which atmospheric model should be used.

PR #909, opened 2023-05-08
Use standard day SL density for VEAS in `FGInitialCondition`.

bcoconni:
Following the bug fix that has been released by @seanmcleod in PR #900, this PR applies the same fix to the `FGInitialCondition` class (it is not using the standard day SL density either for the computation of the [equivalent airspeed](https://en.wikipedia.org/wiki/Equivalent_airspeed)).

I took this opportunity to check that `FGAtmosphere::GetDensitySL()` was not called elsewhere and remove the member `FGAtmosphere::in.DensitySL` that is no longer needed. In the process I also removed the member `FGAtmosphere::in.PressureSL` which is not used anywhere.


seanmcleod:
Good catch.

bcoconni:
> Good catch.

@agodemar If this PR looks good to you as well then I guess you can push the button 😃

Issue #808, opened 2023-01-09
Compilation Failed While Opening in UE 5.1


junaidbinsarfraz:
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**
Cloned the master branch and try to open `jsbsim/UnrealEngine/UEReferenceApp.uproject` from Unreal Engine 5.1. But it didn't open the project rather threw a compilation error

```
The project could not be compiled. Would you like to open it in Visual Studio?

Running C:/Program Files/Epic Games/UE_5.1/Engine/Build/BatchFiles/Build.bat Development Win64 -Project="D:/jsbsim/UnrealEngine 5.1 - 2/UEReferenceApp.uproject" -TargetType=Editor -Progress -NoEngineChanges -NoHotReloadFromIDE
Running UnrealBuildTool: dotnet "..\..\Engine\Binaries\DotNET\UnrealBuildTool\UnrealBuildTool.dll" Development Win64 -Project="D:/jsbsim/UnrealEngine 5.1 - 2/UEReferenceApp.uproject" -TargetType=Editor -Progress -NoEngineChanges -NoHotReloadFromIDE
Log file: C:\Users\PC\AppData\Local\UnrealBuildTool\Log.txt
Using 'git status' to determine working set for adaptive non-unity build (D:\jsbsim).
Creating makefile for UEReferenceAppEditor (no existing makefile)
JSBSim.dll not found in D:\jsbsim\UnrealEngine 5.1 - 2\Plugins\JSBSimFlightDynamicsModel\Source\ThirdParty\JSBSim\Lib - Make sure that you have built JSBSimForUnreal.sln first
Unable to instantiate module 'JSBSim': JSBSim.dll not found in D:\jsbsim\UnrealEngine 5.1 - 2\Plugins\JSBSimFlightDynamicsModel\Source\ThirdParty\JSBSim\Lib - Make sure that you have built JSBSimForUnreal.sln first
(referenced via Target -> JSBSimFlightDynamicsModel.Build.cs)
```

**What is the current behavior?**
![image](https://user-images.githubusercontent.com/18396574/211257571-670acb17-e54b-4880-93da-d63c0439c7aa.png)

**Please tell us about your environment:**
 - OS: Windows 10
 - JSBSim version: clone of master branch
[...]

Issue #740, opened 2022-09-30
Unreal 5, crash on play in editor instantly with default editor starting map. 


nonlin:
**I'm submitting a ...**
  - [X ] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**
Getting crash on play

**What is the current behavior?**
Fresh clone, build solution, dll made, build unreal project, launch, play in editor, instant crash

**What is the expected behavior?**
Play in editor doesn't cause a crash 

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

**Other information**
Stack trace 

Unhandled Exception: 0x80000003

```
ucrtbased
ucrtbased
ucrtbased
ucrtbased
JSBSim
JSBSim
JSBSim
JSBSim
UnrealEditor_JSBSimFlightDynamicsModel!UJSBSimMovementComponent::InitializeJSBSim() [J:\Git\jsbsim\UnrealEngine\Plugins\JSBSimFlightDynamicsModel\Source\JSBSimFlightDynamicsModel\Private\JSBSimMovementComponent.cpp:410]
UnrealEditor_JSBSimFlightDynamicsModel!UJSBSimMovementComponent::LoadAircraft() [J:\Git\jsbsim\UnrealEngine\Plugins\JSBSimFlightDynamicsModel\Source\JSBSimFlightDynamicsModel\Private\JSBSimMovementComponent.cpp:151]
UnrealEditor_JSBSimFlightDynamicsModel!UJSBSimMovementComponent::BeginPlay() [J:\Git\jsbsim\UnrealEngine\Plugins\JSBSimFlightDynamicsModel\Source\JSBSimFlightDynamicsModel\Private\JSBSimMovementComponent.cpp:351]
UnrealEditor_Engine!AActor::BeginPlay() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\Actor.cpp:3887]
UnrealEditor_Engine!AActor::DispatchBeginPlay() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\Actor.cpp:3849]
UnrealEditor_Engine!AWorldSettings::NotifyBeginPlay() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\WorldSettings.cpp:283]
[...]

PR #912, opened 2023-05-12
Changed UE5 plugin to default as Release build and fixed Debug build

gallonmate:
New users experience trouble building UE5 plugin due to mismatching default build settings between Visual Studio and UE5. Which causes an error until the user selects the correct matching build options. Examples of user issues: https://github.com/JSBSim-Team/jsbsim/issues/808 https://github.com/JSBSim-Team/jsbsim/issues/740 https://github.com/JSBSim-Team/jsbsim/discussions/664

Visual Studio defaults to the build option that is alphabetically first. so a simple fix is to rename the build options as:  1_Release and 2_Debug. This way new users will always by default build JSBSimforUnreal as Release. And errors from UE5 users should be reduced.

Another issue is the UE5 plugin's Debug build is broken for unknown reasons. A simple change by using the non-debug dll but with all debug options still enabled, can allow successful debugging of UE5 & JSBSim library. Original discussion here: https://github.com/JSBSim-Team/jsbsim/discussions/761

### DIFF
--- a/JSBSimForUnreal.vcxproj
+++ b/JSBSimForUnreal.vcxproj
@@ -56,7 +56,7 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(ProjectDir)src;$(ProjectDir)src\simgear\xml;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>JSBSIM_EXPORT;JSBSIM_VERSION="1.2.0.dev1";WIN32;NOMINMAX;XML_STATIC;_NDEBUG;_CONSOLE;HAVE_EXPAT_CONFIG_H;_USE_MATH_DEFINES;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>JSBSIM_EXPORT;JSBSIM_VERSION="1.2.0.dev1";WIN32;NOMINMAX;XML_STATIC;NDEBUG;_CONSOLE;HAVE_EXPAT_CONFIG_H;_USE_MATH_DEFINES;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <MinimalRebuild>false</MinimalRebuild>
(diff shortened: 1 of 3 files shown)


JSBSimForUnreal.vcxproj
```diff
@@ -56,10 +56,10 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(ProjectDir)src;$(ProjectDir)src\simgear\xml;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>JSBSIM_EXPORT;JSBSIM_VERSION="1.2.0.dev1";WIN32;NOMINMAX;XML_STATIC;_DEBUG;_CONSOLE;HAVE_EXPAT_CONFIG_H;_USE_MATH_DEFINES;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>JSBSIM_EXPORT;JSBSIM_VERSION="1.2.0.dev1";WIN32;NOMINMAX;XML_STATIC;_NDEBUG;_CONSOLE;HAVE_EXPAT_CONFIG_H;_USE_MATH_DEFINES;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
bcoconni, line 59: Shouldn't the flag `NDEBUG` be added to this list as well ?

According to [C++ docs](https://en.cppreference.com/w/cpp/error/assert), the flag `NDEBUG` need to be defined for the `assert` macro to check and report issues. Unless UE5 defines `NDEBUG` as soon as `_NDEBUG` is defined ?

gallonmate: :sweat_smile: Thanks! `_NDEBUG `was typo. Corrected to `NDEBUG`

gallonmate: Also if this PR is okay and merged, than Issue #808 is able to be closed

bcoconni:
Your PR is merged. Thanks @gallonmate !

PR #914, opened 2023-05-19
Update Concorde.xml

sryu1:
I'm not too sure how GNU GPL works, but it seems like flightgear's one was more up to date... If this violates the license of the file, please reject the PR


sryu1:
Actually, disregard this.

PR #916, opened 2023-05-27
Update the `FGMSIS` class

bcoconni:
This PR is a followup of the PRs #902 and #908.

This patch brings a major update to the `FGMSIS` class but does not plug it in yet. The ability to switch JSBSim atmosphere model to "MSIS" will be brought by another PR meaning that this PR code is non operational.

The changes are:
* The NRLMSIS00 code will be pure C code rather than being wrapped in a C++ class as is currently done in `FGMSIS.cpp`, `FGMSIS.h` and `FGMSISData.cpp`. This change is motivated by maintenance considerations: as the C code is kept pristine with respect to [Dominik Brodowski original code](https://www.brodo.de/space/nrlmsise/index.html), any update from Brodowski will only require a simple copy/paste of the files to update JSBSim rather manually updating some C++ code based on a diff.
* `FGMSIS` is now inheriting from `FGStandardAtmosphere` instead of `FGAtmosphere`. This is needed to compute the pressure and density altitudes.
  * The default value of `FGStandardAtmosphere::SaturatedVaporPressure` as been modified: it is now equal to `StdDaySLpressure` rather than being initialized to 0. This change is required to avoid a division by zero during the bind/unbind process (which is already tested by the unit test `FGMSISTest`).
* The management of the current day and hour has been moved from `FGAuxiliary` to `FGMSIS`. The latter is the only class that uses the current day and hour so this avoids splitting responsibilities between classes. IMHO would the current day and hour be needed by another class, I'd rather move them in `FGFDMExec` as it already manages the simulation time.
* A unit test `FGMSISTest` has been added.
* The CI workflow now validates the C code of NRLMSIS00 by checking that the output from `nrlmsise-00_test.c` complies with the reference output documented in the `DOCUMENTATION` file from D. Brodowski.

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -37,9 +37,34 @@ jobs:
               xmllint --noout --schema JSBSimSystem.xsd $filename
           done
 
+  MSIS-validation:
+    name: MSIS code validation
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{matrix.os}}
+    steps:
+      - name: Checkout JSBSim
+        uses: actions/checkout@v3
+      - name: Configure the MSIS test program
+        run: |
+          mkdir build && cd build
+          cmake ../src/models/atmosphere/MSIS
+      - name: Build the JSBSim test program
+        working-directory: build
+        run: make -j2
+      - name: Run the NRLMSIS-00 C package
+        working-directory: src/models/atmosphere/MSIS
+        run: |
+          if [[ $(head -261 DOCUMENTATION | tail -104 | diff -urN <(../../../../build/nrlmsise-test) - | wc -c) -ne 0 ]]; then
+            echo "Failed."
+            exit 1
+          fi
+
   Linux:
     name: C/C++ build (Linux)
-    needs: XML-validation
+    needs: [XML-validation, MSIS-validation]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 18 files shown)


bcoconni:
Right. It's been 6 weeks since I submitted this PR and it does not seem to attract much comment/attention/suggestion/criticism. I suppose that submitting an unsolicited feature was not a good idea and I no longer want to force my vision to the JSBSim community so I'm closing this PR.

As for the PRs #902 and #908, they are now irrelevant so you can revert them with the following commands:
```bash
> git revert db830c68
> git revert 868aed43
> git push origin master
```

agodemar:
@bcoconni sorry for my late message. I think you can merge your PR with the master branch. No problem on my side.

bcoconni:
Well, thanks @agodemar but what about you @seanmcleod ? I know you had some doubts about the features in the PR #902 and #908 and maybe we merged them too hastily ?

seanmcleod:
@bcoconni I suggest you continue/re-open this pull request. I hadn't specifically commented on this pull request since I didn't see any specific issues with it, i.e. "no news is good news" 😉  

I did raise some concerns with regards to #902 and #908 mainly in terms of wanting to keep things as simple and clear as possible, both in terms of users wanting to run their aircraft models with different atmospheres, different planet constants etc. and also in terms of trying to keep the required code changes as simple and clear as possible.

In terms of FDM users I suggested rather than adding the atmosphere selection and planet constants into the FDM that they should be specified separately/externally to the FDM via something similar to initial conditions, maybe "environment conditions"?

You had replied with:

> I hadn't thought about that and your suggestion is making sense. I guess we could add a method FGFDMExec::LoadPlanet() that would load the planet constants and the atmosphere before LoadModel() is called. I like your idea, I'll investigate it further 😄

So all good.

bcoconni:
Great ! Thanks for your feedback @agodemar and @seanmcleod 😄

PR #941, opened 2023-07-22
Enable the usage of the NRLMSIS00 atmospheric model.

bcoconni:
This PR is the last step to enable the MSIS atmospheric model in JSBSim (the previous PRs were #902, #908 and #916).

After this PR will be merged, the usage of the MSIS atmosphere will be triggered by the following XML code:
```xml
<planet name="Earth">
  <atmosphere model="MSIS">
    <day>172</day>    <!--  Jun 21st  -->
    <utc>29000</utc>  <!-- 08:03:20AM -->
  </atmosphere>
</planet>
```
The XML can either be embedded in an aircraft definition file or supplied to `JSBSim.exe` via the `--planet` option or loaded with the C++ method `FGFDMExec::LoadPlanet()` or the Python method `FGFDMExec.load_planet()`.

At the moment the date shall be supplied as the day of the year, counting from Jan 1st which is day 1 to Dec 31st which is day 365. The time zone is [UTC+00:00](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) and the time is supplied as the number of seconds elapsed since 00:00 (12:00AM).

This PR adds a test to `TestPlanet.py` to check that the pressure, density and temperature are computed by the MSIS model when the XML file above is supplied.

### DIFF
--- a/src/FGFDMExec.cpp
+++ b/src/FGFDMExec.cpp
@@ -46,6 +46,7 @@ INCLUDES
 
 #include "FGFDMExec.h"
 #include "models/atmosphere/FGStandardAtmosphere.h"
+#include "models/atmosphere/FGMSIS.h"
 #include "models/atmosphere/FGWinds.h"
 #include "models/FGFCS.h"
 #include "models/FGPropulsion.h"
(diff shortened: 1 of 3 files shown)


seanmcleod:
Looks good.

agodemar:
Sounds good. Please go ahead with the merge step

rickechler:
Hi, after these changes were merged, I was unable to build JSBSimForUnreal.sln. Got the error that FGMSISData.cpp was missing from src/models/atmosphere/. When I reverted, I was able to build. Thanks!
image

bcoconni:
@rickechler, this should now be fixed by the commit 8a807a7b6f59a6195c5b1c6bfc7df84f9a75ac01.

PR #944, opened 2023-08-04
Unreal Engine Mac compatibility

finger563:
This PR addresses the following issues related to the UnrealEngine JSBSim project:

1. The default `UnrealEngine` project does not support Mac as a build target
2. The existing movement component fails to compile on UE 5.2 Mac 

and additionally adds a `build_macos_ue5.sh` script which uses cmake to build the `jsbsim` library (.a), then copies the relevant headers, resources, and libs into the `ThirdParty` directory within the unreal engine plugin for building on MacOS.

### DIFF
--- a/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp
+++ b/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp
@@ -403,13 +403,15 @@ void UJSBSimMovementComponent::InitializeJSBSim()
 		UE_LOG(LogJSBSim, Display, TEXT("Initializing JSBSimFlightDynamicsModel using Data in '%s'"), *RootDir);
 
 		// Set data paths...
-		FString AircraftPath(TEXT("aircraft"));
-		FString EnginePath(TEXT("engine"));
-		FString SystemPath(TEXT("systems"));
-		Exec->SetRootDir(SGPath(*RootDir));
-		Exec->SetAircraftPath(SGPath(*AircraftPath));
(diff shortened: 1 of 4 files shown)


bcoconni:
Hi @finger563, thanks for the PR.

Is this based on @demonixis PR #873 ? If not, I'd suggest to coordinate both pull requests to get something consistent on both the Linux & MacOSX platforms. In particular, your PR does not seem to address the issue linked to the generation of a dynamic library (`*.dylib`) which is required by the UE license (see discussion #631 for details).

finger563:
Hi @bcoconni, thanks for getting back to me and letting me know. I actually hadn't seen his PR until after I'd already made my changes. I understand the need for dynamic linking in a packaged project (which mine is not) and since I understanf many people likely need to adhere to that use I'll update mine to use dynamic instead of static linking and then I'll see about coordinating to not have duplicate / overlapping PRs.

demonixis:
Hello,
Sorry for the wait about my PR, I'll update it today. My PR is not designed for Macos but Macos, Linux and Android (partially because it requires a change in JSBSim sources). 
@bcoconni I've an M2 mac and I wondering if it's a good idea to compile the lib with universal support or not. Today the lib is compiled for x86_64 or arm64. We can compile for both as an universal package. Is it relevant or not? If the plugin is distributed in binary format, it'll be required.

@finger563 After looking at your PR, I see that you added a script, maybe we can collaborate on my PR?

bcoconni:
> @bcoconni I've an M2 mac and I wondering if it's a good idea to compile the lib with universal support or not. Today the lib is compiled for x86_64 or arm64. We can compile for both as an universal package. Is it relevant or not? If the plugin is distributed in binary format, it'll be required.

@demonixis, yes it'd be better releasing the dynamic library as a universal binary. This is what we are already doing for the Python packages for MacOSX so that they can be used on both Intel and M1/M2.

finger563:
This PR is being closed in favor of #873

PR #952, opened 2023-08-26
Maintenance / Minor changes

bcoconni:
This PR brings a number of minor changes:
* Fixes a number of typos.
* Fixes the list of supported versions of the Python interpreter in `README.md`.
* The last parameter of `FGInitialCondition::Load()` is renamed from `useStoredPath` to `useAircraftPath` which, I think, is more explicit.
* Adds a helper method `FGPropertyManager::Unbind()` that takes a `std::shared_ptr` argument. This avoids explicitly calling the method `std::shared_ptr::get()` when using `Unbind()` i.e. this is syntactic sugar :smiley:
* Fixes the function `FindDifferences` in `JSBSim_utils.py` as the `DataFrames.lookup` method [is now obsolete in the library `pandas` 2.0](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#removal-of-prior-version-deprecations-changes).
* The sleep time of the method `getOutput` has been increased from 0.5s to 1.0s in the test case `TestInputSocket.py`. This is to avoid this test to randomly fail with MSVC when our CI workflow is executed on GitHub.

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -737,7 +737,7 @@ jobs:
       - name: Upload to Codecov
         uses: codecov/codecov-action@v3
         with:
-          working-directory: build/lcov/data/capture
+          directory: build/lcov/data/capture
           files: all_targets.info
 
   Rolling-Release:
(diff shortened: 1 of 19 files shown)


seanmcleod:
All looks good to me.

bcoconni:
Good !

I've also added some further changes to the unit tests:
* The method `GetGamma()` has been removed from the class `DummyAtmosphere` in the test `FGAtmosphereTest` as it was a duplication of `FGAtmosphere::SHRatio`.
* The test `FGauxiliaryTest` now calls `FGPropertyManager::Unbind()` to avoid having messages about properties already being tied.
  * The class name in `FGAuxiliaryTest` has also been renamed `FGAuxiliaryTest` instead of `FGAtmosphereTest` (the code was obviously copied from `FGAtmosphereTest` 😉).

Issue #989, opened 2023-11-20
Linearization and simplex trim from console executable and xml script


vtpasquale:
**I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**

- Previous versions of JSBSim appear to have supported linearization and simplex trim from the console executable and xml scripts.

- For example, this current xml script attempts to access both features in the same run: scripts/[737_cruise_steady_turn_simplex.xml](https://github.com/JSBSim-Team/jsbsim/blob/master/scripts/737_cruise_steady_turn_simplex.xml) but fails because the xml interface to the features has been discontinued. 

- The script interface to xml was removed after this version of [FGFDMExec.cpp](https://github.com/JSBSim-Team/jsbsim/blob/3b44b79e817a85966aedf800e8c185217a7c1ade/src/FGFDMExec.cpp); comments suggest that the features were not working at the time of removal. 

- JSBSim linearization features have since been resurrected (#227, #370) and exposed through the Python interface. I'm unsure if simplex trim features have been resurrected; the corresponding classes are still in the source. 

- I attempted to fork JSBSim and borrow a few lines of code from previous versions of FGFDMExec.cpp and FGFDMExec.h to bring the functionality back, but I was unsuccessful. I imagine this might be easy for an experienced JSBSim developer.

**What is the current behavior?**

When running:
```
JSBSim --script=737_cruise_steady_turn_simplex
```

The run fails with the following console error:
```
FATAL ERROR: JSBSim terminated with an exception.
The message was: No property, "simulation/do_simplex_trim" is defined.
```

Simple trim can be used instead of simplex trim by changing
`<set name="simulation/do_simplex_trim" value="1"/>`
to
`<set name="simulation/do_simple_trim" value="1"/>`
in
`737_cruise_steady_turn_simplex.xml`.
 
In that case, when running
```
JSBSim --script=737_cruise_steady_turn_simplex
```
The run fails with the following console error:
```
FATAL ERROR: JSBSim terminated with an exception.
The message was: No property, "simulation/do_linearization" is defined.
```

**What is the expected behavior?**

- Linearization: FGLinearization::WriteScicoslab to write the matrices to a file or to the console.
- Simplex trim to work like simple trim.

**What is the motivation / use case for changing the behavior?**
- Expose the features from console executable and xml script.

**Please tell us about your environment:**

- Windows and Linux
[...]

PR #991, opened 2023-11-22
Linearization from console executable and xml script 

vtpasquale:
This pull request corresponds to issue #989. Enables linearization from console executable and xml script; this workflow was available in previous JSBSim versions. The example script shows how to use the feature with the existing 737 aircraft model.

I think we can ultimately do better than this workaround:
https://github.com/JSBSim-Team/jsbsim/issues/989#issuecomment-1819929733


agodemar:
@bcoconni is there a problem in this check?

C/C++ build / C/C++ build (MSVC) (OFF) (pull_request)

seanmcleod:
There was a single test that failed, the `TestInputSocket` test.

bcoconni:
This test randomly fails. Nothing to worry about.
Simply clicking "Re-run failed tests" fixed the problem.

PR #993, opened 2023-12-02
Use cached PropertyNodes in python API for property reads

seanmcleod:
Based on performance tests in https://github.com/JSBSim-Team/jsbsim/discussions/977 implement the PropertyNode caching in the Python API for property reads.

### DIFF
--- a/python/jsbsim.pyx.in
+++ b/python/jsbsim.pyx.in
@@ -701,6 +701,9 @@ cdef class FGFDMExec(FGJSBBase):
         self.set_aircraft_path("aircraft")
         self.set_systems_path("systems")
 
+        # Dictionary cache of property nodes
+        self.properties = { }
+
     def __dealloc__(self) -> None:
         del self.thisptr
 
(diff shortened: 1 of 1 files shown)


bcoconni (commented):
So Python key hashing is faster than `SGPropertyNode` name lookup ? :smile: 

Nice patch :+1:, just a tiny comment: I think the dictionary should have a more explicit name such as `properties_cache`.

bcoconni:
Yes the code of `__getitem__` was implemented before `FGPropertyNode` was included in the Python binding. At that time, the problem was that `get_property_value` always returns a value whether the property exists or not. So to inform users if they were requesting the value of a property that did not exist, the existence was first tested via `has_node` prior to calling `get_property_value`.

Indeed, now that `FGPropertyNode` is available the sequence of testing the existence of the property then returning the value of that property is now feasible with a single name lookup, which is faster. And it is even faster with a cache. Your patch hits 2 birds with a stone 🐦 👍

seanmcleod:
Ah, was wondering why it was coded that way.

I'm used to the more aggressive form, "to kill two birds with one stone" :wink:

bcoconni:
> I'm used to the more aggressive form, "to kill two birds with one stone" :wink:

Well, I'll try to pretend my error was made on purpose because aviation as a whole needs to be friendlier to the environment 😄

PR #1030, opened 2024-02-03
Run tests on M1 MacOS runners.

bcoconni:
Until now we were only testing that JSBSim was successfully compiling on Apple M1 chips using cross-compilation on Intel MacOS GitHub runners. The resulting binaries were not tested on the actual hardware so we were happily assuming that if it compiles then it runs swiftly. And everyone knows how reliable that criterion is :smile:

These dark ages are now over: [GitHub just introduced M1 MacOS runners](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source), so this PR implements the testing of JSBSim on that platform.

Note that the floating point precision requirement had to be slightly relaxed at some places in the unit tests ...despite using the flag `-fno-fast-math`. If someone who has access to an M1 Macbook or equivalent knows better then by all means submit fixes :+1:

Note also that the issue that was fixed by PR #143 (random failure of `TestActuator.py`) is resurfacing with M1 MacOS runners. This has been fixed again by extending further the waiting time of `test_regression_bug_1503`:
https://github.com/JSBSim-Team/jsbsim/blob/6f0c4d805c499cf18280bcf8d7af358aee494876/tests/TestActuator.py#L94-L130

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -430,18 +430,21 @@ jobs:
   MacOSX:
     name: C/C++ build (MacOSX)
     needs: [XML-validation, MSIS-validation]
-    runs-on: macos-latest
     strategy:
       fail-fast: false
       matrix:
-        arch: [x86_64, arm64]
+        os: [macos-12]
+        python-version: [3.8]
+        include:
+          - os: macos-14
+            python-version: '3.10'
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


bcoconni:
And in case someone asks: the test job on M1 chips uses Python 3.10 (instead of Python 3.8 for other platforms) because Python<=3.9 is not available on the M1 MacOS runners.

Issue #1037, opened 2024-02-11
Fixing the mixture for JSBSim piston aircraft


hbeni:
Hi,
David Megginson found (and fixed) a bug in the mixture control.

See: https://forum.flightgear.org/viewtopic.php?f=49&t=42055
> Over on the flightgear-devel mailing list, I've been discussing a patch to fix the mixture input for JSBSim piston-engine aircraft.
> 
> Right now, the mixture control in src/FDM/JSBSim/models/propulsion/FGPiston.cpp is incorrectly clamped so that 1.0 (full rich) sets the mixture to peak power efficiency. In real life, the mixture control continues a ways rich of peak power to avoid predetonation at high power/low density altitude, which is why you have to lean a little for takeoff above 3,000 ft DA (or so). Anyone who's flown a simple trainer like a Cessna 172 or Piper PA-28 also knows how the RPM will initially increase as you start leaning, before it decreases again, which doesn't happen with our current JSBSim piston aircraft.

-----
```diff
diff --git a/src/FDM/JSBSim/models/propulsion/FGPiston.cpp b/src/FDM/JSBSim/models/propulsion/FGPiston.cpp
index 1e73a503d..c1c75bd57 100644
--- a/src/FDM/JSBSim/models/propulsion/FGPiston.cpp
+++ b/src/FDM/JSBSim/models/propulsion/FGPiston.cpp
@@ -752,7 +752,8 @@ void FGPiston::doAirFlow(void)
 
 void FGPiston::doFuelFlow(void)
 {
-  double thi_sea_level = 1.3 * in.MixturePos[EngineNumber]; // Allows an AFR of infinity:1 to 11.3075:1
+  // use square root to increase sensitivity in the higher end
+  double thi_sea_level = 1.65 * sqrt(in.MixturePos[EngineNumber]); // Allows an AFR of infinity:1 to (no longer accurate) 11.3075:1
   equivalence_ratio = thi_sea_level * 101325.0 / p_amb;
   m_dot_fuel = (m_dot_air * equivalence_ratio) / 14.7;
   FuelFlowRate =  m_dot_fuel * 2.2046;  // kg to lb
```

bcoconni:
Thanks @hbeni for submitting this issue.

[...]

Issue #922, opened 2023-06-06
CMAKE Configure Error on MacOS (M1 aarch64)


cryptik:
***

**I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**

When running the cmake step on jsbsim build, the configure step appears to generate all of the necessary files; however, it generates an error relative to a missing class function (`JSBSim::FGPropertyManager::GetName`) 

**What is the current behavior?**

Running the CMAKE command as follows:
`cmake -DCMAKE_CXX_FLAGS_RELEASE="-O3 -march=native -mtune=native" -DCMAKE_C_FLAGS_RELEASE="-O3 -march=native -mtune=native" -DCMAKE_BUILD_TYPE=Release ..`
results in the following errors:
```
-- Found Cython: /opt/homebrew/Caskroom/miniconda/base/envs/ardu/bin/cython  
Traceback (most recent call last):
  File "/Users/cryptik/Development/other/jsbsim/build/python/doxy2pydocs.py", line 201, in <module>
    raise IOError(
OSError: File classJSBSim_1_1FGPropertyManager.xml does not contain JSBSim::FGPropertyManager::GetName
-- Found CxxTest: /opt/homebrew/include  
CMake Warning at CMakeLists.txt:149 (message):
  -- Could NOT find LSB_Release (missing: LSB_RELEASE_EXECUTABLE)
-- Configuring done (42.9s)
-- Generating done (0.1s)
-- Build files have been written to: /Users/cryptik/Development/other/jsbsim/build

```

**What is the expected behavior?**

No OSError should be generated on build

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

OS - macOS (Ventura, M1 arm)

**Other information**

bcoconni:
As a temporary fix, you should be able to work around this issue by disabling the generation of documentation. For that purpose, you need to `-DBUILD_DOCS=OFF` to the CMake command:
```bash
cmake -DCMAKE_CXX_FLAGS_RELEASE="-O3 -march=native -mtune=native" -DCMAKE_C_FLAGS_RELEASE="-O3 -march=native -mtune=native" -DBUILD_DOCS=OFF -DCMAKE_BUILD_TYPE=Release ..
```

bcoconni:
Sorry @cryptik, I had lost sight of this ticket.

[...]

PR #1048, opened 2024-03-02
Documentation updates.

bcoconni:
This PR addresses 2 issues related to documentation:
* Issue #922 where the build process crashes if the Python documentation generation is triggered.
* Issue #1037 where the documentation is misleading in claiming that mixture position and command shall be constrained within the range from 0.0 to 1.0

Regarding the first issue, the problem is due to the lack of Doxygen documentation in `src/simgear/props/props.hxx` which makes `python/Doxy2PyDocStrings.py` crash. The problem has been fixed by including the documentation of `FGPropertyNode.get_double_value()` and `FGPropertyNode.get_double_value()` directly in `python/JSBSim.pyx.in`.

In the process, `FGPropertyNode.set_double_value()` has been fixed to behave like its C++ counterpart (returns a boolean to indicate if the update of the property value has succeeded).

### DIFF
--- a/python/jsbsim.pxd
+++ b/python/jsbsim.pxd
@@ -76,7 +76,7 @@ cdef extern from "input_output/FGPropertyManager.h" namespace "JSBSim":
         const string& GetName() const
         const string& GetFullyQualifiedName() const
         double getDoubleValue() const
-        void setDoubleValue(double value)
+        bool setDoubleValue(double value)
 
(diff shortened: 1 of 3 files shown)


davidmegginson:
Since all of the other controls are normalised between 0.0 and 1.0, it seems strange that the mixture wouldn't be, but understood why that's necessary to work around an issue in one of the FDMs.

PR #1022, opened 2024-01-20
fix: target include directories

carlocorradini:
Fix #1021

PR #1050, opened 2024-03-03
Maintenance / Minor changes

bcoconni:
This PR brings a number of minor changes and fixes:
* In the CI workflow, `MinGW32` has been renamed to `MinGW` as MinGW is no longer limited to 32 bits [as stated on their website](https://www.mingw-w64.org).
* Fix the codecov job which was broken since PR #1032: the coverage data were [no longer uploaded to codecov.io](https://app.codecov.io/github/JSBSim-Team/jsbsim/pulls).
* Extend the usage of CMake `target_include_directories`, following the lead of PR #1022 (there are still a few places that use `include_directories` and this will be addressed in a later PR).
* Make the unit test `test_inherit_with_extra_args` more robust. Indeed [there are some subtleties between `__cinit__` and `__init__`](https://stackoverflow.com/questions/62827538/in-cython-class-whats-the-difference-of-using-init-and-cinit):
  * The test now checks that the parameters `root_dir` and `pm_root` are properly handled.
  * Since the class `FGFDMExec` has no constructor `__init__`, there is no need to call it.


bcoconni:
The codecov report is back, confirming that this PR fixes the coverage job in `cpp-python-build.yml` :smile:

PR #1077, opened 2024-04-26
CI workflow fixes

bcoconni:
This PR fixes a number of issues that have crept in our CI workflow:

* GitHub's `macos-latest` is now set to `macos-14` which is a runner with M1 chips but:
  * Julia is no longer installed by default on these runners (most likely due to the fact that [Julia on M1 chips is untested according to the Julia community itself](https://github.com/julia-actions/setup-julia)).
  * We are limited to Matlab R2022a since the SimuLink file [ex737cruise.slx](https://github.com/JSBSim-Team/jsbsim/blob/master/matlab/ex737cruise.slx) is only compatible with R2022a and [Apple Silicon chips are only supported since Matlab R2023b](https://www.mathworks.com/matlabcentral/answers/641925-is-matlab-supported-on-apple-silicon-macs).
* Took the opportunity to remove `macos-11` as [it has been deprecated and will no longer be supported after June 28, 2024](https://docs.github.com/fr/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-private-repositories).
* [GitHub runners now provide 4 CPUs for open source projects !](https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/) Our CI workflow now automatically picks the number of cores to take the opportunity of this speed increase and to save ourselves the trouble of updating the number of cores used by parallel jobs (mostly compiler & test jobs):
  * Using the variable `$(nproc)` on Linux
  * Using the environment variable `%NUMBER_OF_PROCESSORS%` (or `$Env:NUMBER_OF_PROCESSORS` in PowerShell) on Windows
  * Calling `$(sysctl -n hw.logicalcpu)` on MacOSX.
* The coverage job has been updated:
  * Automatic search of coverage data by the Codecov script is now disabled so that only the relevant coverage data is uploaded to Codecov.
  * CMake-codecov : The PR RWTH-HPC/CMake-codecov#40 has been pushed to JSBSim to support lcov 2.0 (which is the default version on my Linux PC since I've upgraded it to Fedora 39).

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -737,6 +737,7 @@ jobs:
         with:
           file: ./build/lcov/data/capture/all_targets.info
           token: ${{ secrets.CODECOV_TOKEN }}
+          disable_search: true
           fail_ci_if_error: true
 
   Rolling-Release:
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -169,7 +169,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ macos-latest, ubuntu-22.04, windows-latest ]
+        os: [ macos-12, ubuntu-22.04, windows-latest ]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout JSBSim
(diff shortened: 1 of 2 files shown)


bcoconni:
The substantial decrease in coverage from 41.37% to 24.97% is an anticipated result of this PR. Indeed, PR #1050 has caused a regression by including the tests themselves in the overall score, which is nonsensical.

PR #1081, opened 2024-05-02
Some improvements to the socket code.

bcoconni:
While investigating the "socket" code to address the topics discussed in #1079, I have found a number of small details that could be improved. I don't think these will resolve the discussion #1079 but I'm pushing this PR nonetheless for the record.

### Minor changes / documentation
* The first thing that stroke me is that the 2 constructors of `FGfdmSocket` have actually very different purposes as one constructor is returning a socket for a **client connection** while the other one returns a socket for a **server connection**. To document that, I have added comments to `FGfdmSocket.h`. Note that the documentation has been generated by an AI (*GitHub Copilot*) and reviewed by a non native English speaker (myself) so brace yourself before reading 😄
* I've moved `connected=true` to a location that makes its purpose more obvious as the previous location of this line of code was misleadingly suggesting that `connected=true` was indicating that the call to `accept()` had succeeded.
* In `FGInputSocket.cpp`, when the command `quit`is issued to JSBSim, the method `FGfdmSocket::Send()` is now called to return "*Closing connection*" instead of calling `FGfdmSocket::Reply()`. This avoids displaying the prompt `JSBSim>` just after the message "*Connection closed*" which looks weird to me.

### Improvements to the error management
* If the calls to `bind()` or `listen()` fail, `FGfdmSocket` now closes the socket `sckt` and sets it to `INVALID_SOCKET` which is a good practice in order to release unused file descriptors to the OS (see StackOverflow [Should I call close() on bind() or listen() error](https://stackoverflow.com/questions/48400232/should-i-call-close-on-bind-or-listen-error))
* Following the same idea, the destructor `~FGfdmSocket()` shutdowns `sckt_in` and closes `sckt` (in that order otherwise it crashes according to my tests) to release the file descriptors to the OS.
* The call to `recv` in `FGfdmSocket::Receive()` is now enclosed in an `if (Protocol == ptTCP) { ... }` as calling `recv` is meaningless for UDP sockets.
* It is now checked that `sckt != INVALID_SOCKET` before calling `send` in `FGfdmSocket::Send()`. Also `send` is now using `sckt_in` if the socket is using the TCP protocol. Indeed this method is called from the class `FGOutputSocket` which can define a socket with either protocol UDP or TCP.
* It is now checked that `sckt != INVALID_SOCKET` before calling `select` in `FGfdmSocket::WaitUntilReadable()`. Also I have added an assert to check that `Protocol == ptTCP` so that we are sure that this method is never called for a UDP socket. At the moment, `FGfdmSocket::WaitUntilReadable()` is only called by the class `FGInputSocket` which is guaranteed to be a TCP socket so the assert passes.

### DIFF
--- a/src/input_output/FGInputSocket.cpp
+++ b/src/input_output/FGInputSocket.cpp
@@ -243,7 +243,7 @@ void FGInputSocket::Read(bool Holding)
       } else if (command == "quit") {               // QUIT
 
         // close the socket connection
-        socket->Reply("Closing connection\r\n");
+        socket->Send("Closing connection\r\n");
         socket->Close();
 
       } else if (command == "info") {               // INFO
--- a/src/input_output/FGfdmSocket.cpp
+++ b/src/input_output/FGfdmSocket.cpp
@@ -270,38 +270,31 @@ string FGfdmSocket::Receive(void)
         fcntl(sckt_in, F_SETFL, flags | O_NONBLOCK);
 #endif
         if (send(sckt_in, "Connected to JSBSim server\n\rJSBSim> ", 36, 0) == SOCKET_ERROR)
-          PrintSocketError("Receive - TCP connection acknowledgement");
+          LogSocketError("Receive - TCP connection acknowledgement");
       }
     }
 
     if (sckt_in != INVALID_SOCKET) {
-    int num_chars;
+      int num_chars;
 
-    while ((num_chars = recv(sckt_in, buf, sizeof buf, 0)) > 0) {
-      data.append(buf, num_chars);
-    }
+      while ((num_chars = recv(sckt_in, buf, sizeof buf, 0)) > 0)
+        data.append(buf, num_chars);
 
-    if (num_chars == SOCKET_ERROR) {
+      if (num_chars == SOCKET_ERROR || num_chars == 0) {
 #ifdef _WIN32
-      if (WSAGetLastError() != WSAEWOULDBLOCK)
+        if (WSAGetLastError() != WSAEWOULDBLOCK)
 #else
-      if (errno != EWOULDBLOCK)
+        if (errno != EWOULDBLOCK)
 #endif
-        PrintSocketError("Receive - TCP data reception");
-    }
-
-#ifdef _WIN32
-      // when nothing received and the error isn't "would block"
-      // then assume that the client has closed the socket.
-      if (num_chars == 0) {
-        DWORD err = WSAGetLastError();
-        if (err != WSAEWOULDBLOCK) {
+        {
+          LogSocketError("Receive - TCP data reception");
+          // when nothing received and the error isn't "would block"
+          // then assume that the client has closed the socket.
           cout << "Socket Closed. Back to listening" << endl;
           closesocket(sckt_in);
           sckt_in = INVALID_SOCKET;
         }
       }
-#endif
     }
   }
 
@@ -311,7 +304,14 @@ string FGfdmSocket::Receive(void)
     socklen_t fromlen = sizeof addr;
     int num_chars = recvfrom(sckt, buf, sizeof buf, 0, (struct sockaddr*)&addr, &fromlen);
     if (num_chars > 0) data.append(buf, num_chars);
-    if (num_chars == SOCKET_ERROR) PrintSocketError("Receive - UDP data reception");
+    if (num_chars == SOCKET_ERROR) {
+#ifdef _WIN32
+      if (WSAGetLastError() != WSAEWOULDBLOCK)
+#else
+      if (errno != EWOULDBLOCK)
+#endif
+        LogSocketError("Receive - UDP data reception");
+    }
   }
 
   return data;
@@ -322,11 +322,12 @@ string FGfdmSocket::Receive(void)
 int FGfdmSocket::Reply(const string& text)
 {
   int num_chars_sent=0;
+  assert(Protocol == ptTCP);
 
   if (sckt_in != INVALID_SOCKET) {
     num_chars_sent = send(sckt_in, text.c_str(), text.size(), 0);
-    if (num_chars_sent == SOCKET_ERROR) PrintSocketError("Reply");
-    send(sckt_in, "JSBSim> ", 8, 0);
+    if (num_chars_sent == SOCKET_ERROR) LogSocketError("Reply - Send data");
+    if (send(sckt_in, "JSBSim> ", 8, 0) == SOCKET_ERROR) LogSocketError("Reply - Prompt");
   } else {
     cerr << "Socket reply must be to a valid socket" << endl;
     return -1;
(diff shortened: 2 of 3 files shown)


seanmcleod:
Yep, I've found the socket code over the years when I've needed to take a look at things always a bit confusing in terms of the same class handling client versus server plus tcp and udp all in one.

Will take a look at the commits.

> first thing that stroke me

You did mention not being native English speaker 😉 Should be "...that struck me..", like being struck by a bolt of lightning as opposed to stroking a cat.

jonsberndt:
I have to confess that when I wrote the initial implementation of the telnet interface I kind of made it up as I went. It was the first time I wrote any socket code. Feels like a hundred years ago. 😊 

 

Jon

bcoconni:
> Yep, I've found the socket code over the years when I've needed to take a look at things always a bit confusing in terms of the same class handling client versus server plus tcp and udp all in one.

Indeed but I'm wondering which would be the more painful ? A *Swiss army knife* class such as `FGfdmSocket`, a collection of independent classes with lots of code duplication or a pure object-oriented class hierarchy which avoids code duplication but is difficult to grasp ?

> > first thing that stroke me
>
> You did mention not being a native English speaker 😉 Should be "...that struck me..", like being struck by a bolt of lightning as opposed to stroking a cat.

Haha :laughing: You've been warned !
Well, I confirm that these server/client constructors struck me but did not stroke me :smile:

bcoconni:
> I have to confess that when I wrote the initial implementation of the telnet interface I kind of made it up as I went. It was the first time I wrote any socket code. Feels like a hundred years ago. 😊

I guess that's where we all are. I'm no socket/network code specialist either and I learn as I go. But note that the current code still looks very much like the code you wrote many years ago, that's not so bad for "made up" code !

bcoconni:
I have just pushed another commit to this PR:
* Code legibility improvement: new `#defines` for `closesocket` and `SD_BOTH` have been added to remove a number of `#ifdef _WIN32`.
* Error management: the value returned by `select` is now tested for errors. Also the call to `select()` is now using a syntax that should be compatible with both Windows and BSD/Unix platforms. In particular, the first parameter of `select` has been set to `FD_SETSIZE`:
  * This first parameter is ignored by Windows anyway (see their [documentation about select](https://learn.microsoft.com/fr-fr/windows/win32/api/winsock2/nf-winsock2-select))
  * For BSD/Unix platforms, `FD_SETSIZE` is the maximum number of socket descriptors which avoid weird situation where `sckt_in == FD_SETSIZE` and `sckt_in+1>FD_SETSIZE`. In addition, this makes this piece of code more robust as it no longer depends on the value of `sckt_in`.

bcoconni:
Another commit to the PR to check the returned of socket functions and print the error message when `SOCKET_ERROR` is returned.

bcoconni:
Following my previous commit, I now see that Windows is returning lots of errors "Socket error: A non-blocking socket operation could not be completed immediately" in `TestInputSocket.py`. Similarly, Linux is returning lots of "Socket error: Resource temporarily unavailable". I guess these are due to the fact that telnet and JSBSim are running synchronously on the same thread in `TestInputSocket.py` so the telnet socket is not responding immediatly when JSBSim is interrogating it.

In addition, the messages lack context (where the error message has been issued from ?) and are flooding the console. So I have pushed one last commit to add context to the messages and print them only if `debug_lvl` is strictly positive.

seanmcleod:
In terms of tons of "error messages" now, I think this particular case of printing/logging an error is misplaced and shouldn't be there.

https://github.com/JSBSim-Team/jsbsim/blob/84768c7dd161e3d79e0c7b7e6e6c97034c21f168/src/input_output/FGfdmSocket.cpp#L280-L284

So, in this case since we've set the socket for the telnet connection to be non-blocking, the return value of `SOCKET_ERROR/-1` isn't necessarily an error, particularly when the error code/reason is `Would Block`. The OS is simply telling us in this case that there is no waiting data from the telnet client.

And in particular if the FDM is running at 120Hz and you have say a human sending intermittent telnet commands there are going to be thousands of these sorts of "error messages" logged in just say a 10s period when the human isn't sending a telnet command.

So I would change this code to only log an error message along these lines (having to still deal with Windows vs Unix last error checking differences):

```c++
if (num_chars == SOCKET_ERROR && WSAGetLastError() != WSAEWOULDBLOCK)
   PrintSocketError("Receive - TCP data reception");
```
```

**Comment by bcoconni** [comment]
@seanmcleod Good point ! I have amended the code according to your feedback.

**Comment by seanmcleod** [comment]
Yep, looks good.

One more thing :wink:

As I mentioned in the original discussion - https://github.com/JSBSim-Team/jsbsim/discussions/1079 I'm pretty sure that the following block of code, although it can also be simplified, needs to be executed for Unix systems as well.

https://github.com/JSBSim-Team/jsbsim/blob/7590e8d31ea9c6e95bf57b8165dbf37d100c6b84/src/input_output/FGfdmSocket.cpp#L293-L304 

In other words, in the case that `num_chars == 0` we need to perform the socket shutdown for all operating systems.

So I'd suggest replacing that `#ifdef` block with.

```c++
if (num_chars == 0) {
  PrintSocketError("Receive - TCP data reception- Socket Closed. Back to listening"); 
  closesocket(sckt_in); 
  sckt_in = INVALID_SOCKET;   
}
```

This would fix the issue that the user reported when running JSBSim on Mac (Unix based), i.e. killing the telnet connection with ctrl-c or equivalent as opposed to sending `quit` via the telnet connection resulted in her not being able to re-establish a telnet connection.

Actually, thinking about it, we probably want to perform this also in the case of `-1 (SOCKET_ERROR)` being returned when last error isn't `WOULDBLOCK`.

seanmcleod:
It seems there is always one more thing with regards to the socket code :wink:

So in the UDP case, in the same `Receive()` method, note that the UDP socket has also been set to be non-blocking in the ctor. Which means we can expect to get a lot of returns on line 314 with `num_chars == SOCKET_ERROR` that aren't really errors to log. We need similar logic that we have above for the TCP case.

https://github.com/JSBSim-Team/jsbsim/blob/7590e8d31ea9c6e95bf57b8165dbf37d100c6b84/src/input_output/FGfdmSocket.cpp#L308-L315

bcoconni:
@seanmcleod Just pushed another commit that implements your suggestions (both of them).

I took this opportunity to add a check of the value returned by `send` when sending the prompt `JSBSim>`:
https://github.com/JSBSim-Team/jsbsim/blob/92ebccaf993032faada7d09754383e3d858fd67f/src/input_output/FGfdmSocket.cpp#L306
I've also added some additional asserts in `FGfdmSocket::Close()` and `FGfdmSocket::Reply()` as both of these methods are using `sckt_in` which is only initialized for TCP sockets.

seanmcleod:
@bcoconni it looks good.

bcoconni:
Thanks for the detailed review, @seanmcleod 
The PR is now merged.

PR #1091, opened 2024-05-18
Fix inconsistent property name (sea-level-radius-ft)

cbirkhold:
In all other instances the unit appendix is separated by a dash rather than underscore.

Should the property be published under the inconsistent name also to maintain compatibility with existing code?

### DIFF
--- a/src/models/FGInertial.cpp
+++ b/src/models/FGInertial.cpp
@@ -255,7 +255,7 @@ void FGInertial::SetGravityType(int gt)
 
 void FGInertial::bind(void)
 {
-  PropertyManager->Tie("inertial/sea-level-radius_ft", &in.Position,
+  PropertyManager->Tie("inertial/sea-level-radius-ft", &in.Position,
                        &FGLocation::GetSeaLevelRadius);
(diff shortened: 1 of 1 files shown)


bcoconni:
This property is spelled that way since 2011 (see commit 3c05fee1b3a726a0b10c7584f0afa8f5c25be17c). Changing its spelling would break all flight dynamics models (FDMs) that make use of it, such as one of our own:
https://github.com/JSBSim-Team/jsbsim/blob/56c666265ac1db3216cc76c7980a87ec424a6546/aircraft/J246/Systems/J246GuidanceExecutive.xml#L230
https://github.com/JSBSim-Team/jsbsim/blob/56c666265ac1db3216cc76c7980a87ec424a6546/aircraft/J246/Systems/J246GuidanceExecutive.xml#L248
Generally speaking, there are a lot of inconsistencies in property naming in JSBSim, starting with `lbs` that is used to designate either mass pounds or pound force. Fixing/modifying these would break a lot of FDMs so we have to live with the current state of property naming.

Sorry but this PR is rejected.

cbirkhold:
Understood. Many thanks for the feedback/context!

PR #1092, opened 2024-05-18
Reduce console output when debug level is zero

cbirkhold:
Throughout the codebase console output is generally only happening when the debug level is greater than zero. That is a useful feature and this change checks the debug level for additional cases.


bcoconni:
PR merged. Thanks for your contribution @cbirkhold

PR #1093, opened 2024-05-18
Add support for setting ic/vw-mag-fps

seanmcleod:
Based on the issue reported in discussion - https://github.com/JSBSim-Team/jsbsim/discussions/1088

### DIFF
--- a/src/initialization/FGInitialCondition.cpp
+++ b/src/initialization/FGInitialCondition.cpp
@@ -590,11 +590,6 @@ void FGInitialCondition::SetWindDownKtsIC(double wD)
   calcAeroAngles(_vt_NED);
 }
 
-void FGInitialCondition::SetWindMagFpsIC(double mag)
-{
-  SetWindMagKtsIC(mag*fpstokts);
-}
-
 //******************************************************************************
 // Modifies the wind velocity (in knots) while keeping its direction unchanged.
 // The vertical component (in local NED frame) is unmodified. The aircraft
@@ -941,7 +936,7 @@ FGColumnVector3 FGInitialCondition::GetWindNEDFpsIC(void) const {
 
 //******************************************************************************
 
-double FGInitialCondition::GetWindFpsIC(void) const
+double FGInitialCondition::GetWindMagFpsIC(void) const
 {
   const FGMatrix33& Tb2l = orientation.GetTInv();
   FGColumnVector3 _vt_NED = Tb2l * Tw2b * FGColumnVector3(vt, 0., 0.);
@@ -1524,7 +1519,7 @@ void FGInitialCondition::bind(FGPropertyManager* PropertyManager)
   PropertyManager->Tie("ic/vw-down-fps", this,
                        &FGInitialCondition::GetWindDFpsIC);
   PropertyManager->Tie("ic/vw-mag-fps", this,
-                       &FGInitialCondition::GetWindFpsIC,
+                       &FGInitialCondition::GetWindMagFpsIC,
                        &FGInitialCondition::SetWindMagFpsIC);
   PropertyManager->Tie("ic/vw-dir-deg", this,
                        &FGInitialCondition::GetWindDirDegIC,
--- a/src/initialization/FGInitialCondition.cpp
+++ b/src/initialization/FGInitialCondition.cpp
@@ -590,6 +590,11 @@ void FGInitialCondition::SetWindDownKtsIC(double wD)
   calcAeroAngles(_vt_NED);
 }
 
+void FGInitialCondition::SetWindMagFpsIC(double mag)
+{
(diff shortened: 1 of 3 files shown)


src/initialization/FGInitialCondition.cpp
```diff
@@ -590,6 +590,11 @@ void FGInitialCondition::SetWindDownKtsIC(double wD)
   calcAeroAngles(_vt_NED);
 }
 
+void FGInitialCondition::SetWindMagFpsIC(double mag)
+{
+  SetWindMagKtsIC(mag*fpstokts);
+}
+
```
bcoconni, line 597: Since the code is trivial, I'd rather move that definition into the header so that the C++ compiler can inline it when possible (see my other comment below):
```suggestion
```

src/initialization/FGInitialCondition.cpp
```diff
@@ -1519,7 +1524,8 @@ void FGInitialCondition::bind(FGPropertyManager* PropertyManager)
   PropertyManager->Tie("ic/vw-down-fps", this,
                        &FGInitialCondition::GetWindDFpsIC);
   PropertyManager->Tie("ic/vw-mag-fps", this,
-                       &FGInitialCondition::GetWindFpsIC);
+                       &FGInitialCondition::GetWindFpsIC,
+                       &FGInitialCondition::SetWindMagFpsIC);
```
bcoconni, line 1528: The getter and setter have dissimilar names: should be `SetWindFpsIC()` ?

src/initialization/FGInitialCondition.h
```diff
@@ -443,6 +443,10 @@ class JSBSIM_API FGInitialCondition : public FGJSBBase
       @param wD Initial wind velocity in local down direction, feet/second   */
   void SetWindNEDFpsIC(double wN, double wE, double wD);
 
+  /** Sets the initial total wind speed.
+      @param mag Initial wind velocity magnitude in feet/second */
+  void SetWindMagFpsIC(double mag);
```
bcoconni, line 448: As per my comment above.
```suggestion
  void SetWindMagFpsIC(double mag) { SetWindMagKtsIc(mag*fpstokts); }
```

bcoconni (changes requested):
A couple of minor updates could be made and it would be good if the unit test could be updated to test the new setter:
https://github.com/JSBSim-Team/jsbsim/blob/56c666265ac1db3216cc76c7980a87ec424a6546/tests/unit_tests/FGInitialConditionTest.h#L347-L372

src/initialization/FGInitialCondition.cpp
seanmcleod: Will do, although C++ compilers/linkers have been able to inline methods in .cpp files for many years :wink:

seanmcleod: I'll add an additional test to the unit test.

src/initialization/FGInitialCondition.cpp
seanmcleod: Yep, did wonder about that a bit. Matched the new setter to match the name of the existing setter that is identical except for having `Kts` in the name.

Looking at it again I'm actually inclined to rename the getter from `GetWindFpsIC()` to `GetWindMagFpsIC()` to make it clear it's returning the magnitude as opposed to a vector or some vector component.

src/initialization/FGInitialCondition.cpp
bcoconni: > Will do, although C++ compilers/linkers have been able to inline methods in .cpp files for many years 😉

OK I'm lagging many years behind :smile: but I've always had this idea that when the body of a method is given in the header then compilers would intepret that as an `inline` requirement (that they can happily ignore by the way). Anyway that doesn't hurt either way but having a one-liner body in the header saves the trouble of opening the `.cpp` file :wink:

src/initialization/FGInitialCondition.cpp
bcoconni: > Looking at it again I'm actually inclined to rename the getter from `GetWindFpsIC()` to `GetWindMagFpsIC()` to make it clear it's returning the magnitude as opposed to a vector or some vector component.

Looks good to me.

bcoconni:
@seanmcleod Tests are failing: it seems you've omitted changing `GetWindFpsIC()` to `GetWindMagFpsIC()` in some other unit tests.

seanmcleod:
Yep, I had done a search for `GetWindFpsIC()` when I first changed the method name, but that was only in JSBSim, and JSBSim compiled fine locally. I thought my search had been across the whole source tree so didn't pick up the references in the unit tests, which I didn't compile locally.

bcoconni (approved):

bcoconni:
All good :+1:

Issue #1118, opened 2024-07-03
Robocopy step errors if spaces in path.


jamesford42:
This post build copy step (or however it is hooked in) does not properly quote its paths. It appears to break when there is a space in the folder structure.
```
>C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(499,5): warning MSB8004: Output Directory does not end with a trailing slash.  This build instance will add the slash as it is required to allow proper evaluation of the Output Directory.
1>JSBSimForUnreal.vcxproj -> D:\Projects\Unreal Projects\JSBSim\UnrealEngine\Plugins\JSBSimFlightDynamicsModel\Source\ThirdParty\JSBSim\LibDebug\JSBSim.dll
1>A subdirectory or file UnrealEngine\Plugins\JSBSimFlightDynamicsModel\Source\ThirdParty\JSBSim\Include already exists.
1>
1>-------------------------------------------------------------------------------
1>   ROBOCOPY     ::     Robust File Copy for Windows
1>-------------------------------------------------------------------------------
1>
1>  Started : Wednesday, July 3, 2024 12:52:40 PM
1>   Source - D:\Projects\Unreal\
1>     Dest - D:\Projects\Unreal Projects\JSBSim\Projects\JSBSim\src\
1>
1>    Files :
1>  Options : /DCOPY:DA /COPY:DAT /R:1000000 /W:30
1>
1>------------------------------------------------------------------------------
1>
1>EXEC : error : Invalid Parameter #3 : "UnrealEngine\Plugins\JSBSimFlightDynamicsModel\Source\ThirdParty\JSBSim\Include"
1>Missing operator.
1>
1>       Simple Usage :: ROBOCOPY source destination /MIR
1>
1>             source :: Source Directory (drive:\path or \\server\share\path).
1>        destination :: Destination Dir  (drive:\path or \\server\share\path).
1>               /MIR :: Mirror a complete directory tree.
1>
1>    For more usage information run ROBOCOPY /?
```

jamesford42:
fixes #1118

bcoconni:
As the PR #1119 has been merged, I suggest that to close this issue.

PR #1119, opened 2024-07-03
fixing robocopy step

jamesford42:
paths needed to be quoted to function with spaces in directory 
the destination path was not prefixed with $(SolutionDir) 
added newlines so its actually readable in the ide

### DIFF
--- a/JSBSimForUnreal.vcxproj
+++ b/JSBSimForUnreal.vcxproj
@@ -86,19 +86,6 @@
       </DataExecutionPrevention>
       <AdditionalDependencies>ws2_32.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
-    <PostBuildEvent>
-      <Command>mkdir UnrealEngine\Plugins\JSBSimFlightDynamicsModel\Source\ThirdParty\JSBSim\Include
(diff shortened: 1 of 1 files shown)


jamesford42:
fixes #1118

bcoconni:
@gallonmate, @demonixis can you comment on this ?

seanmcleod:
Looks good to me, although I don't have Unreal installed so I haven't tested it myself.

Basically quoting the source and destination path arguments to robocopy and then it also looks like he's removed the duplication of this work from being duplicated in the project file for both debug and release builds and has it once applied for all build types?

gallonmate:
Should be okay, thanks!

> @gallonmate, @demonixis can you comment on this ?

bcoconni:
OK. The PR is merged.
Thanks @jamesford42 for the PR
Thanks @seanmcleod and @gallonmate for the review.

NexRX:
I tried openning the project in VS 2022 and im getting this
`JSBSimForUnreal.vcxproj(389,4): The project file could not be loaded. There are multiple root elements. Line 389, position 4.`

Im very new to Visual Studio and all that but is `<ItemDefinitionGroup>` meant to be in it's own XML element outside the Root/`<Project>` element. Perhaps this is some VS thing i don't understand but for me at least VS 2022 doesn't want to open this up because of this unless i move it within. 

Should I raise an issue or something else going on here?

seanmcleod:
I get the same error here with VS2022.

PR #1122, opened 2024-07-08
Trent/aero debug

TrentConley:


seanmcleod:
You shouldn't be doing your active development and debugging via a pull request. Once you have finished local development and debugging and you think it's ready for submission then generate a pull request.

PR #1123, opened 2024-07-13
Fixes and updates to the CI workflow

bcoconni:
This PR addresses a fix and a couple of updates to the CI workflow:
* Remove the space between `--` and `parallel` in the file `.github/workflows/cpp-python-build.xml` as it results in the failure of the tests run on the Windows/MSVC platform. For some reason, although this was a syntax error, it is only recently that this has prevented the CI workflow to run.
https://github.com/JSBSim-Team/jsbsim/blob/754a4aca640fb008d921853d009135bf99723610/.github/workflows/cpp-python-build.yml#L344
* Fixes the capitalization of `JSBSim` in the `.github/workflows/cpp-python-build.xml`. It must be all lower cases now (most likely a change in [setuptools](https://setuptools.pypa.io/en/latest/) ?)
https://github.com/JSBSim-Team/jsbsim/blob/754a4aca640fb008d921853d009135bf99723610/.github/workflows/cpp-python-build.yml#L624
* Update [cibuildwheel](https://github.com/pypa/cibuildwheel) to its latest revision i.e. 2.19
* Drop the action `crazy-max/ghaction-github-pages` and use the [GitHub pages workflow](https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages) instead.

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -895,6 +895,13 @@ jobs:
     needs: Stable-Release
     name: Deploy Documentation
     runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.pages_url }}
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
     steps:
       - name: Install Ubuntu packages
         run: |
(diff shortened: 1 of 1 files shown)


bcoconni:
The PR is immediately pulled to fix the space issue (first item in the PR description) that prevents the CI workflow to run on the Windows/MSVC platform.

PR #1136, opened 2024-08-10
Logging redesign (cont'd)

bcoconni:
This PR is the continuation of the PR #1094. The replacement of `cout` and `cerr` by `FGLogging` is propagated to some additional classes.

The class `FGLogging` has also had some changes:
* It skips calling `FGLogger::Message` if its buffer is empty. This is because we cannot assume that calling `FGLogger::Message` is a cheap operation.
* `FGLogging::operator<<` is extended to all number formats, `FGColumnVector3` and some formatting from `std::ostringstream` such as `std::setw` and `std::ios_base`. This is to ensure that `FGLogging` behaves the same way as `cout` and `cerr`.
  * There is a subtlety regarding `setw_t` as Windows considers `std::setw` and `std::setprecision` to be of the same type while Linux and MacOSX don't.
* `ERROR` has to be undef'ed in `FGfdmSocket.h` because there is a clash between `ERROR` being defined in the Windows headers and `ERROR` being used in our `LogLevel` enum:
https://github.com/JSBSim-Team/jsbsim/blob/b41fcddc505baf041a3082ac3be2312e392c7792/src/input_output/FGLog.h#L62-L69

Regarding the exception management:
* When an error will throw an exception, the error level is always set to **FATAL**.
* The message that is passed to the exception `JSBSim::BaseException` is a copy of the error message so that the same message is logged and send to the exception manager. As a result the code looks like:
```c++
FGLogging log(FDMExec->GetLogger(), LogLevel::FATAL);
log << "Something terrible happened !" << endl;
throw BaseException(log.str());
```

### DIFF
--- a/src/input_output/FGLog.cpp
+++ b/src/input_output/FGLog.cpp
@@ -48,18 +48,28 @@ CLASS IMPLEMENTATION
 
 void FGLogging::Flush(void)
 {
-  logger->Message(buffer.str());
-  buffer.str("");
-  logger->Format(LogFormat::RESET);
-  logger->Flush();
+  std::string message = buffer.str();
+
+  if (!message.empty()) {
+    logger->Message(message);
+    buffer.str("");
+    logger->Format(LogFormat::RESET);
+    logger->Flush();
+  }
+
   buffer.clear();
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 FGLogging& FGLogging::operator<<(LogFormat format) {
-  logger->Message(buffer.str());
-  buffer.str("");
+  std::string message = buffer.str();
+
+  if (!message.empty()) {
+    logger->Message(message);
+    buffer.str("");
+  }
+
   logger->Format(format);
   return *this;
 }
--- a/src/input_output/FGLog.h
+++ b/src/input_output/FGLog.h
@@ -59,7 +59,10 @@ class Element;
 // The return type of these functions is unspecified by the C++ standard so we
 // need some C++ magic to be able to overload the operator<< for these functions.
 using setprecision_t = decltype(std::setprecision(0));
+// For MSVC set_precision_t and setw_t are the same type
+#ifndef _MSC_VER
 using setw_t = decltype(std::setw(0));
+#endif
 
 enum class LogLevel {
   BULK,  // For frequent messages
@@ -121,7 +124,11 @@ public:
   FGLogging& operator<<(std::ostream& (*manipulator)(std::ostream&)) { buffer << manipulator; return *this; }
   FGLogging& operator<<(std::ios_base& (*manipulator)(std::ios_base&)) { buffer << manipulator; return *this; }
   FGLogging& operator<<(setprecision_t value) { buffer << value; return *this; }
+  // Avoid duplicate definition for MSVC for which set_precision_t and setw_t
+  // are the same type
+#ifndef _MSC_VER
   FGLogging& operator<<(setw_t value) { buffer << value; return *this; }
+#endif
   FGLogging& operator<<(const SGPath& path) { buffer << path; return *this; }
   FGLogging& operator<<(const FGColumnVector3& vec) { buffer << vec; return *this; }
   FGLogging& operator<<(LogFormat format);
--- a/src/input_output/FGLog.h
+++ b/src/input_output/FGLog.h
@@ -46,6 +46,7 @@ INCLUDES
 
 #include "simgear/misc/sg_path.hxx"
 #include "FGJSBBase.h"
+#include "math/FGColumnVector3.h"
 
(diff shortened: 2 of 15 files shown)


seanmcleod:
Looks good. Was a bit tedious glancing through the diff given how many lines were changed, but I guess even more tedious for you to edit them all 😉

bcoconni:
> Was a bit tedious glancing through the diff given how many lines were changed, but I guess even more tedious for you to edit them all 😉

Yep ! It's either a lot of small PRs or a few big PRs. I'm not sure which option is the best. Small PRs are easier for the reviewer but less PRs is best for the author. It really is depending on which side of the fence you stand 😉

seanmcleod:
Hmm, I decided to run some before versus after comparisons to double-check for any performance degradation. Basically using the same tests I did here - https://github.com/JSBSim-Team/jsbsim/pull/1094#issuecomment-2123177017 when we first discussed the logging changes.

So I built a release build before pulling these latest changes, and then another release build after pulling the latest changes. Ran each of them 3 times and averaged their runs.

```sh
$env:JSBSIM_DEBUG=0
measure-command { .\Release\JSBSim.exe --script=scripts\737_cruise.xml }
```

| Debug Level | Before (s) | After (s) |
| --- | --- | --- |
| 0 |  0.204 | 0.205 |
| 1 |  0.221 | 0.213 |
| 2 | 0.284 | 3.275 |
| 9 | 4.668 | 4.754 |

Note that for `JSBSIM_DEBUG=2` that the new build is more than 10x slower.

I did a quick comparison of the output, expecting to see maybe 10x more output, but there is very little difference.

| Redirected output file | Size | Lines |
| --- | --- | --- |
| Before | 779,302 bytes |  13,404 |
| After | 782,220 bytes | 13,400 |

While glancing through I did notice some differences in output number format.

```sh
Before:
  Mass Properties Report (English units: lbf, in, slug-ft^2)
                                      Weight    CG-X    CG-Y    CG-Z         Ixx         Iyy         Izz         Ixy         Ixz         Iyz
    Base Vehicle                       83000.0   639.0     0.0   -40.0    562000.0   1473000.0   1894000.0        -0.0      8000.0        -0.0
0   Fuel                                 10000     520     -80     -18           0           0           0
1   Fuel                                 10000     520      80     -18           0           0           0
2   Fuel                                  4000     480       0     -18           0           0           0
                                                                                                                                            
    Total:                          107000.0   610.8     0.0   -35.1    591572.3   1539552.7   1986235.4         0.0     19109.1         0.0


After:
  Mass Properties Report (English units: lbf, in, slug-ft^2)
                                      Weight    CG-X    CG-Y    CG-Z         Ixx         Iyy         Izz         Ixy         Ixz         Iyz
    Base Vehicle                         8e+04   6e+02       0  -4e+01       6e+05       1e+06       2e+06          -0       8e+03          -0
0   Fuel                                 10000     520     -80     -18           0           0           0
1   Fuel                                 10000     520      80     -18           0           0           0
2   Fuel                                  4000     480       0     -18           0           0           0
                                                                                                                                            
    Total:                             1e+05   6e+02       0  -4e+01       6e+05       2e+06       2e+06           0       2e+04           0

```

bcoconni:
Thanks for the feedback :+1:

I'm actually positively surprised by the performance of the code with debug level 0 and 1 :smile: Regarding, the other levels of debug that's more what I was expecting, including for levels 0 and 1. I'd suspect that heavily calling the constructor/destructor of `FGLogging` takes a significant toll but that's only a guess.

I'll investigate and most likely submit a PR - at least to fix the numbers formats.

seanmcleod:
So glancing at the constructors and destructors I noticed the call to `Flush()` was the only real work happening. So as a quick test I simply commented it out.

```c++
virtual ~FGLogging() { /*Flush();*/ }
```

And sure enough execution time dropped from 4.7s to 234ms!

Not sure we need so much flushing anyway 😉 

Anytime `std::endl` is written to the stream it generates a flush.

bcoconni:
> I noticed the call to `Flush()` was the only real work happening. So as a quick test I simply commented it out.
> 
> ```c++
> virtual ~FGLogging() { /*Flush();*/ }
> ```
> 
> And sure enough execution time dropped from 4.7s to 234ms!

Given that among other things `Flush()` displays the contents of the buffer (see line 54 of the code below), this is no surprise that you see such a drop in execution time: you're basically stripping significant bits of the message :wink: 

https://github.com/JSBSim-Team/jsbsim/blob/0b99bf8b173fe21cd3a1481695ff199b5e4a4335/src/input_output/FGLog.cpp#L49-L61

> Not sure we need so much flushing anyway 😉

As I explained in the PR #1094, the `Flush()` method is meant to be an equivalent to `EOT` (End Of Transmission):
> The `Flush` method is called to indicate the "end of transmission." At this point, the class developer can format or parse the message, append a timestamp, etc., then output the finalized message via their application interface. It is in this method that the class developer may trigger the opening of a window to show an error message and request user feedback, such as "Ok" or "Cancel."

The thing is that `FGLogConsole` sends messages to `std::cout` as they are received. It does not buffer the data so, for that particular class, the `Flush()` method indeed looks a bit over engineered.

> Anytime `std::endl` is written to the stream it generates a flush.

The occurrences of `std::endl` that are passed to `FGLogging` (i.e. to its buffer) are not transfered to `FGLogConsole` (i.e. to `std::cout`) so the flush signal of `std::endl` is swallowed by `FGLogging` and never seen by `FGLogConsole`.

PR #1145, opened 2024-08-24
Unit tests for the logging feature.

bcoconni:
As the title says.

~~In addition, this PR "#ifdef"s the format code in `FGLogConsole::Format` so that this method is not overloaded on the Windows platform. Indeed in the legacy code, the output formatting is only taking place on Unix platforms (i.e. Linux & MacOSX) which terminals are compatible with the escape codes below:~~

https://github.com/JSBSim-Team/jsbsim/blob/941da038d843bcf232f27bffbfc57ef7d6b34da8/src/FGJSBBase.cpp#L49-L73

~~As a result, there is little interest in overloading `FGLogConsole::Format` on Windows because the current code in `FGLogConsole::Format` just does nothing for that platform.~~

See discussion below for details.

### DIFF
--- a/src/input_output/FGLog.cpp
+++ b/src/input_output/FGLog.cpp
@@ -104,6 +104,7 @@ void FGLogConsole::Flush(void) {
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+#ifndef _WIN32
 void FGLogConsole::Format(LogFormat format) {
   switch (format)
   {
@@ -134,4 +135,5 @@ void FGLogConsole::Format(LogFormat format) {
     break;
   }
 }
+#endif
 }
--- a/src/input_output/FGLog.cpp
+++ b/src/input_output/FGLog.cpp
@@ -1,9 +1,9 @@
 /*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
- Module:       FGOutputType.cpp
+ Module:       FGLog.cpp
  Author:       Bertrand Coconnier
  Date started: 05/03/24
- Purpose:      Manage output of sim parameters to file or stdout
+ Purpose:      Manage the logging of messages
 
   ------------- Copyright (C) 2024 Bertrand Coconnier -------------
 
@@ -26,8 +26,8 @@
 
 FUNCTIONAL DESCRIPTION
 --------------------------------------------------------------------------------
-This is the place where you create output routines to dump data for perusal
-later.
+This is the place where the logging of messages is managed. The messages can be
+sent to the console, to a file, etc.
 
 HISTORY
 --------------------------------------------------------------------------------
--- a/src/input_output/FGLog.cpp
+++ b/src/input_output/FGLog.cpp
@@ -104,7 +104,6 @@ void FGLogConsole::Flush(void) {
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-#ifndef _WIN32
 void FGLogConsole::Format(LogFormat format) {
   switch (format)
   {
@@ -135,5 +134,4 @@ void FGLogConsole::Format(LogFormat format) {
     break;
   }
 }
-#endif
 }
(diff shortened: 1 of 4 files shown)


bcoconni:
This PR also fixes a bug in
https://github.com/JSBSim-Team/jsbsim/blob/941da038d843bcf232f27bffbfc57ef7d6b34da8/src/input_output/FGLog.h#L100
which is basically a no-op as it is interpreted by the compiler as:
```c++
virtual void SetLevel(LogLevel level) { this->level = this->level; }
```
i.e. `this->level` is affected to itself and the parameter `level` passed to the method is ignored. The fix consists in having different names for the method parameter `l` and the class member `level`. The unit tests included in this PR have allowed to detect this silly error.

seanmcleod:
> In addition, this PR "#ifdef"s the format code in FGLogConsole::Format so that this method is not overloaded on the Windows platform. Indeed in the legacy code, the output formatting is only taking place on Unix platforms (i.e. Linux & MacOSX) which terminals are compatible with the escape codes below:

Not so fast, I present to you on Windows 😉 Blue text, green text, bold text, underlined text.



So in effect the legacy code is outdated in terms of it's Windows console assumptions.

Now the trickier part. Figuring out which version of Windows included this support, Windows 10, or earlier? And then what happens on earlier versions of Windows that don't include support for this, do you get weird characters in the console output, or are they silently eaten/ignored?

bcoconni:
> So in effect the legacy code is outdated in terms of it's Windows console assumptions.

That's correct :+1: If I remove the #ifdef in `FGJSBBase` to allow using the escape codes, I get some color in both VS Code terminal and [Windows terminal](https://github.com/microsoft/terminal) (which you seem to be using as well according to the screenshot you've taken).

I've followed blindly the legacy code and always assumed that Windows could not support escape codes.

> Now the trickier part. Figuring out which version of Windows included this support, Windows 10, or earlier? And then what happens on earlier versions of Windows that don't include support for this, do you get weird characters in the console output, or are they silently eaten/ignored?

Escape codes appear explicitly for devices that do not support them. For instance, when the output is redirect to a file, you get this kind of garbage.

image

My guess is that escape characters are supported by the Windows terminal so anyone using it should see the coloring.

seanmcleod:
Looks like we could detect support at runtime by checking for `ENABLE_VIRTUAL_TERMINAL_PROCESSING`.

https://learn.microsoft.com/en-us/windows/console/getconsolemode?redirectedfrom=MSDN

seanmcleod:
> Escape codes appear explicitly for devices that do not support them. For instance, when the output is redirect to a file, you get this kind of garbage.

Yep, just seen that now in testing. What happens on Linux when you redirect the output to a file?

bcoconni:
> What happens on Linux when you redirect the output to a file?

Well, the screenshot was taken on Linux :wink:

seanmcleod:
Ah, I guess I've been so used to not seeing any virtual terminal code sequences on Windows when running JSBSim for years now and therefore also not seeing them show up in any redirected output files 😉 

Ah, never needed this JSBSim option until now. So if you need to redirect the output to view in some editor that doesn't support virtual terminal codes, as opposed to say typing/catting it to the console afterwards.

```sh
    --nohighlight  specifies that console output should be pure text only (no color)
```

bcoconni:
> Ah, never needed this JSBSim option until now. So if you need to redirect the output to view in some editor that doesn't support virtual terminal codes, as opposed to say typing/catting it to the console afterwards.
> 
> ```shell
>     --nohighlight  specifies that console output should be pure text only (no color)
> ```

Yep !

However it is useful that the escape sequences are recorded when the output is redirected to a file as it allows me to check that the new logging feature does not modify red into blue or that some underlining has not been lost in the transition process. Rather than visually checking that blue messages are still blue, I use `diff` between files and if some formatting has been lost, the modified escape sequences will be found by `diff`.

bcoconni:
SO I've removed the #ifdef mentioned in the original post so that Windows will be able to use the escape sequences later.

Note that the disabling of escape sequences only takes place if the code is compiled with MSVC. If the code is compiled with MinGW then the escape sequences are activated for Windows. This is the legacy behavior.

bcoconni:
> Looks like we could detect support at runtime by checking for `ENABLE_VIRTUAL_TERMINAL_PROCESSING`.
> 
> https://learn.microsoft.com/en-us/windows/console/getconsolemode?redirectedfrom=MSDN

I suggest that this will be integrated in a separate PR.

seanmcleod:
Yep, no problem doing this in a separate PR.

seanmcleod:
Submitted PR https://github.com/JSBSim-Team/jsbsim/pull/1146 to enable VT code based highlighting on Windows if supported.

PR #1020, opened 2024-01-20
fix: toupper is not a member of std

carlocorradini:
Fix #1019

PR #1146, opened 2024-08-24
Enable VT code based highlighting on Windows if supported

seanmcleod:


bcoconni:
The compilation fails on MinGW. It seems the constant `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is not recognized.

seanmcleod:
Yep, busy looking at it. Bit weird. Not sure if MinGW is making use of some other set, possibly older set of Windows header files compared to the MSVC builds.

I never made any changes to the set of Windows specific header files to get this code to build on MSVC. So suspecting MinGW may be using an older set before this define was included. 

https://github.com/JSBSim-Team/jsbsim/blob/dbf2eecf70dd92bf13e73f58ca7f8007e8250cb4/src/JSBSim.cpp#L60-L69

seanmcleod:
The define was added to the Windows 10 SDK around 2017.

seanmcleod:
Where and how do we install MinGW?

I thought it might be pre-installed in the `windows-2019` runner that we use, but don't see it listed here:

https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md

No immediate, obvious section in here that I spot in terms of installing MinGW.

https://github.com/JSBSim-Team/jsbsim/blob/master/.github/workflows/cpp-python-build.yml

bcoconni:
> I thought it might be pre-installed in the windows-2019 runner that we use, but don't see it listed here

Well, MinGW is also a name for "compiling with the gcc compiler on Windows". So in the document you mentioned, the `windows-2019 runner` is using gcc 8.1.0. And according to https://github.com/godotengine/godot/issues/58595#issuecomment-1060606212, the flag `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is supported since 7.0.0.

bcoconni:
OK I think I've found the issue. Still in the link I mentioned previously they say:
> GCC version is irrelevant, but GCC 8.x is 4 years old and might be shipped with the old MinGW headers 6.0.0.
>
> `ENABLE_VIRTUAL_TERMINAL_PROCESSING` should be available in the MinGW headers 7.0.0+ (current version is 9.0.0).

And since the `windows-2019` runner is using gcc 8.1.0 it means that it is using a version of MinGW older than 7.0.0 i.e. which does not define `ENABLE_VIRTUAL_TERMINAL_PROCESSING`.

seanmcleod:
Talking of `windows-2019` runner, I've been wondering if we should switch to `windows-2022` to move from Visual Studio 2019 to Visual Studio 2022, or whether there is a specific or good reason to stick with 2019.

bcoconni:
> Talking of `windows-2019` runner, I've been wondering if we should switch to `windows-2022` to move from Visual Studio 2019 to Visual Studio 2022, or whether there is a specific or good reason to stick with 2019.

Well, the idea is to support the oldest platform available to support users that cannot upgrade their platform for whatever reason. It's the same idea that makes the script support all versions of Python since 3.8.

seanmcleod:
> Well, the idea is to support the oldest platform available to support users that cannot upgrade their platform for whatever reason.

What sort of support are we thinking of? If we compile JSBSim on github using VS2022 that doesn't mean a user can't compile JSBSim on their own PC using VS2019. Unless we specifically switch to say assuming and using C++20 features and VS2019 (I haven't double-checked) only supports say C++17.

bcoconni:
> > Well, the idea is to support the oldest platform available to support users that cannot upgrade their platform for whatever reason.
> 
> What sort of support are we thinking of? If we compile JSBSim on github using VS2022 that doesn't mean a user can't compile JSBSim on their own PC using VS2019. Unless we specifically switch to say assuming and using C++20 features and VS2019 (I haven't double-checked) only supports say C++17.

Well, according to PR #1020 successfully compiling on MSVC 2019 does not mean a successful compilation on VS 2015... I'm no specialist in this area but just acknowledging the user feedback.

bcoconni:
I've a pushed a commit and rebased this PR so it is now successfully compiling on MinGW. However the unit test `FGLogTest.h` fails because it assumes that MSVC would not use escape sequences:
https://github.com/JSBSim-Team/jsbsim/blob/dbf2eecf70dd92bf13e73f58ca7f8007e8250cb4/tests/unit_tests/FGLogTest.h#L321-L325
and at some other places.

seanmcleod:
Yep, so as long as the unit tests are executed on Windows 10 or later then they'll pass if you remove the ifdefs.

bcoconni:
> Yep, so as long as the unit tests are executed on Windows 10 or later then they'll pass if you remove the ifdefs.

Done ! The PR now compiles successfully on all platforms and I have gorgeous colors in my terminal on Windows as well 😄 Thanks for the PR.

seanmcleod:
You'll have to spend more time on Windows then to enjoy it 😉

PR #1141, opened 2024-08-17
Fixes to the logging code

bcoconni:
Following @seanmcleod's comment https://github.com/JSBSim-Team/jsbsim/pull/1136#issuecomment-2286717524, this PR introduces some fixes to the logging format as well as a bug fix. So when this PR will be merged, the output should have the same format as previously.

Regarding the bug fix: it is occurring when using the debug level 2 (`JSBSIM_DEBUG=2`) where the member `FGFDMExec::Log` is referenced after it is deleted. Below is a diagram of the call sequence when the instance of `FGFDMExec` is destroyed. Note that `FGAuxiliary::Debug` calls the constructor of `FGLogging` only if `if (debuglvl & 2 || debuglvl & 16)`:
```mermaid
graph TD;
  C1[("`**FGFDMExec**`")];
  M1[/"`*Log*`"/];
  D1[["~FGFDMExec"]];
  D2[["`*~FGInitialConditions*`"]];
  D3[["`*~FGAuxiliary*`"]];
  D4[["`*FGAuxiliary::Debug*`"]];
  L[["`*FGLogging*`"]];
  C1-->D1-->D2-->D3-->D4;
  D4-.->|"if (debuglvl & 2 || debuglvl & 16)"|L;
  C1-.->M1;
  L-->M1;
```
So the member `FGFDMExec::Log` is referenced by the destructor `~FGAuxiliary` which means that `FGFMExec::Log` must be destroyed **after** the instance of `FGAuxiliary`. This is achieved by moving the `Log` member at the top of the members declarations in the class `FGFDMExec` since C++ ensures that members are constructed in the order they are declared, and destructed in the reverse order of their declaration.

PR #1147, opened 2024-08-25
Logging redesign - part 3

bcoconni:
After some consolidation of the logging code (PR #1141, ##1142 and ##1145), this PR resumes the migration of the code to the new logging system.

I took this opportunity to:
* Rename the member of `FGLogging` from `level` to `log_level` for more clarity and to avoid confusion with the method parameters of `FGLogging` which tend to be named `level` as well.
* The minimum level feature has been moved from `FGLogger` to `FGLogConsole` as I think the implementation of this feature should not be imposed by the mother class `FGLogger`.
  * The unit test of `FGLogConsole` has been extended to check that the minimum logged level behave as intended.
  * The minimum logged level default is set to `LogLevel::BULK` so that it does not filter anything by default.

There are also a few white spaces changes because I've set my editor to remove trailing spaces before saving. Yes I'm that kind of person who can't resist taking care of this kind of futile details :smile:

### DIFF
--- a/src/input_output/FGLog.cpp
+++ b/src/input_output/FGLog.cpp
@@ -85,7 +85,7 @@ FGXMLLogging::FGXMLLogging(std::shared_ptr<FGLogger> logger, Element* el, LogLev
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 void FGLogConsole::Flush(void) {
-  switch (level)
+  switch (log_level)
   {
   case LogLevel::BULK:
   case LogLevel::DEBUG:
--- a/src/input_output/FGLog.h
+++ b/src/input_output/FGLog.h
@@ -147,7 +147,7 @@ class JSBSIM_API FGLogConsole : public FGLogger
 public:
   void SetMinLevel(LogLevel level) { min_level = level; }
   void FileLocation(const std::string& filename, int line) override
-  { buffer << std::endl << "In file " << filename << ": line " << line << std::endl; }
+  { buffer << "\nIn file " << filename << ": line " << line << "\n"; }
   void Format(LogFormat format) override;
   void Flush(void) override;
 
--- a/src/input_output/FGLog.h
+++ b/src/input_output/FGLog.h
@@ -97,13 +97,13 @@ class JSBSIM_API FGLogger
 {
 public:
   virtual ~FGLogger() {}
-  virtual void SetLevel(LogLevel l) { level = l; }
+  virtual void SetLevel(LogLevel level) { log_level = level;}
   virtual void FileLocation(const std::string& filename, int line) {}
   virtual void Message(const std::string& message) = 0;
   virtual void Format(LogFormat format) {}
   virtual void Flush(void) {}
 protected:
-  LogLevel level = LogLevel::BULK;
+  LogLevel log_level = LogLevel::BULK;
 };
 
 class JSBSIM_API FGLogging
@@ -152,7 +152,7 @@ public:
   void Flush(void) override;
 
   void Message(const std::string& message) override {
-    if (level < min_level) return;
+    if (log_level < min_level) return;
     buffer << message;
   }
 
--- a/src/input_output/FGLog.h
+++ b/src/input_output/FGLog.h
@@ -99,13 +99,11 @@ public:
   virtual ~FGLogger() {}
   virtual void SetLevel(LogLevel l) { level = l; }
   virtual void FileLocation(const std::string& filename, int line) {}
-  void SetMinLevel(LogLevel level) { min_level = level; }
   virtual void Message(const std::string& message) = 0;
   virtual void Format(LogFormat format) {}
   virtual void Flush(void) {}
 protected:
   LogLevel level = LogLevel::BULK;
-  LogLevel min_level = LogLevel::INFO;
 };
 
 class JSBSIM_API FGLogging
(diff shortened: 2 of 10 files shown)


seanmcleod:
Looks good.

One follow-up question, partly triggered by https://github.com/JSBSim-Team/jsbsim/issues/834#issuecomment-2305032022 in terms of state reports from notify sections in scripts.

When we update `FGScript.cpp` do these state reports requested by a user via a script naturally fall into our `LogLevel`?

https://github.com/JSBSim-Team/jsbsim/blob/f33f20e18f18acb3b55c1c9bb093a6ff5ac08337/src/input_output/FGLog.h#L66-L73

Or are they separate from debug logging, with it's varying levels of debug detail/level?

bcoconni:
> One follow-up question, partly triggered by [#834 (comment)](https://github.com/JSBSim-Team/jsbsim/issues/834#issuecomment-2305032022) in terms of state reports from notify sections in scripts.
> 
> When we update `FGScript.cpp` do these state reports requested by a user via a script naturally fall into our `LogLevel`?
> 
> https://github.com/JSBSim-Team/jsbsim/blob/f33f20e18f18acb3b55c1c9bb093a6ff5ac08337/src/input_output/FGLog.h#L66-L73
> 
> Or are they separate from debug logging, with it's varying levels of debug detail/level?

That's a good question.

All messages are meant to be processed by `FGLogging` to save having several classes. Maybe we could add an additional log level (`USER` ?) that has a higher priority than all the other log levels to guarantee it is always displayed ?

bcoconni:
And speaking about #834, I have just committed some changes to replace the occurrences of `std::endl` by `\n`. The changes are just applied to the files that were planned to be updated by this PR. The replacement of `std::endl` in the files that have been modified by the PRs #1094 and #1136 will be addressed in a separate PR

seanmcleod:
> replace the occurrences of `std::endl` by `\n`.

Yep, I understand the performance reason, personally just find it more difficult to read the text in the editor with `\n' appended to the last word 😢 

Not sure if the following compile time concatenation would be to the liking of other developers 😉 

```c++
cout << " aircraft config file. (LIFT DRAG)\n";

cout << " aircraft config file. (LIFT DRAG)" "\n";
```

> Maybe we could add an additional log level (USER ?) that has a higher priority than all the other log levels to guarantee it is always displayed?

Hmm, wondering whether we end up needing a bitmask approach?

bcoconni:
> Not sure if the following compile time concatenation would be to the liking of other developers 😉
> 
> ```c++
> cout << " aircraft config file. (LIFT DRAG)\n";
> 
> cout << " aircraft config file. (LIFT DRAG)" "\n";
> ```

I have no strong opinion regarding this. Feel free to modify this PR branch (`bcoconni/logging3`) to your liking 😄 

> > Maybe we could add an additional log level (USER ?) that has a higher priority than all the other log levels to guarantee it is always displayed?
> 
> Hmm, wondering whether we end up needing a bitmask approach?

What do you mean ? Can you detail your thoughts ?

bcoconni:
I've just pushed another commit to this PR to replace a couple of `normint` by `LogLevel::NORMAL` that have been overlooked in PR #1136.

bcoconni:
> I have no strong opinion regarding this. Feel free to modify this PR branch (`bcoconni/logging3`) to your liking

Or alternatively you can create your own branch and I'll cherry pick your commits to reformat the `\n`'s.

seanmcleod:
> Or alternatively you can create your own branch and I'll cherry pick your commits to reformat the `\n`'s.

Or should I wait for you to finish updating the whole code base to use the new logging mechanism, and then submit a PR that does this reformatting across the whole code base as a single commit?

bcoconni:
> > Or alternatively you can create your own branch and I'll cherry pick your commits to reformat the `\n`'s.
> 
> Or should I wait for you to finish updating the whole code base to use the new logging mechanism, and then submit a PR that does this reformatting across the whole code base as a single commit?

Sounds good :+1:

seanmcleod:
> > > Maybe we could add an additional log level (USER ?) that has a higher priority than all the other log levels to guarantee it is always displayed?
> > 
> > Hmm, wondering whether we end up needing a bitmask approach?
>
> What do you mean ? Can you detail your thoughts ? 

So we currently have:

https://github.com/JSBSim-Team/jsbsim/blob/e3ac75afabf0a57c343629293ec2c1114e3076be/src/input_output/FGLog.h#L66-L73

With a rough assumption of decreasing verbosity from `BULK` to `FATAL` (typically a maximum of 1 of these 😉) , so a JSBSim user/developer gets to set a minimum level via `SetMinLevel(LogLevel)`. Which probably works for the vast majority of uses?

But I was wondering whether a user/developer may want to cherry pick, i.e. via a bitmask, e.g. selecting for example, `DEBUG`, `WARN`, `ERROR`, `FATAL`.

bcoconni:
> With a rough assumption of decreasing verbosity from `BULK` to `FATAL` (typically a maximum of 1 of these 😉) , so a JSBSim user/developer gets to set a minimum level via `SetMinLevel(LogLevel)`. Which probably works for the vast majority of uses?
> 
> But I was wondering whether a user/developer may want to cherry pick, i.e. via a bitmask, e.g. selecting for example, `DEBUG`, `WARN`, `ERROR`, `FATAL`.

I agree that the method `FGLogConsole::SetMinLevel()` has extremely basic capabilities at filtering the log messages but this can be changed later on.

As you mentioned, different users may have very different needs and for that very reason I think that filtering strategies must not be dictated from the top by the `FGLogging` class but entirely delegated to the inherited class.

That being said, if we want to allow using a bitmask, I think it's possible to attribute power of 2's values to each element of `enum class LogLevel`, something like:
```c++
 enum class LogLevel { 
   BULK  = 1,  // For frequent messages 
   DEBUG = 2,  // Less frequent debug type messages 
   INFO  = 4,  // Informatory messages 
   WARN  = 8,  // Possible impending problem 
   ERROR = 16, // Problem that can be recovered 
   FATAL = 32  // Fatal problem => an exception will be thrown 
 };
```

agodemar:
Will we merge this PR? @bcoconni

bcoconni:
> Will we merge this PR? @bcoconni

Yes, sure. Done ! 😉

PR #1152, opened 2024-09-07
Fix exceptions thrown by FGFunction and caught by FGAerodynamics

bcoconni:
The class `FGAerodynamics` currently catches `std::string` exceptions thrown by the constructor of `FGFunction`. However `FGFunction` is now throwing a mix of exceptions deriving from both `std::string` and `JSBSim::BaseException`, so `FGAerodynamics` is no longer catching all exceptions thrown by the constructor of `FGFunction`.

This PR is changing:
* the remaining `std::string` exceptions in `FGFunction` to `JSBSim::BaseException` 
* the format of exceptions caught by `FGAerodynamics` from `std::string` to `JSBSim::BaseException`.

Also the `try`/`catch` block in `FGAerodynamics` is simplified in the process. Currently the code looks like:
```c++
if (condition) {
  try {
    function1 = new FGFunction();
  }
  catch (const string& str) {
    log << "Fatal error: " << str;
  }
}
else {
  try {
    function2 = new FGFunction();
  }
  catch (const string& str) {
    log << "Fatal error: " << str;
  }
}
```
This PR avoids the duplication of the `try`/`catch` block by moving the `if`/`else` condition into the `try` block:
```c++
try {
  if (condition) {
    function1 = new FGFunction();
  } else {
    function2 = new FGFunction();
  }
}
catch (const BaseException& e) {
    log << "Fatal error: " << e.what();
}
```

### DIFF
--- a/src/math/FGFunction.cpp
+++ b/src/math/FGFunction.cpp
@@ -53,19 +53,19 @@ constexpr unsigned int MaxArgs = 9999;
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-class WrongNumberOfArguments : public runtime_error
+class WrongNumberOfArguments : public BaseException
 {
 public:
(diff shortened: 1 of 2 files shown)


seanmcleod:
Looks good.

PR #1161, opened 2024-09-22
Render debug data and visuals after actor's pose is updated

seanmcleod:
Based on the discussion in https://github.com/JSBSim-Team/jsbsim/discussions/1160 it was determined that the debug data strings and lines for gear positions etc. were being rendered before the actor's pose had been updated for the latest frame. Which meant the debug data would be displaying slightly out of date data causing potential confusion.

### DIFF
--- a/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp
+++ b/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp
@@ -291,13 +291,6 @@ void UJSBSimMovementComponent::TickComponent(float DeltaTime, ELevelTick TickTyp
             // Get the results from JSBSim
 			CopyFromJSBSim();
 
-            // Basic debugging string and symbols
-			if (DrawDebug)
-			{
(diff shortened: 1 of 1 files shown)


gallonmate:
Great find! Tested and works.

bcoconni:
Thanks for the PR, @seanmcleod. Just a tiny comment: the indent seems different after than it was before. Some tabs changed to spaces during copy/pasting or something like that ?

seanmcleod:
My VS editor is using the JSBSim standard 😉 

https://github.com/JSBSim-Team/jsbsim/blob/master/.editorconfig

```sh
(base) C:\source\jsbsim>type .\.editorconfig
root = true

[*.{h,hpp,cpp}]
indent_style = space
indent_size = 2
```

The diff - https://github.com/JSBSim-Team/jsbsim/pull/1161/commits/2f177d9edaa467b77eb346f4f77d341814b44943 in Github online shows thing aligned, except for the explicit change I made to the comment indent. Although double-checking it looks like the original code used tabs.

I was planning on submitting a separate commit to fix up all the other comment indents, in which case I'll get the editor to convert all the tabs as well to match the JSBSim standard.

bcoconni:
> I was planning on submitting a separate commit to fix up all the other comment indents, in which case I'll get the editor to convert all the tabs as well to match the JSBSim standard.

All good then :smile:

Issue #1159, opened 2024-09-18
Request: round tag


Octal450:
**I'm submitting a ...**
  - [x] feature request

**Describe the request**

Inside `<function>`, currently I am using the integer tag with a summation of 0.5 to round values. It would be great to have a tag that works like this:

```
<round>
    <property>myprop</property>
    <unit>5</unit>
</round>
```

In this case, rounding myprop to 5.

Which support rounding to any arbitrary number (10, 5, 12.5, etc) including decimal like 0.1, 0.2, etc.

**What is the motivation / use case for changing the behavior?**
Not needing to do this all over my code :D
![image](https://github.com/user-attachments/assets/ef0bdb69-97f8-4082-9f48-bc10e4e19041)

Thanks,
Josh

bcoconni:
You could define a function template (note the hashtag `#` that will be replaced by the actual property when the template function will be applied later):
```xml
<function name="my-correct-round" type="template">
  <quotient>
    <integer>
      <sum>
        <product>
          <property> # </property>
          <value> 10.0 </value>
        </product>
        <value> 0.5 </value>
      </sum>
    </integer>
    <value> 10.0 </value>
  </quotient>
</function>
```
then apply it to the property `hydraulics/stabilizer/final-deg` using the `apply=""` attribute:
```xml
<sum>
  <property apply="my-correct-round"> hydraulics/stabilizer/final-deg </property>
  <property> some/other/property </property>
</sum>
```

seanmcleod:
Ah, I had forgotten about the support for template functions, even though I added a bunch a number of years ago 😉 

Add a set of template functions for unit conversions to SI units - https://github.com/JSBSim-Team/jsbsim/commit/cfd181b5f8a99f9ee77abe209474125ef25ca4d6

seanmcleod:
Although I'm sure the next request will be to allow parameters to be passed to the template functions 😉

Octal450:
I didn't realize you could do that. But I would have to copy it thru all the planes. Hence the request. :)

-J

seanmcleod:
Ideally you'd stick this in a single `rounding.xml` file and then include/reference it in each aircraft's FDM that requires it.

https://github.com/JSBSim-Team/jsbsim/blob/57361d0756a7a4d477d80f786e381317d8722671/scripts/737_cruise.xml#L78-L83

seanmcleod:
So I tried a quick test with the aerodynamics section.

```xml
[...]

PR #1170, opened 2024-10-03
Add roundmultiple function

seanmcleod:
Based on the request and discussions in issue - https://github.com/JSBSim-Team/jsbsim/issues/1159

I'm in 2 minds whether to rather add a separate `round(x)` function in addition to `roundmultiple(x, m)` where currently if no second argument is provided to `roundmultiple` then it assumes a value of 1.

### DIFF
--- a/src/math/FGFunction.cpp
+++ b/src/math/FGFunction.cpp
@@ -527,6 +527,15 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
                  return y != 0.0 ? fmod(p[0]->GetValue(), y) : HUGE_VAL;
                };
       Parameters.push_back(new aFunc<decltype(f), 2>(f, fdmex, element, Prefix, var));
+    } else if (operation == "roundmultiple") {
+      auto f = [](const decltype(Parameters)& p)->double {
+                 double multiple = p.size() == 1 ? 1.0 : p[1]->GetValue();
+                 return round((p[0]->GetValue() / multiple)) * multiple;
+               };
+      if (element->GetNumElements() == 1)
+        Parameters.push_back(new aFunc<decltype(f), 1>(f, fdmex, element, Prefix, var));
+      else
+        Parameters.push_back(new aFunc<decltype(f), 2>(f, fdmex, element, Prefix, var));
     } else if (operation == "atan2") {
       auto f = [](const decltype(Parameters)& p)->double {
                  return atan2(p[0]->GetValue(), p[1]->GetValue());
--- a/src/math/FGFunction.cpp
+++ b/src/math/FGFunction.cpp
@@ -533,7 +533,7 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
                  return round((p[0]->GetValue() / multiple)) * multiple;
                };
       if (element->GetNumElements() == 1)
-        Parameters.push_back(new aFunc<decltype(f), 1>(f, fdmex, element, Prefix, var));
+        Parameters.push_back(make_MathFn(round, fdmex, element, Prefix, var));
       else
         Parameters.push_back(new aFunc<decltype(f), 2>(f, fdmex, element, Prefix, var));
     } else if (operation == "atan2") {
--- a/src/math/FGFunction.cpp
+++ b/src/math/FGFunction.cpp
@@ -535,7 +535,7 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
       if (element->GetNumElements() == 1)
         Parameters.push_back(make_MathFn(round, fdmex, element, Prefix, var));
       else
-        Parameters.push_back(new aFunc<decltype(f), 2>(f, fdmex, element, Prefix, var));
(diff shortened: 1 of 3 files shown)


Octal450:
Many thanks. My personal option is that unless there's some reason, having the shortcut tag would be nice. 

Kind Regards,
Josh

src/math/FGFunction.cpp
```diff
@@ -527,6 +527,15 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
                  return y != 0.0 ? fmod(p[0]->GetValue(), y) : HUGE_VAL;
                };
       Parameters.push_back(new aFunc<decltype(f), 2>(f, fdmex, element, Prefix, var));
+    } else if (operation == "roundmultiple") {
+      auto f = [](const decltype(Parameters)& p)->double {
+                 double multiple = p.size() == 1 ? 1.0 : p[1]->GetValue();
+                 return round((p[0]->GetValue() / multiple)) * multiple;
+               };
+      if (element->GetNumElements() == 1)
+        Parameters.push_back(new aFunc<decltype(f), 1>(f, fdmex, element, Prefix, var));
```
bcoconni, line 536: If there is only one parameter just use plain `round()` instead of using the lambda `f` which divide/multiply by 1.0
```suggestion
        Parameters.push_back(make_MathFn(round, fdmex, element, Prefix, var));
```

src/math/FGFunction.cpp
```diff
@@ -527,6 +527,15 @@ void FGFunction::Load(Element* el, FGPropertyValue* var, FGFDMExec* fdmex,
...
+      else
+        Parameters.push_back(new aFunc<decltype(f), 2>(f, fdmex, element, Prefix, var));
```
bcoconni, line 538: Fix the function declaration so that it reports an error if the number of arguments is either 0 or higher than 2. With the current code in the PR, the error message will tell that exactly 2 arguments must be supplied which is not what we want since `<roundmultiply>` also accepts a single argument.
```suggestion
        Parameters.push_back(new aFunc<decltype(f), 1>(f, fdmex, element, Prefix, var, 2));
```
Indeed if `<roundmultiply>` is supplied with only one argument, this branch will not be executed. If there are no arguments or more than 2 arguments, this branch will be executed and will report an error. Therefore, this branch will only succeed if 2 arguments are supplied to `<roundmultiply>`. Hence the proposed modification.

bcoconni (changes requested):
Thanks for the PR @seanmcleod. I suggest a couple of modifications:
* Use plain `round()` when `<roundmultiply>` is used with a single argument.
* Fix the parameter passed to `aFunc<>` so that the error message do not tell that 2 arguments are required when `<roundmultiply>` is used with zero or more than 2 arguments.

bcoconni:
I must express that I'm not excited about this PR since `<roundmultiply>` is a trivial function, which could simply be replaced by `<round>` with a single parameter as I've explained in https://github.com/JSBSim-Team/jsbsim/issues/1159#issuecomment-2383766723. I don't believe adding such trivial functions to `FGFunction` is beneficial. However, if everyone else agrees to proceed in this direction, I won't be the one to cause issues.

bcoconni (approved):

Octal450:
I think it's really useful, so thank you gents.

bcoconni:
> I think it's really useful, so thank you gents.

I would have much preferred a solid argument about why this change is making a difference rather than having a mere opinion that *you think* it is useful. Anyway, whatever…

bcoconni:
@seanmcleod, as far as I am concerned, you can push the "merge" button whenever you feel this PR ready for being pushed to `master` (well, I mean you mentioned you were in 2 minds).

Octal450:
> > I think it's really useful, so thank you gents.
> 
> I would have much preferred a solid argument about why this change is making a difference rather than having a mere opinion that _you think_ it is useful. Anyway, whatever…

I didn't think you were asking for one, so I didn't want to argue :)

But fair enough.

When creating instrumentation, JSBSim systems are the best way to go. They are much more efficient and fast than FlightGear's Nasal and in some cases, more flexible. However, instrumentation almost always displays nice formatted numbers to the pilot, rather than 3.5308928409832904 or whatever.

Rounding is the quick solution, hence the nice round tag... But - 

Often times, displaying things to the 1th decimal place can be as unhelpful at a glance, for numbers that are very large (Ie, fuel flow on jetliners). So, these instruments are rounded to the 10th, or sometiimes the 50th.

Second case, I use JSBSim to do all my performance computations when simulating aircraft flight computers and I need to be able to round the numbers to various places there.

There is just 2 use-cases where a flexible rounding operation is useful, I can list more.

I think that just like mod allows you to specify the value, and isn't just fixed to a value like 2, I think rounding should be equally flexible, as a basic function.

Absolutely you right that doing more complex things with it makes no sense to include in the code here, but I think, round to an nth is a nice standard operation.

Bests,
Josh

agodemar:
@seanmcleod I'm sorry for the inconvenience. I assumed that the motivations given by @Octal450 are acceptable.
Please, go ahead with your work on this refinement.

seanmcleod:
@agodemar no problem, not a big deal 😉

PR #1172, opened 2024-10-09
Simplify code slightly in roundmultiple

seanmcleod:


seanmcleod:
@agodemar although fairly trivial changes I'd prefer to wait for @bcoconni to double-check them first, especially since they were made in response to some of his suggestions. And he's very good at spotting issues/better ways to do things as well 😉 

Lastly there is no rush. There is an implementation already in master that users can start making use of, this is really just a bit of tidying up of the code.

bcoconni:
>  I'd prefer to wait for @bcoconni to double-check them first, especially since they were made in response to some of his suggestions. And he's very good at spotting issues/better ways to do things as well 😉

Thanks for the pat on the back 😄 
Looks good to me.

Issue #475, opened 2021-07-11
Floating Point Exception in Actions build process


bcoconni:
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**

**Describe the issue**
The build for MSVC repeatedly fails while executing the test `CheckScripts.py` and more specifically while running the script `c172_runway_at_rest_cg_shift.xml`.

**What is the current behavior?**
The test reports an FPE (Floating Point Exception) which seems to be located in the waypoint/geodetic computations code.
[Here is the stack trace as reported by the test failure](https://github.com/JSBSim-Team/jsbsim/runs/3041318962?check_suite_focus=true#step:14:966):
```text
#11   Source "D:\a\jsbsim\jsbsim\build\python\jsbsim.cxx", line 14008, in __pyx_pf_6jsbsim_9FGFDMExec_10run [00007FFF462BEB9D]
       14005:  */
       14006:   __Pyx_XDECREF(__pyx_r);
       14007:   try {
      >14008:     __pyx_t_1 = __pyx_v_self->thisptr->Run();
       14009:   } catch(...) {
       14010:     convertJSBSimToPyExc(); if (!PyErr_Occurred())PyErr_SetString(PyExc_RuntimeError, "Error converting c++ exception.");
       14011:     __PYX_ERR(0, 579, __pyx_L1_error)
#10   Source "D:\a\jsbsim\jsbsim\src\FGFDMExec.cpp", line 423, in JSBSim::FGFDMExec::Run [00007FFF46310F82]
        420:   // returns true if success, false if complete
        421:   if (Script && !IntegrationSuspended()) success = Script->RunScript();
        422: 
      > 423:   for (unsigned int i = 0; i < Models.size(); i++) {
        424:     LoadInputs(i);
        425:     Models[i]->Run(holding);
        426:   }
#9    Source "D:\a\jsbsim\jsbsim\src\models\FGFCS.cpp", line 171, in JSBSim::FGFCS::Run [00007FFF4637492E]
        168:   for (i=0; i<SystemChannels.size(); i++) {
        169:     if (debug_lvl & 4) cout << "    Executing System Channel: " << SystemChannels[i]->GetName() << endl;
        170:     ChannelRate = SystemChannels[i]->GetRate();
      > 171:     SystemChannels[i]->Execute();
        172:   }
        173:   ChannelRate = 1;
[...]

PR #1173, opened 2024-10-13
Fix an uninitialized variable in GeographicLib.

bcoconni:
Fixes the issue #475.

~~Unlike the option that has been chosen by the GeographicLib project (see https://github.com/geographiclib/geographiclib/issues/33#issuecomment-2408610762), the variable `m12x` is set to zero, instead of `Math::NaN`, to avoid raising floating point exceptions.~~

A new floating point exception flag `FE_OVERFLOW` has also been added to `JSBSim.cpp` because it is the exception that is triggered by this bug and `JSBSim.cpp` was not catching it on Linux. This setting helped debugging the issue while running the script `c3101.xml` on an Ubuntu distribution.

**UPDATE:** The variable `m12x` is initialized to `Math::NaN` as in the GeographicLib project. See the discussion below for details.

### DIFF
--- a/src/GeographicLib/Geodesic.cpp
+++ b/src/GeographicLib/Geodesic.cpp
@@ -206,7 +206,7 @@ namespace GeographicLib {
     // check, e.g., on verifying quadrants in atan2.  In addition, this
     // enforces some symmetries in the results returned.
 
-    real sbet1, cbet1, sbet2, cbet2, s12x, m12x;
+    real sbet1, cbet1, sbet2, cbet2, s12x, m12x = 0.0;
 
     Math::sincosd(lat1, sbet1, cbet1); sbet1 *= _f1;
     // Ensure cbet1 = +epsilon at poles; doing the fix on beta means that sig12
--- a/src/GeographicLib/Geodesic.cpp
+++ b/src/GeographicLib/Geodesic.cpp
@@ -206,7 +206,7 @@ namespace GeographicLib {
     // check, e.g., on verifying quadrants in atan2.  In addition, this
     // enforces some symmetries in the results returned.
 
-    real sbet1, cbet1, sbet2, cbet2, s12x, m12x = 0.0;
+    real sbet1, cbet1, sbet2, cbet2, s12x, m12x = Math::NaN();
 
     Math::sincosd(lat1, sbet1, cbet1); sbet1 *= _f1;
     // Ensure cbet1 = +epsilon at poles; doing the fix on beta means that sig12
--- a/src/JSBSim.cpp
+++ b/src/JSBSim.cpp
@@ -292,7 +292,7 @@ int main(int argc, char* argv[])
   _controlfp(_controlfp(0, 0) & ~(_EM_INVALID | _EM_ZERODIVIDE | _EM_OVERFLOW),
            _MCW_EM);
 #elif defined(__GNUC__) && !defined(sgi) && !defined(__APPLE__)
-  feenableexcept(FE_DIVBYZERO | FE_INVALID);
+  feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
 #endif
 
   try {


seanmcleod:
@bcoconni I'm refreshing my memory regarding the original issue and trying to understand the issue with setting `m12x = Math::NaN`.

So first off in terms of Python versus JSBSim exe, it looks to me like in both cases for MSC which I'm using to test, that the set of enabled floating point exceptions are the same? 

https://github.com/JSBSim-Team/jsbsim/blob/0f00b806a74027d71a3e758bbb1e177e4b6130f2/python/fpectl/fpectlmodule.cpp#L142-L148

https://github.com/JSBSim-Team/jsbsim/blob/0f00b806a74027d71a3e758bbb1e177e4b6130f2/src/JSBSim.cpp#L288-L294

Testing a number of scripts I hit a breakpoint here:

https://github.com/JSBSim-Team/jsbsim/blob/0f00b806a74027d71a3e758bbb1e177e4b6130f2/src/GeographicLib/Geodesic.cpp#L406

Originally, using a debug build and without `m12x` being initialized I was seeing a constant value for it, I'm assuming the compiler was setting since I saw the same value for different uninitialized doubles in multiple functions, set to `9.2559631349317831e+61`. Which didn't cause any floating point exceptions since it's not too big to cause an overflow etc.

I then initialized `m12x` to `Math:NaN()` and checked what happened when I hit the breakpoint. And no floating point exception, the result is just `NaN` after the multiplication with `b`.

> If any input has a quiet NaN value, it will cause the result of nearly any floating-point operation that consumes it to be NaN without raising the invalid exception.

And checking what `Math::NaN()` returns:

https://github.com/JSBSim-Team/jsbsim/blob/0f00b806a74027d71a3e758bbb1e177e4b6130f2/src/GeographicLib/Math.cpp#L260-L270

So when you say:

> variable `m12x` is set to zero, instead of `Math::NaN`, to avoid raising floating point exceptions.

Are you seeing examples of `Math::NaN()` returning `numeric_limits<T>::max()` on some of the platforms you're testing on and then causing an overflow exception? Or are you just being cautious given that some platforms may not have a `numeric_limits<T>::quiet_NaN()`?

I'm assuming when we were seeing this intermittent bug it was because `m12x` was randomly sometimes ending up with a value close to max and then causing an overflow exception.

But if we're confident that all modern (?) compilers/platforms have a quiet NaN then wouldn't it be better to go with `Math::NaN()` so that we don't have to worry about having to perform an update to the GeoGraphicLib each time we pull in an update of it?

bcoconni:
> So first off in terms of Python versus JSBSim exe, it looks to me like in both cases for MSC which I'm using to test, that the set of enabled floating point exceptions are the same?

Yes, correct.

> Originally, using a debug build and without `m12x` being initialized I was seeing a constant value for it, I'm assuming the compiler was setting it since I saw the same value for different uninitialized doubles in multiple functions, set to `9.2559631349317831e+61`. Which didn't cause any floating point exceptions since it's not too big to cause an overflow etc. But big enough and somewhat unique to stand out so that developers would notice it when debugging.

In the case of the Ubuntu distribution (which is the one used by the GitHub runners), `m12x` can sometimes be set to a value which is high enough to trigger an overflow FPE. Of course this condition is met by accident and is dependent on the stack history since `m12x` is most likely allocated in the stack by C++ compilers. Hence the randomness of the bug.

By chance, this condition was systematically fulfilled at [some point of the development of my `logging4` branch](https://github.com/bcoconni/jsbsim/actions/runs/10835603092). This was the moment when the bug became reproducible and when I have finally been able to track the bug down to the lack of initialization of `m12x`.

> Are you seeing examples of `Math::NaN()` returning `numeric_limits<T>::max()` on some of the platforms you're testing on and then causing an overflow exception? Or are you just being cautious given that some platforms may not have a `numeric_limits<T>::quiet_NaN()`?

No, I have **assumed** that multiplying a NaN by a number would raise an invalid FPE. And you are entirely right: my assumption is wrong and setting `m12x` to `Math::Nan()` does not trigger any floating point exception. All the tests that I have run following your feedback have passed and proven your point. I now know what "quiet" means in `quiet_NaN` :smile:

> I'm assuming when we were seeing this intermittent bug it was because `m12x` was randomly sometimes ending up with a value close to max and then causing an overflow exception.

Agreed, I guess that compilers are allocating `m12x` in the stack so its value depends on the previous content of the stack which is more or less random.

> But if we're confident that all modern (?) compilers/platforms have a quiet NaN then wouldn't it be better to go with `Math::NaN()` so that we don't have to worry about having to perform an update to the GeoGraphicLib each time we pull in an update of it?

Yes, that's a good point. Added to the fact that I was incorrectly assuming that using `quiet_Nan` in `m12x *= b` would raise an FPE, I see no longer a reason to initialize `m12x` to zero.

bcoconni:
@seanmcleod have all your concerns be addressed ?

seanmcleod:
Yep.

bcoconni:
Right, the PR is now merged.
Thanks for the feedback.

PR #1175, opened 2024-10-25
Make FGInertial::GetGM const

nsunderland1:
I couldn't think of any reason this shouldn't be const, it's a simple getter.


seanmcleod:
Yep, can't see why it shouldn't. I've approved the workflow run to build and test in the meantime.

bcoconni:
PR merged. Thanks @nsunderland1 for spotting this.

PR #1185, opened 2024-11-08
Reset Logic Improvments

mhand1998:


bcoconni:
Sorry but what is this PR ?

Warning39:
@bcoconni (Sean here), this is the second or third time that I've seen this sort of pull request from someone at machindustries.

I'm pretty sure it's an accident, i.e. they have their own fork of JSBSim, and this PR was more than likely intended for their own repo and not the official JSBSim repo that they forked from.

bcoconni:
> @bcoconni (Sean here), this is the second or third time that I've seen this sort of pull request from someone at machindustries.
> 
> I'm pretty sure it's an accident, i.e. they have their own fork of JSBSim, and this PR was more than likely intended for their own repo and not the official JSBSim repo that they forked from.

@mhand1998, is it indeed the case ?

PR #1191, opened 2024-11-20
fix some bounds checks in FGFCS

roccoblues:
This looks like copy-n-paste leftovers.


bcoconni:
Thanks for the PR @roccoblues 

All these vectors are supposed to have the same size since their size increases simultaneously when `FGFCS::AddThrottle` is called.
https://github.com/JSBSim-Team/jsbsim/blob/bb4cd5b6c25366d60cce98954426927e83d65ff9/src/models/FGFCS.cpp#L697-L710
However, it indeed improves the code legibility so PR merged.

PR #1193, opened 2024-11-22
Fix: Don't polute global namespace with std.

cbirkhold:
`using namespace ...;` in a (especially a public) header file is bad form.

props.hxx does collapse the std namespace into the global namespace which can lead to problems in client libraries.

Of course there may also be client code out there now that depends on this which would require that code to either add `std::` in some places or add `using namespace std` in the right places (which hopefully is inside another namespace or a compilation unit.


bcoconni:
Good catch ! PR merged.

PR #1194, opened 2024-11-22
Optimizations

cbirkhold:
1/ Avoid copies and temporaries.
2/ Faster access to atmosphere values when multiple altitude dependent values are queried.

### DIFF
--- a/tests/unit_tests/FGAtmosphereTest.h
+++ b/tests/unit_tests/FGAtmosphereTest.h
@@ -150,8 +150,24 @@ public:
     const double nu0 = mu0/rho0;
 
     for(double h=-1000.0; h<10000; h+= 1000) {
-      double T = T0 + 0.1*h;
-      double P = P0 + 1.0*h;
+      const double T = T0 + 0.1*h;
+      const double P = P0 + 1.0*h;
+      const double rho = P/(R*T);
+      const double a = sqrt(gama*R*T);
+
+      FGAtmosphere::AltitudeDependentValues adv;
+      atm.GetAltitudeDependentValues(adv, h);
+      TS_ASSERT_DELTA(adv.Temperature, T, epsilon);
+      TS_ASSERT_DELTA(adv.Pressure, P, epsilon);
+      TS_ASSERT_DELTA(adv.Density, rho, epsilon);
+      TS_ASSERT_DELTA(adv.SoundSpeed, a, epsilon);
+
+      FGAtmosphere::AltitudeDependentValues adv_0;
+      atm.GetAltitudeDependentValues(adv_0, 0.0);
+      TS_ASSERT_EQUALS(adv_0.Temperature, T0);
+      TS_ASSERT_EQUALS(adv_0.Pressure, P0);
+      TS_ASSERT_DELTA(adv_0.Density, rho0, epsilon);
+      TS_ASSERT_DELTA(adv_0.SoundSpeed, a0, epsilon);
 
       TS_ASSERT_DELTA(atm.GetTemperature(h), T, epsilon);
       TS_ASSERT_EQUALS(atm.GetTemperature(0.0), T0);
@@ -160,11 +176,9 @@ public:
       TS_ASSERT_DELTA(atm.GetPressure(h), P, epsilon);
       TS_ASSERT_EQUALS(atm.GetPressure(0.0), P0);
 
-      double rho = P/(R*T);
       TS_ASSERT_DELTA(atm.GetDensity(h), rho, epsilon);
       TS_ASSERT_DELTA(atm.GetDensity(0.0), rho0, epsilon);
 
-      double a = sqrt(gama*R*T);
       TS_ASSERT_DELTA(atm.GetSoundSpeed(h), a, epsilon);
       TS_ASSERT_DELTA(atm.GetSoundSpeed(0.0), a0, epsilon);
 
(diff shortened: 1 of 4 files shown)


bcoconni:
It seems to me this PR should be split into 2 PRs because it is addressing 2 unrelated topics. Could you please do that ?

Regarding the atmosphere optimization, it seems to me that the speed improvement of your PR is only for those who use `FGDefaultGroundCallback`. Indeed your change adds a level of indirection to existing implementations of `FGGroundCallback` which complexifies the code and most certainly does not give any speed improvement in return. AFAIC `FGDefaultGroundCallback` exists to avoid implementing boiler plate code for POC or basic usage of JSBSim so this is not the path I'd spontaneously optimize. I'd rather do the opposite i.e. optimize the complex path for custom implementations of `FGDefaultGroundCallback`.

However I'm a bit skeptical that your PR brings any measurable improvement to the speed of execution. In which case, the additional complexity would not be payed back. So could you please provide some data to substantiate your claim ?

Thanks.

cbirkhold:
> It seems to me this PR should be split into 2 PRs because it is addressing 2 unrelated topics. Could you please do that ?

Certainly!

> Regarding the atmosphere optimization, it seems to me that the speed improvement of your PR is only for those who use `FGDefaultGroundCallback`. Indeed your change adds a level of indirection to existing implementations of `FGGroundCallback` which complexifies the code and most certainly does not give any speed improvement in return. AFAIC `FGDefaultGroundCallback` exists to avoid implementing boiler plate code for POC or basic usage of JSBSim so this is not the path I'd spontaneously optimize. I'd rather do the opposite i.e. optimize the complex path for custom implementations of `FGDefaultGroundCallback`.

In the case where a custom FGGroundCallback implementation is used but does not provide an overload for GetAGLevel(const FGLocation&) the extra indirection would indeed be incurred, however this only replaces the indirection already present in the code path without the change so there actually is no delta in cost or complexity of the code path.

I intended to follow the pattern of the other non-pure overload just above the proposed change which has the same effect in terms of indirections and is the default method called on the interface (FGInertial).

If I'm not mistaken there is an automatic benefit to clients using the default implementation. When customized and only the pure-virtual method is overridden the cost is the same as before. At the same time, clients that care about performance would now have a simple way to support the optimized path by adding the overload to their custom implementation. The only negative effect would be for a custom implementation that provides an override for the non-pure-virtual method above the proposed change (which is the one used internally by FGInertial for instance) but not the new proposed overload.

> However I'm a bit skeptical that your PR brings any measurable improvement to the speed of execution. In which case, the additional complexity would not be payed back. So could you please provide some data to substantiate your claim ?

You will probably see a fairly small effect in the test cases/samples of this project. In applications that mix high-fidelity JSBSim simulated entities with a large number of simpler entities stepped outside JSBSim but operating in the same world and querying the JSBSim environment for certain properties the effect can become quite relevant.

Totally understand though if this is not a change that makes sense to integrate from your perspective, it certainly is easy enough to maintain as a patch on the client side. Much larger gains are certainly possible when touching/limiting features, though I haven't been able to wrap those changes in a way that makes sense for the project repository yet. With this I've been trying to push back some changes that have come out of profile guided optimization and are more localized in case they make sense for the community.

I will split up the PR to make it easier for you to integrate (or not).

As always thanks for your work on this valuable project!

bcoconni:
> In the case where a custom FGGroundCallback implementation is used but does not provide an overload for GetAGLevel(const FGLocation&) the extra indirection would indeed be incurred, however this only replaces the indirection already present in the code path without the change so there actually is no delta in cost or complexity of the code path.

I'm afraid not. Currently, for all classes inheriting from `FGGroundCallabck`, the call graph is the following:
```mermaid
flowchart LR;
A[FGInertial::GetAltitudeAGL];
B["FGGroundCallback::GetAGLevel(FGLocation&, FGLocation&, FGColumnVector3&, FGColumnVector3&, FGColumnVector3&)"];
A-->B;
```
With your PR, for classes inheriting from `FGGroundCallabck` and which do not implement the new `GetAGLevel` method, the call graph would be:
```mermaid
flowchart LR;
A[FGInertial::GetAltitudeAGL];
B(["FGGroundCallback::GetAGLevel(FGLocation&)"]);
C["FGGroundCallback::GetAGLevel(FGLocation&, FGLocation&, FGColumnVector3&, FGColumnVector3&, FGColumnVector3&)"];
A-->B-->C;
```
So there is indeed an additional level of indirection and therefore a delta in cost and complexity.

> If I'm not mistaken there is an automatic benefit to clients using the default implementation.

Not sure what you mean, here.

> When customized and only the pure-virtual method is overridden the cost is the same as before.

That's only the case for classes that implement the new method `GetAGLevel(FGLocation&)`.

> At the same time, clients that care about performance would now have a simple way to support the optimized path by adding the overload to their custom implementation. The only negative effect would be for a custom implementation that provides an override for the non-pure-virtual method above the proposed change (which is the one used internally by FGInertial for instance) but not the new proposed overload.

There are 2 strong assumptions behind these statements:
1. Clients are willing to modify the code of their implementation of `FGGroundCallback` i.e. have the time, workforce and budget of doing so.
2. Your change brings some performance improvements which could be an incentive if the assumption above is met.

However, none of these assumptions can be taken for granted.

> You will probably see a fairly small effect in the test cases/samples of this project. In applications that mix high-fidelity JSBSim simulated entities with a large number of simpler entities stepped outside JSBSim but operating in the same world and querying the JSBSim environment for certain properties the effect can become quite relevant.

I'd certainly merge your change if you could **prove** that it brings some improvements. From what I'm seeing from your PR, the changes in performance - if any - will be negligible so there is not much point (performance-wise) in modifying the code. I'd be happy to be proven wrong but at the moment there are no facts to substantiate your claims so I'm staying on the conservative side.

bcoconni:
This PR is on hold for more than 1 year now without any progress.
PR closed.

PR #1200, opened 2024-11-26
Cygwin build fix

landswellsong:
Ensures `strtod_l` is provided which is normally hidden for non-GNU source

### DIFF
--- a/src/input_output/string_utilities.cpp
+++ b/src/input_output/string_utilities.cpp
@@ -34,6 +34,9 @@ string::find() functions were formerly used.
 INCLUDES
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
+#ifdef __CYGWIN__
+#define _GNU_SOURCE 1
+#endif
 #include <errno.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


Saleh-At:
add these on top of file `"jsbsim\src\input_output\string_utilities.cpp"`

`#define _GNU_SOURCE
#include <_ansi.h>
#include <errno.h>
#include <stdlib.h>
#include <string.h>
#include <cstdlib>
#include <stdlib.h>
#include <stdio.h>
#include <errno.h>
#include <iostream>
#include <sstream>
#include <stdio.h>`

bcoconni:
@landswellsong, @Saleh-At I'm not sure I'm understanding your request ? Is the PR in its current form complete for the compilation of JSBSim with CygWin or should some further changes be added ?

Please note that some headers are appearing twice in @Saleh-At suggestion (`errno.h`, `stdlib.h`, `stdio.h`, …) and would the PR be modified to include these additional headers then please remove redundant code and update the `#ifdef __CYGWIN__` accordingly to avoid polluting the other compilers.

Thanks for the PR.

landswellsong:
@bcoconni it works for me as is without additional headers, just the commit which I've sent. I'm unsure why do any additional headers need to be included, it seemed to have only be a problem with one particular unit file.

bcoconni:
PR merged. Thanks !

Issue #1187, opened 2024-11-18
Reset command causes rapid time-catch-up in realtime mode


coasho:
Bug Description
When running JSBSim in realtime mode and sending a set simulation/reset 0 command, the simulation momentarily resets but then rapidly accelerates until it catches up to the pre-reset time before resuming realtime execution. This behavior appears to be caused by time tracking variables not being reset along with the simulation state.
Steps to Reproduce

Run JSBSim in realtime mode with any aircraft
Let the simulation run for some time (e.g., 30 seconds)
Send command: set simulation/reset 0
Observe the simulation behavior

Current Behavior
After the reset command:

Simulation briefly resets to initial conditions
Immediately accelerates at maximum speed
Runs until reaching the pre-reset simulation time
Finally resumes normal realtime execution

Expected Behavior
The simulation should reset to initial conditions and continue running in realtime from that point, without any rapid time catch-up phase.
Root Cause
In the realtime execution loop, the time lag calculation uses initial_seconds and current_seconds to determine how far the simulation is behind real time:
```cpp
actual_elapsed_time = current_seconds - initial_seconds;
sim_lag_time = actual_elapsed_time - FDMExec->GetSimTime();
```
When ResetToInitialConditions() resets sim_time to 0 but doesn't reset initial_seconds and current_seconds, it creates a large artificial time lag that the simulation tries to catch up to.
Proposed Solution
Time tracking variables need to be reset when ResetToInitialConditions() is called. A possible approach would be to:

Move time tracking variables out of JSBSim.cpp's real_main()
Make them accessible to both the main loop and FGFDMExec
Reset them in ResetToInitialConditions()

Impact
This bug affects any application using JSBSim in realtime mode that needs to reset the simulation state during execution. It can cause unexpected behavior in flight training scenarios or interactive simulations where resets are common.
Additional Notes

Severity: Medium
Component: Core/Runtime
The bug does not affect batch mode operation

bcoconni:
Thanks for the report @coasho and thanks for the proposed solution but I'd rather use `SGPropertyChangeListener` to detect when the value of the property `simulation/reset` is set to 0.
https://github.com/JSBSim-Team/jsbsim/blob/bb4cd5b6c25366d60cce98954426927e83d65ff9/src/simgear/props/props.hxx#L722-L749
Indeed the class `FGFDMExec` should remain isolated from the considerations of running the FDM in real time or in batch mode.

[...]

PR #1213, opened 2024-12-30
Prevent time catch up in real time mode

bcoconni:
This PR fixes the bug reported in the issue #1187.

A new class `Timer` is added that manages elapsed time. It is inheriting `SGPropertyChangeListener` so that every time the property `simulation/reset` is set the timer is properly reset.

The CI workflow is also modified to check that executing the following command takes at least 90 seconds (i.e. 2 runs of 25 seconds and 1 run of 40 seconds):
```bash
JSBSim scripts/c172_cruise_8K.xml --realtime --nice
```
The script `c172_cruise_8K.xml` is also simplified:
* It runs quicker: the end time and the time that triggers the reset have been shortened.
* The termination event has been removed. Instead the last run is executed until the end time is reached.

### DIFF
--- a/.github/workflows/cpp-python-build.yml
+++ b/.github/workflows/cpp-python-build.yml
@@ -509,6 +509,17 @@ jobs:
           rm -f _jsbsim.cxx  # Make sure that jsbsim.cxx is not stored in the source distribution
           python -m build --sdist
           echo "::endgroup::"
+      - name: Check reset for real time execution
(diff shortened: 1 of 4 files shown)


coasho:
thanks @bcoconni for implementing this fix! The new Timer class with SGPropertyChangeListener handles the reset behavior really well. I tested PR #1213 and it works exactly as expected now.

PR #1215, opened 2025-01-07
fixing argparsing of planet file

T-Wainwright:
Minor bug fix- looks like there's a name mismatch of the argparse definition and namespace name of the planet file. Currently python doesn't work as is.

### DIFF
--- a/python/JSBSim.py
+++ b/python/JSBSim.py
@@ -53,7 +53,7 @@ parser.add_argument("--suspend", default=False, action="store_true",
                     help="specifies to suspend the simulation after initialization")
 parser.add_argument("--initfile", metavar="<filename>",
                     help="specifies an initialization file")
-parser.add_argument("--planet", metavar="<filename>",
+parser.add_argument("--planetfile", metavar="<filename>",
                     help="specifies a planet definition file")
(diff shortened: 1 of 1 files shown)


agodemar:
@bcoconni do you approve this?

T-Wainwright:
Happy to elaborate, I expect I'm running into this as someone coming at it from a pure python scripting perspective, so might have slipped through the testing cracks. But either way no way of running in python script mode currently works.

1. pip install jsbsim
2. JSBSim.py

You'll get the error:
```JSBSim startup beginning ...

  File "/Users/tom.wainwright/Documents/cases/HELSING/JSBSim/jsb_env/bin/JSBSim.py", line 184, in <module>
    if args.planetfile:
       ^^^^^^^^^^^^^^^```
       
Alternative change is to change L184 to `if args.planet:`

bcoconni:
> Alternative change is to change L184 to `if args.planet:` But either way the argparse name, and the reference to it need to match up. I wasn't aware there was convention in other bits of the code so happy to just change L184 too.

Thanks for further explanations. Good catch ! So yes, please replace L184 as you suggested.

bcoconni:
The PR has been merged.
Thanks for your contribution !

PR #1218, opened 2025-01-21
Solve post-impact collapse problem by adding wind simulation

Xi3Chen:
1. Fixed plane crash caused by rendering thread after impact.
2. Added support for wind simulation

### DIFF
--- a/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Public/FDMTypes.h
+++ b/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Public/FDMTypes.h
@@ -439,15 +439,15 @@ UENUM(BlueprintType)
 enum class ETurbType:uint8
 {
 	//No turbulence model is used. The aircraft will not experience any random wind effects. This mode can be beneficial for initial testing or when the influence of wind needs to be simplified.
-	None = 0 UMETA(ToolTip = "turbulence disabled"),
+	None = 0 UMETA(DisplayName ="turbulence disabled"),
 	//A basic turbulence model that uses standard statistical characteristics to simulate wind fluctuations. This model is generally used for simple flight simulations.
-	Standard UMETA(ToolTip = "Ordinary turbulence"),
+	Standard UMETA(DisplayName = "Ordinary turbulence"),
 	//The Culp turbulence model typically provides a more detailed simulation of turbulence (considering factors such as turbulence intensity and frequency). The Culp model adapts well to different aircraft and flight conditions, making it suitable for more complex simulations.
-	Culp UMETA(ToolTip = "Culp turbulence model"),
+	Culp UMETA(DisplayName = "Culp turbulence model"),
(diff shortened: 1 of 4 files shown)


gallonmate:
Thanks for adding UE5 support for wind types and the solution to stop the program crashing!
I added a review for small changes regarding the comments/tool tips.
It is causing a compile error on engine version UE5.4 and UE5.0.
 `error : Metadata key 'ToolTip'` basically, from my understanding the engine can only use one "tool tip" for each enum. 

If I remove either the tooltip or the code comment above each enum, then it compiles successfully.

For example. this code below doesn't work because the engine is trying to use both the comment and the tooltip:
```c++
  //None = No turbulence model is used. The aircraft will not experience an
  None = 0 UMETA(ToolTip="turbulence disabled"),
```

The following code works and the compile is successful:
``` c++ 
  //No turbulence model is used. The aircraft will not experience any random wind effects
  None =0,
```

This also works and the compile is successful:
``` c++ 
  None = 0 UMETA(ToolTip="turbulence disabled"),
```

UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Public/FDMTypes.h
```diff
@@ -433,3 +433,70 @@ struct FAircraftState
...
+	//The Culp turbulence model typically provides a more detailed simulation of turbulence (considering factors such as turbulence intensity and frequency). The Culp model adapts well to different aircraft and flight conditions, making it suitable for more complex simulations.
+	Culp UMETA(ToolTip = "Culp turbulence model"),
+	//This model uses the Dryden spectrum to simulate turbulence, adhering to the guidelines set in the MIL-F-8785C document. The parameters are designed differently for flights at altitudes below 1000 feet and above 2000 feet, with linear interpolation applied for altitudes in between. This model is well-suited for military applications and scenarios where specific turbulence characteristics are required.
+	Milspec UMETA(ToolTip = "Milspec turbulence model (Dryden spectrum)"),
+	//Similar to ttMilspec, this model also uses the Dryden spectrum. The main difference lies in how the transfer functions are implemented based on the specifications in the military document. It helps in simulating realistic turbulence under similar conditions as the Milspec model.
+	Tustin UMETA(ToolTip = "Tustin turbulence model (Dryden spectrum)") ,
```
gallonmate, line 450: Please remove either the  `UMETA(ToolTip = "xxxx")` from each enum or remove the `//comments`. 
Only one is need for each enum. The `//comments` are automatically used to create the tool tips in the editor and it is causing compile build error if used in combination with UMETA ToolTip.

gallonmate (changes requested):

UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Public/FDMTypes.h
Xi3Chen: Yeah, I was going to write displayName. I'll correct him now

gallonmate (approved):

bcoconni:
PR merged.
Thanks @Xi3Chen for your contribution and thanks @gallonmate for the code review.

PR #1225, opened 2025-02-01
Switched to the Enhanced Input system and included the Raw Input Plug…

jdcrayme:
Switched to the Enhanced Input system and included the Raw Input Plugin for flight controller support.


gallonmate:
Thanks for doing the work to update/switch to Enhanced Input!
Which engine version did you do the update with? 
I'm not sure it's time yet to update the base project version in this repo yet. Since the nature of these changes breaks compatibility with previous UE5 versions. And the engine version regarding .uassets, cannot be opened in prior engine versions.
Some discussion needs to take place. 

My thought would be some kind of separate update, or add the necessary files to a new folder that users have easy access to. You could also fork the repo and then we just update the ReadMe to link to your fork. Or we can ask the JSBSim repo owners if we can add a branch or archive of the current project so that we have a legacy version easily accessible for all UE5 versions. And then the main repo can be for the latest UE5 updates.

But since the current Input System in UE5 still works. I'm not sure it's worth breaking compatibility of the sample project in the main repo just yet. I'm thinking maybe when UE5.6 or UE5.7 comes out, then we can look at updating the base project version. If there is a way to support Enhanced Input while not breaking Compatibility with prior UE5 versions, then I suggest we do that. I briefly looked into Enhanced Input and seeing it it was possible to keep the project compatibility with all UE5 versions, but I did not initially see a solution.

jdcrayme:
That makes sense. 

We could move the input logic to the player controller, then select either the EI or legacy controller in the world settings. While I was converting the logic, I observed that both the legacy and enhanced input systems worked simultaneously, so apart from needing to enable the Enhanced Input and RAW Input plugins, having an enhanced input player controller to choose from shouldn't break anything else in the project..

gallonmate:
After another brief look, the beta version of Enhanced Input that comes with UE5.0 might be compatible with UE5.5. It could be missing some settings but it probably works fine.

But I like your idea of using the player controller class. Since that keeps the old input system intact which is simple to use, especially to newcomers. And still gives the option to easily switch to Enhanced Input. 

I'm curious how much complexity or confusion that adds down the line. For example if a user switches to Enhanced Input, and then makes changes to the BP_Airliner's legacy input controls, is there any input conflict or confusion in general? I wonder if having two different BP_Airliners is less complex than a new controller class.

AirshipSim:
> After another brief look, the beta version of Enhanced Input that comes with UE5.0 might be compatible with UE5.5. It could be missing some settings but it probably works fine.
It is compatible, I have projects in 5.0 with more complex inputs than the demo that I upgraded to 5.5 without any change to the Enhanced Input code.

> But I like your idea of using the player controller class. Since that keeps the old input system intact which is simple to use, especially to newcomers. And still gives the option to easily switch to Enhanced Input.
In a realistic scenario, I think a newcomer downloads the latest UE 5.5 and then just tries to run the demo from there without much thinking. The demo is intended as a starting point anyway, anyone a bit serious will immediately write custom code with own blueprints. At least this is what I did, and what Epic recommends in the official DIY tuto. The added benefit is that hardcore sim fans, the most likely to contribute to JSBSim, can quickly use their fancy joysticks.

> I'm curious how much complexity or confusion that adds down the line. For example if a user switches to Enhanced Input, and then makes changes to the BP_Airliner's legacy input controls, is there any input conflict or confusion in general? I wonder if having two different BP_Airliners is less complex than a new controller class.
I agree that there may be confusion, and two BP_Airliners does not sound good. In particular, remember there are already two airliners blueprints - from the marketplace asset and from the Epic Game employee. It is already confusing enough.

Again, in the perspective that the demo is only a starting point, I propose to simplify:
Let's `git tag` the latest commit before the merge of this PR as "legacy input demo" then move on.

@gallonmate if you need a tester for @jdcrayme 's branch before merge, let me know.

gallonmate:
Updating the project and specific the blueprints to a higher engine version completely ends compatibility for prior versions. At this time there is no technical reason to break compatibility. You mentioned enhanced input does work with UE5.0, which supports  there is no need to update the base engine version.

Going to the newest base version will effectively end any JSBSim plugin updates, fixes, etc, for the users who are not using UE5.5.
For example this JSBSim user posted 4 days: "I am new to the UE and jsbsim." - "How can I modify the aircraft in UE5.3?" https://github.com/JSBSim-Team/jsbsim/discussions/1238

And in general looking at the poll from a few weeks ago, 50% of UE5 users are using a previous engine version. At this time, I cannot recommend upgrading the base version of this sample project to UE5.5. Although I have no strong objection to upgrading the base version to UE5.3

https://www.reddit.com/r/unrealengine/comments/1ifw9ys/poll_which_unreal_version_do_you_currently_use/
image

gallonmate:
To clarify, if jdcrayme, or another dev, wishes to update the Enhanced Input starting by using engine version UE5.0-UE5.3, then I support that. Otherwise I support we simply add a link to their fork in the Readme. The third option is the UE5.5 Blueprint Airliner with enhanced input gets renamed BP_Airliner_EnhancedInput. Which allows UE5.5 users easy access to enhanced input, while maintaining compatibility for all engine versions.

bcoconni:
> Or we can ask the JSBSim repo owners if we can add a branch or archive of the current project so that we have a legacy version easily accessible for all UE5 versions.

> Let's `git tag` the latest commit before the merge of this PR as "legacy input demo" then move on.

I have no strong opinion regarding the merging of this PR but I would like to mention that I would be reluctant to have 2 different versions of the UE plugin in this repo. Either we drop the old one in favor of this PR or we keep the legacy code.

Having 2 different versions would mean maintaining the 2 of them in a working stage thus doubling the maintenance effort. And by maintenance I mean the maintenance of the plugin itself as well as the main library. Users of one plugin version will most likely open issues or discussions to be on par feature-wise with the other version.

Just my 2 cents.

bcoconni:
> Or we can ask the JSBSim repo owners if we can add a branch or archive of the current project so that we have a legacy version easily accessible for all UE5 versions.

> Let's `git tag` the latest commit before the merge of this PR as "legacy input demo" then move on.

I have no strong opinion regarding the merging of this PR but I would like to mention that I would be reluctant to have 2 different versions of the UE plugin in this repo. Either we drop the old one in favor of this PR or we keep the legacy code.

Having 2 different versions would mean maintaining the 2 of them in a working stage thus doubling the maintenance effort. And by maintenance I mean the maintenance of the plugin itself as well as the main library. Users of one plugin version will most likely open issues or discussions to be on par feature-wise with the other version.

Just my 2 cents.

AirshipSim:
> Going to the newest base version will effectively end any JSBSim plugin updates, fixes, etc, for the users who are not using UE5.5.

I am not sure I follow you @gallonmate. The input code and blueprints are only limited to the UEReferenceApp and its corresponding uproject (/Source folder), I dont see how it impacts the plugin (Plugins/JSBSimFlightDynamicsModel). I dont remember any hard-coded input code in the plugin, it is all blueprint in the UEReferenceApp. The 47 modified PR files are none of the plugin files.

When I switched my own project to Enhanced Input, I did not have to change the plugin in any way. I continue to benefit from all new plugin update. It is recompiled according to my selected UE version.

In any case, @jdcrayme 's input mapping needs to be updated according to the FlighGear mapping, or the README input table must be updated. As long as this is not done, this PR is broken User side and should not be merged in my humble opinion.

gallonmate:
> I dont see how it impacts the plugin (Plugins/JSBSimFlightDynamicsModel).

I'm referring to this entire Unreal Engine sample project. Yes the actual plugin code within the project is not affected by this Pull Request. If using that logic, that means this update has no benefit to that plugin code itself. 
> When I switched my own project... I continue to benefit....

The update only benefits UE5.5 users who are using the sample project, or in the situation someone is cherry picks the plugin, along with the enhanced input files, and is using UE5.5. Meanwhile this update breaks general compatibility for users of previous engine versions for this Unreal Engine JSBSim Plugin UEReferenceApp Project. 


 image It still remains that this Pull Request is based using UE5.5. The project and all .uassets files in this update will only work in UE5.5+. While the project version can be simply edited to previous version in order to open, the blueprint airliner cannot easily open. Which is the entire point of the sample project. In this manner, the update effectively ends any updates or fixes for people who use the entire sample project and are not using UE5.5

I included statistics of the percentage of users per engine version. It's good practice to keep this Unreal Engine JSBSim Plugin UEReferenceApp Project available to as many users as possible. I do not understand your wish to exclude users. As I demonstrated a new user 4 days ago is using UE5.3, and is using the entire Unreal Engine JSBSim Plugin UEReferenceApp Project, not just the plugin code.

It's also not just a preference to use whichever engine version users have on hand. There are also technical reasons why people use previous engine versions, such as various engine bugs. If you're not familiar with people experiencing engine bugs, issues, or specific reasons why they use different engine versions, then I suggest you spend time researching the topic and history of Unreal Engine versions. 

There is no hardship or technical limitation in making this Enhanced Input update using a previous engine version, in order to keep the Unreal Engine JSBSim Plugin UEReferenceApp Project available to as many users as possible. I also provided suggestions on how include this Enhanced Input update mostly as-is and not break compatibility.

I'm not going to further discuss the topic of engine compatibility or why keeping this Unreal Engine JSBSim Plugin UEReferenceApp Project to the minimum engine version is beneficial.

bcoconni:
> I'm not going to further discuss the topic of engine compatibility or why keeping this Unreal Engine JSBSim Plugin UEReferenceApp Project to the minimum engine version is beneficial.

Fair enough. I suggest we keep this PR open until it is deemed that we no longer need to support UE5.3. Or one find a way to make it compatible with all supported versions of UE. Does that make sense ?

gallonmate:
> Fair enough. I suggest we keep this PR open until it is deemed that we no longer need to support UE5.3. Or one find a way to make it compatible with all supported versions of UE. Does that make sense ?

Absolutely. 

To get back to the original point, if there is a direction that @jdcrayme can choose from. This PR is compatible by doing one the following changes:

-use a previous engine version to redo this pull request. This is the best option but understandably it's a large request to recreate work.

-rename the bp_airliner file, allowing previous version to still exist. (and undo/remove the .uproject version change, etc).

-share their fork link in the Unreal project ReadMe so UE5.5 users have easy access, while main project stays compatible with all UE versions until some time in the future.

Or if there is some other suggestion/idea that's suitable.

bcoconni:
This PR has been submitted one year ago. @gallonmate should we merge it now ?

gallonmate:
I wouldn't recommend merging it yet. Unreal Engine is currently at version 5.7, this PR would bump the plugin to 5.5 but standard UE practice is to support at least 3 engine versions back. So when UE reaches 5.8 (maybe June this year) the current PR could be merged, unless the PR author or someone else would like to remake this PR with support for UE 5.4 or 5.3 and so it could be merged sooner.

gallonmate:
@bcoconni This PR for Unreal can finally be merged at your leisure, as Unreal Engine reached version 5.8 preview.

bcoconni:
Thanks for the heads up @gallonmate, I suggest this PR will be merged in the `master` branch after version 1.3.1 has been released (i.e. [hopefully tomorrow](https://github.com/JSBSim-Team/jsbsim/discussions/1456))

bcoconni:
The PR is now merged in the `master` branch.

PR #1252, opened 2025-04-05
Reset turbulence vector when turbulence is disabled

seanmcleod70:
Fix the issue reported by - https://github.com/JSBSim-Team/jsbsim/discussions/1250 in terms of the turbulence effect not being reset when the turbulence was set to None.

### DIFF
--- a/tests/TestTurbulenceGustStop.py
+++ b/tests/TestTurbulenceGustStop.py
@@ -28,90 +28,84 @@ class TestTurbulenceGustStop(JSBSimTestCase):
     BASELINE_WIND_EAST  = 6
     BASELINE_WIND_DOWN  = 7
 
-    def testCosineGust(self):
-        fdm = self.setup()
+    def setUp(self, *args): 
+        super().setUp(*args)
 
-        # Setup cosine gust
-        fdm["atmosphere/cosine-gust/startup-duration-sec"] = 5
-        fdm["atmosphere/cosine-gust/steady-duration-sec"] = 1
-        fdm["atmosphere/cosine-gust/end-duration-sec"] = 5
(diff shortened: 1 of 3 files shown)


tests/TestTurbulenceGustStop.py
```diff
@@ -0,0 +1,120 @@
...
+            fdm.run()
+
+        # Reset turbulence
+        fdm["atmosphere/turb-type"] = 0
+
+        fdm.run()
+
+        self.checkBaselineWind(fdm)
+
+    def setup(self):
```
bcoconni, line 88: The name `setup` is almost colliding with the method `JSBSimTestCase.setUp` (note the capital *U* in `setUp`). I would suggest to either choose a different name or rename `setup` to `setUp` and inherit from `JSBSimTestCase.setUp`. Check the code from `TestInputSocket.py` below for an example:

https://github.com/JSBSim-Team/jsbsim/blob/e1ea4e85524e209037ec868f5a1383ec2ef3c7a1/tests/TestInputSocket.py#L59-L64

If you choose renaming `setup` to `setUp` then please note that you will no longer need to make explicit calls to this method as `setUp` is automatically called before each test (i.e. each method which name starts with `test`) is run.

seanmcleod70: Thanks, I'll take a look. Need to get more up to speed on the test framework 😉

seanmcleod70: I've pushed a commit to update the test code to make use of the existing `setUp()` method.

bcoconni: Looks good to me.

PR #1253, opened 2025-04-06
Allow for a separate random seed for wind turbulence

seanmcleod70:
As per the discussion at https://github.com/JSBSim-Team/jsbsim/discussions/1250 allow the user to specify a separate random seed for the random number generator used by the turbulence code in `FGWinds`.

I've tested this by running `JSBSim.exe` with a script and varying the random seeds and comparing the wind output in the CSV file.

So in this example, running the script twice and toggling ``simulation/randomseed` between 1 and 2 but keeping `atmosphere/randomseed` constant gives the exact same wind output with turbulence enabled.

```xml
  <run start="0" end="15" dt="0.008333">

	  <property value="1"> simulation/randomseed </property>
	  <property value="3"> atmosphere/randomseed </property>
```

As mentioned here - https://github.com/JSBSim-Team/jsbsim/discussions/1250#discussioncomment-12736631 I haven't been able to write a python test case using the test setup we have.

### DIFF
--- a/src/models/atmosphere/FGWinds.h
+++ b/src/models/atmosphere/FGWinds.h
@@ -40,6 +40,7 @@ INCLUDES
 
 #include "models/FGModel.h"
 #include "math/FGMatrix33.h"
+#include <optional>
 
 /*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 FORWARD DECLARATIONS
@@ -402,11 +403,11 @@ private:
   FGColumnVector3 vBurstGust;
   FGColumnVector3 vTurbulenceNED;
 
-  unsigned int RandomSeed;
+  std::optional<unsigned int> RandomSeed;
   std::shared_ptr<RandomNumberGenerator> generator;
 
   void SetRandomSeed(int sr);
-  int  GetRandomSeed(void) const { return RandomSeed; }
+  int  GetRandomSeed(void) const;
 
   void Turbulence(double h);
   void UpDownBurst();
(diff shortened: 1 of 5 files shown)


src/models/atmosphere/FGWinds.h
```diff
@@ -394,8 +402,12 @@ class FGWinds : public FGModel {
   FGColumnVector3 vBurstGust;
   FGColumnVector3 vTurbulenceNED;
 
+  unsigned int RandomSeed;
   std::shared_ptr<RandomNumberGenerator> generator;
 
+  void SetRandomSeed(int sr);
+  int  GetRandomSeed(void) const { return RandomSeed; }
```
bcoconni, line 409: You must account for the case where `SetRandomSeed` has not yet been called. In this case, `GetRandomSeed` must return the random seed from `FGFDMExec`.

Now that we have moved to C++17, this is a good use case for `optional<>`:
```suggestion
  std::optional<unsigned int> RandomSeed;
  std::shared_ptr<RandomNumberGenerator> generator;

  void SetRandomSeed(int sr);
  int  GetRandomSeed(void) const {
    if (RandomSeed)
      return *RandomSeed;
    else
      return FDMExec->SRand();
  }
```

bcoconni (changes requested):
Great job! It's good to have you back on board @seanmcleod70 👍 

I have suggested one change to `GetRandomSeed` to cover the corner case where `FGFDMExec` and `FGWinds` are using the same random number generator in which case `FGWinds::GetRandomSeed` must return the same value than `FGFDMExec::Srand`.

src/models/atmosphere/FGWinds.h
seanmcleod70: Okay, haven't used `optional<>` before, but will need to make `unsigned int FGFDMExec::SRand()` public since it's currently private so `FGWind` currently can't call it.

seanmcleod70:
I've pushed a commit to return `FGFDMExec`'s random seed if one hasn't been assigned to `FGWind` and also an additional test for this.

bcoconni:
Bonus point: this PR is partially addressing #666 by removing some static variables 👍

PR #1254, opened 2025-04-07
Sensor random seed

seanmcleod70:


bcoconni:
For the record, this PR is motivated by https://github.com/JSBSim-Team/jsbsim/discussions/1250#discussioncomment-12669313.

@seanmcleod70, I guess we also need a property `randomseed` for these sensors similar to `simulation/randomseed` as well as `atmosphere/randomseed` from #1253 ? At least, one would need that to reset the simulation to its initial state, no ?!?

seanmcleod70:
The random seed for a particular sensor, i.e. a sensor that you want to control the reproducibility of it's noise is set via a sensor attribute, specific to a specific sensor.

 ```xml
<sensor name="name">
  <input> property </input>
  <lag> number </lag>
  <noise [variation="PERCENT|ABSOLUTE"] [distribution="UNIFORM|GAUSSIAN"] [randomseed="number"] > number </noise>
```

So unlike the wind turbulence case there isn't a single random seed property. 

Not quite following you in terms of "reset the simulation to it's initial state".

In terms of reset are you talking about the case where an FDM is created, it's run for a while, then reset and further time steps, i.e. without creating a new FDM instance?

```python
fdm = create_fdm()
fdm.run()  # N times
fdm.reset()
fdm.run() # N times
```

bcoconni:
> So unlike the wind turbulence case there isn't a single random seed property.

Yep, got that. But my question was rather if a property would not be needed for each sensor that uses this attribute in its XML definition ?
> In terms of reset are you talking about the case where an FDM is created, it's run for a while, then reset and further time steps, i.e. without creating a new FDM instance?

Correct, that was my thought. I was trying to anticipate some cases where having a property would be useful but maybe it's irrelevant ?

seanmcleod70:
I did sort of consider the option of generating a property for the sensor if it had a `<noise/>` element with a `name` attribute along the lines of what is done for `<quantization/>`. 

https://github.com/JSBSim-Team/jsbsim/blob/13008f3045a0c1bf858ec4f5346c09403ed5dd28/src/models/flight_control/FGSensor.cpp#L78

https://github.com/JSBSim-Team/jsbsim/blob/13008f3045a0c1bf858ec4f5346c09403ed5dd28/src/models/flight_control/FGSensor.cpp#L268-L279

But ended up going with the option of the user supplying the random seed value as an attribute of the `<noise/>` element instead.

But thinking about it a bit more, I'm inclined to switch to a scheme matching the `<quantization/>` approach.

seanmcleod70:
I've pushed some commits to switch to generating a sensor specific randomseed property. One of the builds failed due to a disk space issue, but all the other builds built successfully and the tests succeeded.

seanmcleod70:
Looks like, without a subsequent commit the build task that failed due to lack of disk space was re-run and now show all builds/tests are successful/green tick.

src/models/flight_control/FGSensor.h
```diff
@@ -79,7 +80,7 @@ CLASS DOCUMENTATION
 <sensor name="aero/sensor/qbar">
   <input> aero/qbar </input>
   <lag> 0.5 </lag>
-  <noise variation="PERCENT"> 2 </noise>
+  <noise variation="PERCENT" name="qbarsensor"> 2 </noise>
```
bcoconni, line 83: Same as previously, you can remove it from the example.

bcoconni:
> Looks like, without a subsequent commit the build task that failed due to lack of disk space was re-run and now show all builds/tests are successful/green tick.

Yep, that was me ! I pushed the button "Re-run failed jobs" on GitHub and it managed to find a runner with sufficient disk space :wink:

seanmcleod70:
Pushed a commit based on your suggestions. Copilot suggested commit message was pretty accurate so used it as is.

> I pushed the button "Re-run failed jobs" on GitHub

I actually looked for that option, but couldn't find it. Could be because I'm not an 'Owner' of the repo, only a 'Member'.

bcoconni:
PR merged 👍 

> I actually looked for that option, but couldn't find it. Could be because I'm not an 'Owner' of the repo, only a 'Member'.

I have promoted your account to "Owner" (and demoted your old account to "Member" because there is only one Sean 😉)

PR #1255, opened 2025-04-08
Set C++17 for Visual Studio project

seanmcleod70:


gallonmate:
Yes good catch, we need to update the JSBSimForUnreal VS project.

seanmcleod70:
I've pushed a commit to update the JSBSimForUnreal VS project file.

bcoconni:
Thanks for the PR. Since I am using CMake for both Windows and Linux, I am always forgetting to update the MSVC projects, sorry.

PR #1259, opened 2025-04-13
`FGTurbine` does not update the thruster inputs before calling `FGThruster::Calculate`

bcoconni:
As the title says, the bug is silently triggered by the C130 model which has a creative combination of `FGTurbine` engines with `FGPropeller`.
https://github.com/JSBSim-Team/jsbsim/blob/13008f3045a0c1bf858ec4f5346c09403ed5dd28/aircraft/C130/C130.xml#L114-L116
Currently since `FGTurbine` does not call `FGEngine::LoadThrusterInputs` then `FGThruster::in::SoundSpeed` is not initialized.
https://github.com/JSBSim-Team/jsbsim/blob/13008f3045a0c1bf858ec4f5346c09403ed5dd28/src/models/propulsion/FGEngine.cpp#L135-L144
So when the variable `HelicalTipMach` is computed `in.SoundSpeed` contains garbage and is potentially zero which raises a numerical exception.
https://github.com/JSBSim-Team/jsbsim/blob/13008f3045a0c1bf858ec4f5346c09403ed5dd28/src/models/propulsion/FGPropeller.cpp#L225
Fortunately, the C130 model do not use the tables `CP_MACH` and `CT_MACH` hence the bug has no further consequences.

I have found the bug by chance while running the debugger which was halting at the line of code that computes `HelicalTipMach` and was complaining that a numerical exception was occurring. The bug fix is trivial: it just adds a call to `FGEngine::LoadThrusterInputs` before the call to `FGThruster::Calculate`.

### DIFF
--- a/src/models/propulsion/FGTurbine.cpp
+++ b/src/models/propulsion/FGTurbine.cpp
@@ -163,6 +163,7 @@ void FGTurbine::Calculate(void)
     default: thrust = Off();
   }
 
+  LoadThrusterInputs();
   Thruster->Calculate(thrust); // allow thruster to modify thrust (i.e. reversing)
 
   RunPostFunctions();


seanmcleod70:
Saw some interesting things while trying to reproduce this using Visual Studio.

Running a release build under the debugger, it was hit and miss, every 3rd-4th run would stop with a division by 0 exception.

```sh
Exception thrown at 0x00000001400D8F77 in JSBSim.exe: 0xC000008E: Floating-point division by zero (parameters: 0x0000000000000000, 0x0000000000001924).
Unhandled exception at 0x00000001400D8F77 in JSBSim.exe: 0xC000008E: Floating-point division by zero (parameters: 0x0000000000000000, 0x0000000000001924).
```

However running a debug build under the debugger I never saw a division by 0 exception. But did notice a very consistent pattern in the uninitialized structure.

image

Note the value of the uninitialized doubles.

Which is more than likely coming from the debug heap used by the debug build of MSVC.

https://www.nobugs.org/developer/win32/debug_crt_heap.html

Where the debug heap returns heap memory with a specific fill pattern which result in an unusual double value.

So at least for debug MSVC builds we could potentially have some asserts looking for these sorts of patterns of uninitialized doubles etc. in various of the heap based structures in JSBSim.

bcoconni:
@seanmcleod70, based on your investigations, could you confirm that this PR indeed fixes the reported bug ?

seanmcleod70:
Just confirmed that the debug heap in MSVC fills heap memory it returns with `0xCD`, and that 8 bytes of that results in a double with a value of `-6.2774385622041925e+66`, so a good indicator of an uninitialized member. Also the stack seems to also be filled with `0xCD`, so uninitialized local doubles also stand out.

 I tested your proposed fix in both debug and release builds and it does the trick.

bcoconni:
> Just confirmed that the debug heap in MSVC fills heap memory it returns with 0xCD, and that 8 bytes of that results in a double with a value of `-6.2774385622041925e+66`, so a good indicator of an uninitialized member. Also the stack seems to also be filled with 0xCD, so uninitialized local doubles also stand out.

I have tried adding a test before the computation of `HelicalTipMach` on Windows in Debug mode:
```diff
+  char* p = reinterpret_cast<char*>(&in.Soundspeed);
+  assert(*p != 0xCD);
  HelicalTipMach = sqrt(Vtip*Vtip + Vel*Vel) / in.Soundspeed;
```
But the assertion did not fail. Not sure what I did wrong ?
> I tested your proposed fix in both debug and release builds and it does the trick.

Thanks. PR merged !

seanmcleod70:
> But the assertion did not fail. Not sure what I did wrong ?

Hmm, was also stumped for a bit.

```c++
  char* p = reinterpret_cast<char*>(&in.Soundspeed);
  char pp = *p;
  assert(*p != 0xCD);
```

```c++
*p != 0xCD
true
*p
0xcd 'Í'
*p != 0xcd
true
*p == 0xcd
false

pp
0xcd 'Í'
pp != 0xCD
true
pp == 0xcd
false
pp == (char)0xcd
true
```

It's a signed versus unsigned issue.

```c++
0xCD = 1100 1101 (205)
```

image

bcoconni:
Good catch 👍 ! Changing the type to `unsigned char*` fixed the problem.

PR #1262, opened 2025-04-15
Logging redesign - part 5

bcoconni:
Following the PR #1176, this PR continues the migration to the new logging system.

This PR is migrating the complete folder `src/initialization`, the thrusters in `src/models/propulsion` and some further files in `src/input_output`. All these files have a local copy of the `FGFDMExec` instance which makes very easy the access to the logger through the method `FGFDMExec::GetLogger`.

Please note that I have converted some files in `src/initialization` that are no longer compiled in JSBSim (`FGTrimAnalysis*.*` and `FGTrimmer`). I have done that to remove as many occurrences as I could of `cout` and `cerr` so that they no longer appear in the search filters that I am using. I have not tested if these files are compiling after their change because most likely they did not compile before since they are using outdated APIs.

### DIFF
--- a/src/initialization/FGInitialCondition.cpp
+++ b/src/initialization/FGInitialCondition.cpp
@@ -53,6 +53,7 @@ INCLUDES
 #include "FGTrim.h"
 #include "FGFDMExec.h"
 #include "input_output/string_utilities.h"
+#include "input_output/FGLog.h"
 
 using namespace std;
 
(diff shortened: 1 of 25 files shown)


seanmcleod70:
Looks fine. Do you have a rough estimate of what percentage of the source base still needs to be updated?

bcoconni:
> Do you have a rough estimate of what percentage of the source base still needs to be updated?

The branch `release/v1.2` does not include any of the logging patches, so I've used it as the reference. I have counted the `cout` and `cerr` from the `src` directory and excluded those from `src/GeographicLib`, `src/simgear` and `src/utilities`:

||`cout`|`cerr`|
|-|-|-|
|`release/v1.2`|889|370|
|After this PR|293|174|
|% remaining|33%|47%|

So I'd say we have roughly removed 60% of the `cout` and `cerr`'s.

seanmcleod70:
There is an "I" in equipe 😉

bcoconni:
> There is an "I" in equipe 😉

Indeed, but in an *équipe*, I'd say *je* rather than *I*. So "there is no *je* in *équipe*" still holds but you lose the double meaning it has in English.

PR #1266, opened 2025-04-21
Ensure engine set-running works for piston engines

seanmcleod70:
Ensure that if a mixture command isn't specified directly or via a system function a piston engine will start via a running command.

See discussion https://github.com/JSBSim-Team/jsbsim/discussions/1264 for more details.


bcoconni:
PR merged. Thanks :+1:

PR #1270, opened 2025-04-26
Remove the class `FGPropertyNode`

bcoconni:
As per the issue reported in https://github.com/JSBSim-Team/jsbsim/issues/834#issuecomment-2774035954 and the discussion that followed, this PR removes the class `FGPropertyNode` from JSBSim.

Direct calls to `SGPropertyNode` allow to avoid the dubious conversion from `SGPropertyNode*` to `FGPropertyNode*` in methods such as the one below:
https://github.com/JSBSim-Team/jsbsim/blob/e1ea4e85524e209037ec868f5a1383ec2ef3c7a1/src/input_output/FGPropertyManager.cpp#L109-L117

Note that this PR does not remove `FGPropertyNode` from the Python module which keeps its methods unchanged including `FGPropertyNode.get_fully_qualified_name()`. Not sure if renaming it to `SGPropertyNode` and making `get_fully_qualified_name()` a function is a better option ?

### DIFF
--- a/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp
+++ b/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp
@@ -109,7 +109,7 @@ void UJSBSimMovementComponent::CommandConsole(FString Property, FString InValue,
 {
 
   //Property name must be alphanumeric and limited to six []-._/ special characters. This check prevents UE5 editor crash when using invalid characters.
-#if WITH_EDITOR 
+#if WITH_EDITOR
   if (!std::regex_match((TCHAR_TO_UTF8(*Property)), std::regex("^[a-zA-Z0-9\\[\\]\\-._/]+$")))
   {
     FMessageLog("PIE").Error()->AddToken(FTextToken::Create(FText::FromString(FString::Printf(TEXT("%s: JSBSim Command Console Blueprint Node Error: *%s* Property name must be alphanumeric and limited to these []-._/ six characters. Do not use parentheses *(RW)* in your property name"), *this->GetOwner()->GetName(), *Property))));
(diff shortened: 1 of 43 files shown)


gallonmate:
Could you also replace the FGPropertyNode with SGPropertyNode in the Unreal Plugin? Just these two below. The rest of the code should work fine as is. I could submit a PR later but figured it could be included with your PR. Thanks :)
https://github.com/JSBSim-Team/jsbsim/blob/6b81abf2610c3919e2d3893f6f3265e61beb1db5/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp#L121

https://github.com/JSBSim-Team/jsbsim/blob/6b81abf2610c3919e2d3893f6f3265e61beb1db5/UnrealEngine/Plugins/JSBSimFlightDynamicsModel/Source/JSBSimFlightDynamicsModel/Private/JSBSimMovementComponent.cpp#L157

seanmcleod70:
I'm not a great fan of using `auto` a lot. Sure for fairly complicated template based types like this I have no issue.

```c++
map<int,list<string>>::iterator i = m.begin();
auto i = m.begin();
```

But for these sorts of cases when I'm browsing code I'd prefer to see `SGPropertyNode* prop` than `auto prop`. If I want to know whether `prop` is being used appropriately I need to go and take a look at `getNode` (name in this case sort of gives it away) to see what it's returning, is it a `bool` or some node pointer, if so which one `FGPropertyNode*` or `SGPropertyNode*` etc. I think it just makes it more difficult and means more work for the reader.

```c++
    bool HasNode(const std::string& path) const
    {
      std::string newPath = path;
      if (newPath[0] == '-') newPath.erase(0,1);
      auto prop = root->getNode(newPath);
      return prop;
    }
```

Also would prefer `return prop != nullptr` again to make it more explicit to the reader compared to the implicit "type-cast" to `bool`.

seanmcleod70:
Noticed a couple of other places where the dangerous cast up to a `FGPropertyNode*` was happening, e.g. `FGPropertyNode* operator*()`, which you've now fixed. Guessing they weren't code paths that were hit with usban.

bcoconni:
> Could you also replace the FGPropertyNode with SGPropertyNode in the Unreal Plugin?

Yep! Done. Thanks for the heads up @gallonmate :+1: 
> I'm not a great fan of using `auto` a lot. [...] I'd prefer to see `SGPropertyNode* prop` than `auto prop`.

Fair enough I've replaced all the new occurrences of `auto` by explicit type names.
> Also would prefer `return prop != nullptr` again to make it more explicit to the reader compared to the implicit "type-cast" to `bool`.

Done ! Thanks for the comments @seanmcleod70.

bcoconni:
> All looks good.

Thanks. PR merged.

PR #1272, opened 2025-05-02
Maintenance / Minor changes

bcoconni:
This PR brings a number of minor changes and fixes:
* Fix the computation of the constant `minDewPoint` in `FGStandardAtmosphere` where the minus sign in the formula was misplaced.
* Remove the exception handlers around `getenv` because [`getenv` has a no throw guarantee](https://cplusplus.com/reference/cstdlib/getenv).
* Use `std::string` for the buffer of  `FGLogConsole`. We do not need all the features of `std::ostringstream`

### DIFF
--- a/src/FGFDMExec.cpp
+++ b/src/FGFDMExec.cpp
@@ -104,12 +104,10 @@ FGFDMExec::FGFDMExec(FGPropertyManager* root, std::shared_ptr<unsigned int> fdmc
   EnginePath = "engine";
   SystemsPath = "systems";
 
-  try {
-    char* num = getenv("JSBSIM_DEBUG");
-    if (num) debug_lvl = atoi(num); // set debug level
-  } catch (...) {                   // if error set to 1
+  if (const char* num = getenv("JSBSIM_DEBUG"); num != nullptr)
+    debug_lvl = strtol(num, nullptr, 0);
+  else
     debug_lvl = 1;
-  }
 
   if (!FDMctr) {
     FDMctr = std::make_shared<unsigned int>(); // Create and initialize the child FDM counter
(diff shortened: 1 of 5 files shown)


seanmcleod70:
Looks good. I see you're using some new C++17 constructs, `init-statement` in some of the if statements 😉 

I guess no-one noticed the minimum dew point issue, since even the correct minimum dew point temperature is fairly close to absolute zero.

bcoconni:
> Looks good. I see you're using some new C++17 constructs, `init-statement` in some of the if statements 😉

Yep ! Thank you for letting me know about this feature :+1: 
> I guess no-one noticed the minimum dew point issue, since even the correct minimum dew point temperature is fairly close to absolute zero.

Indeed but I figured this out while investigating [this issue from FlightGear](https://gitlab.com/flightgear/flightgear/-/issues/3084#note_2373994335) where users are complaining JSBSim is issuing too many console messages. It was incidental because it does not fix the issue but it is a bug anyway.

In addition, the FlightGear devs are concerned that [I am submitting the fix to FlightGear before it is included in JSBSim](https://gitlab.com/flightgear/flightgear/-/merge_requests/116). Hence this PR which will sync both code bases.

seanmcleod70:
> Yep ! Thank you for letting me know about this feature

Hmm, I'm a bit confused. When I looked at your commit I noticed the extra statements in some of the `if` statements, which I wasn't familiar with, so I took a look at the C++ standard https://en.cppreference.com/w/cpp/language/if and noticed:

`init-statement | - | (since C++17) `

So by using the `init-statment` option in the `if` statements you let me know about the feature, available from C++17, rather than the other way round.

bcoconni:
> you let me know about the feature, available from C++17, rather than the other way round.

No, no you mentioned it first: https://github.com/JSBSim-Team/jsbsim/pull/1094#issuecomment-2120939601 :wink: I remember that because your post made me make a mental note that this syntax could be convenient.

seanmcleod70:
Aaaah, yes, you're right, my memory is fading 😉

PR #1277, opened 2025-05-14
Unreal Plugin: Fix issue in the HUD debug info

gallonmate:
-Fix issues in the HUD debug info, which were preventing compilation for UE5.6 Preview 
-Update the Unreal plugin Readme


bcoconni:
Looks good to me.

gallonmate:
Thanks for checking & merging!

Issue #1279, opened 2025-05-18
JSBSim VTOL/thrust vectoring feature


DuanzzhiWang:
Recently I was working on VTOL feature for V-22 Osprey in jsbsim, and want to test it out inside Unreal Engine. 
From FGThruster.cpp, I assume this is the part that supports thrust vectoring but I didn't see any existing aircraft implenting it:

![Image](https://github.com/user-attachments/assets/303f1598-e1ca-492d-add6-ffaaed69697e)

therefore, I add a line under FGThruster.h  to calculate the thrust vectoring angle:

![Image](https://github.com/user-attachments/assets/de435953-c011-4e79-bc3a-c985620cf93c)

for the aircraft script file, I also delete the orientation part, since the thrust is no longer a fixed value:

![Image](https://github.com/user-attachments/assets/8a0c24c5-0cf1-491f-aa0f-01f48a9f3848)

and added a new channel called thrust vectoring under the FCS part：

![Image](https://github.com/user-attachments/assets/e7c1c319-b666-44ae-91b7-c02a101b86fa)

with these changes, UnrealEngine crashes with following report：

![Image](https://github.com/user-attachments/assets/cb6b5c8c-2eab-49cd-b82f-e9c296f144fe)

So Is there anything I missed while working on the VTOL function? Need some advice plsss

seanmcleod70:
There should be no need to modify the JSBSim source code to implement thrust vectoring.

Here is a simple example of thrust vectoring;

```python
import jsbsim
import math

AIRCRAFT_NAME="737"
# Path to JSBSim files, location of the folders "aircraft", "engines" and "systems"
PATH_TO_JSBSIM_FILES="../.."

# Avoid flooding the console with log messages
#jsbsim.FGJSBBase().debug_lvl = 0

fdm = jsbsim.FGFDMExec(PATH_TO_JSBSIM_FILES)

# Load the aircraft model
fdm.load_model(AIRCRAFT_NAME)

# Set engines running
fdm['propulsion/set-running'] = -1

# Initial conditions
fdm['ic/h-sl-ft'] = 1000
fdm['ic/vc-kts'] = 250
fdm['ic/gamma-deg'] = 0

# Initialize the aircraft with initial conditions
fdm.run_ic() 

# Trim
try:
    fdm['simulation/do_simple_trim'] = 1

except jsbsim.TrimFailureError:
    print("Trim failed....")
    pass  # Ignore trim failure

print('')

[...]

PR #1280, opened 2025-05-18
Add a thrust vectoring python script example

seanmcleod70:
Based on a user asking how to make use of thrust vectoring in JSBSim in - https://github.com/JSBSim-Team/jsbsim/issues/1279


bcoconni:
Nice ! 👍 

Your PR just made me realize that we are not advertising about the Python example scripts in `README.md`. May be you could take the opportunity of your PR to update `README.md` to point to `examples/python` where visitors could find some non trivial examples of what JSBSim is capable of using Python ?

agodemar:
@seanmcleod70 @bcoconni 
If you agree, I can advertise the Python scripting feature of JSBSim on my LinkedIn account. I usually get a lot of impressions there on this kind of topic.

seanmcleod70:
I've pushed a commit with an update to the `README.md`. Feel free to edit it.

@agodemar no problem with you advertising these Python scripting examples on LinkedIn. But rather than just pointing to the `README.md` and/or source code in the repo I think the post would have more impact if you included the example output plots from some/all of the current Python examples.

agodemar:
@seanmcleod70 What about pointing to Jupyter notebooks that render the plots?

seanmcleod70:
Yep, one option is to convert the 4 samples from stand-alone Python scripts to Jupyter notebooks.

agodemar:
@seanmcleod70 Something along the line of these notebooks: [https://github.com/agodemar/WiReS/tree/master/calc](https://github.com/agodemar/WiReS/tree/master/calc)

seanmcleod70:
Okay, I've converted all 4 standalone Python scripts to individual Jupyter notebooks. Easiest way to double-check them is to view the Jupyter notebooks via Github's automatic Jupyter notebook rendering by viewing them in my branch.

https://github.com/seanmcleod70/jsbsim/tree/ThrustVectoringExample/examples/python

I've done the bare minimum to turn them into Jupyter notebooks for now. We could later on make some of them interactive by using ipywidgets to allow a user to use a slider to select for example the airspeed etc. and have the graph update as the slider is moved etc.

The advantage of the Jupyter notebook approach is that users who don't have Python installed, or don't have matplotlib etc. installed can view the code and the generated output simply by viewing them via Github's automatic Jupyter notebook rendering, given there are saved versions of the matplotlib output in the notebooks.

seanmcleod70:
I've pushed another commit which stores the cached matplotlib plot outputs as SVG instead of PNG since the PNGs looked really 'fuzzy/soft'.

Take a look at the notebooks at - https://github.com/seanmcleod70/jsbsim/tree/ThrustVectoringExample/examples/python to see what the plot outputs look like now as SVG renderings.

So @agodemar when you post links now we'll have much better looking plots 😉

README.md
```diff
@@ -193,6 +193,18 @@ print(jsbsim.get_default_root_dir())
 
 A more elaborate example of Python code is [JSBSim.py](https://github.com/JSBSim-Team/jsbsim/blob/master/python/JSBSim.py), the Python equivalent to `JSBSim.exe`.
 
+## Python module examples
+
+The [examples/python](https://github.com/JSBSim-Team/jsbsim/tree/master/examples/python) directory contains a number of example Python based scripts embedded in Jupyter notebooks demonstrating the use of JSBSim to determine and analyse aircraft performance.
+
+- [AoA vs CAS.ipynb](https://github.com/JSBSim-Team/jsbsim/blob/master/examples/python/AoA%20vs%20CAS.ipynb) calculates and plots the AoA versus CAS for level trim for a range in aircraft weight, altitude and cg.
```
bcoconni, line 200: Explain the meaning of the acronyms (AoA = Angle of Attack, CAS=Calibrated Air Speed, cg=center of gravity, etc.)

seanmcleod70: I've pushed a commit spelling out the acronyms. I did sort of assume that users of JSBSim would know these already 😉

bcoconni: I agree with you, but I think we should avoid using acronyms and technical jargon, at least in the README file. Since it's our homepage, it shouldn't intimidate potential users who don't have an aeronautical background — I'm particularly thinking of game developers who might want to use JSBSim with the Unreal Engine.

bcoconni:
Looks good to me.

Issue #1281, opened 2025-05-19
Building python module from source


Dobid:
***

**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**
I'm trying to build the jsbsim python module from source and following the current developer docs.
- I cloned the jsbsim repository `git clone https://github.com/JSBSim-Team/jsbsim.git jsbsim-code`
- Created the build dir `mkdir build && cd build`
- then ``cmake -DINSTALL_JSBSIM_PYTHON_MODULE=ON ..``
- and ``make -j4``
But when I do : ``python python/setup.py install --skip-build``, I have this error:
``
Traceback (most recent call last):
  File "/home/david/sandbox/jsbsim/jsbsim/build/python/setup.py", line 155, in <module>
    for d in os.scandir('jsbsim/aircraft'):
FileNotFoundError: [Errno 2] No such file or directory: 'jsbsim/aircraft'
``
I've tried multiple things like changing the directory in the setup.py script. I've managed to circumvent this error last month, but now I can't find what I did to successfully install the jsbsim module.

**Please tell us about your environment:**
 - OS: Ubuntu 22.04
 - JSBSim version: from latest source on github commit f794e04
 - Language if applicable (C++, Python, Julia, ...) Python 3.9.22

bcoconni:
Direct calls to `setup.py` are now discouraged (see [Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) or ['setup.py install is deprecated' warning shows up every time I open a terminal in VSCode](https://stackoverflow.com/questions/73257839/setup-py-install-is-deprecated-warning-shows-up-every-time-i-open-a-terminal-i)). The correct way to proceed is to use the `build`package:
```bash
> mkdir build && cd build
> cmake -DINSTALL_JSBSIM_PYTHON_MODULE=ON ..
> make -j4
> cd python
> pip install build
> python -m build --wheel
> pip install jsbsim --no-index -f dist
```
Note that:
1. The command `python -m build --wheel` is calling `setup.py` under the hood. `setup.py` will automatically detect that the library has already been built and should only compile and link `_jsbsim.cxx` (which is the code generated by Cython).
2. The wheel package will be built in the directory `build/python/dist` hence the flag `-f dist` that is passed to the last `pip install` command.

Dobid:
[...]

Issue #1285, opened 2025-05-22
Matlab CI action outage


bcoconni:
The Matlab jobs are failing in our CI workflow for a few days:

![image](https://github.com/user-attachments/assets/6304681c-700a-44a0-9546-50bde28e059d)

This is a known issue from the GitHub action `matlab-actions/setup-matlab` (see https://github.com/matlab-actions/setup-matlab/issues/146). The issue is coming from MathWorks and can be monitored at  https://status.mathworks.com/

PR #1286, opened 2025-05-24
[skip ci] Update Python module installation instructions

bcoconni:
Users are encountering issues using the installation instructions from the dev docs (see issue #1281). Actually our docs are outdated since direct calls to `setup.py` are now discouraged (see [Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) or ['setup.py install is deprecated' warning shows up every time I open a terminal in VSCode](https://stackoverflow.com/questions/73257839/setup-py-install-is-deprecated-warning-shows-up-every-time-i-open-a-terminal-i)).

The correct way to proceed is to use the `build`package instead as is done in our CI workflow for more than 3 years (see commit 44f8e511954711dffbcadb6b517de7e0048f9f90).

Please note that the docs have been updated using AI (GitHub Copilot) as I am not so fond of writing docs.


bcoconni:
Also note that I've skipped the CI workflow (by prepending the commit name with `[skip ci]`) since this PR is only updating docs and since the Matlab jobs are still crashing (see issue #1285).

seanmcleod70:
Looks good.

> JSBSim is coded in standard C++11....

Guess we should update this to C++14 while updating the doc.

I need to remember to use `[skip ci]` for document updates, sample updates etc. that don't require any source code building etc. to avoid wasting energy resources.

bcoconni:
> > JSBSim is coded in standard C++11....
> 
> Guess we should update this to C++14 while updating the doc.

Good catch ! We are now using C++17 :wink: since the PR #1251 so I have updated the docs accordingly.

seanmcleod70:
Oops, I just assumed the docs were only 1 revision behind, so assumed we needed to bump it from 11 to 14 😉 But yes, we've moved onto 17 already.

PR #1292, opened 2025-05-30
Upgrade from windows-2019 to windows-latest

seanmcleod70:
From Github:

> The Windows Server 2019 runner image will be fully unsupported by June 30, 2025. To raise awareness of the upcoming removal, we will temporarily fail jobs using Windows Server 2019. Builds that are scheduled to run during the brownout periods will fail. 

Some of our build tasks were already using `windows-latest`.


bcoconni:
Thanks for taking over this change :+1:

I'd prefer that we use `windows-2022` instead of `windows-latest` to make sure that we are compiling on older platforms[^1] (at least as old as GitHub supports). See the exchange we already had on this topic: https://github.com/JSBSim-Team/jsbsim/pull/1146#issuecomment-2308464258

[^1]: At the moment, `windows-latest` is pointing to `windows-2022` (see [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories)) but I guess that at some point it will point to `windows-2025` and we want to decide when we are switching from one version to the other for the reasons exposed above.

seanmcleod70:
Yep, I can switch it to `windows-2022` from `windows-latest`. 

I had taken a look at the different runners and noticed that `windows-latest` was pointing to `windows-2022`, and that `windows-2022` and `windows-2025` both had the same Visual Studio 2022 version installed, plus we were already using `windows-latest` in a handful of other cases in our `.yml`.

bcoconni:
> [...] plus we were already using `windows-latest` in a handful of other cases in our `.yml`.

Yep, there is sort of a logic which is that the tests are run on the *older* version of Windows at the least once in the CI workflow for both MSVC and MinGW. And regarding the build of the Python wheels and the Windows installer, the idea is to link against "*old*" DLLs versions to support users running JSBSim on not-so-recent versions of Windows.

bcoconni:
...and for the other jobs, it does seem to make much difference so using `windows-latest` is allowing to check that we are also running successfully on the bleeding edge version.

seanmcleod70:
Hmm, given I'm not that familiar with the CI build steps, when I was glancing through the file for `windows-xxx` references I saw the following:

https://github.com/JSBSim-Team/jsbsim/blob/6d96ff0cdd2b0b20d0f379df241a92bb9add7826/.github/workflows/cpp-python-build.yml#L544-L551

And so assumed that we were using `windows-latest`, i.e. making use of the VS2022 compiler to build JSBSim and the Python module when packaging up the Python package to upload to PyPI.

Versus using `windows-2019` and therefore VS2019 compiler when building JSBSim on each commit and running all the tests etc.

bcoconni:
> I've pushed a commit that just changes the original `windows-2019` to `windows-2022` and leaves the original `windows-latest` references as they were. Ignore the booboo push in-between, I was in a rush and accidently included a bunch of other local files accidently instead of only staging and pushing the `.yml` file, performed a `git revert` to get rid of them.

No problem PR merged! Thanks.

PR #1293, opened 2025-06-01
[skip ci] Upgrade .vcxproj C++ toolset to VS2022 (v143) to match CI pipeline

seanmcleod70:
The CI pipeline since - https://github.com/JSBSim-Team/jsbsim/commit/d5621b5ffe7bf3c5f17a08eb6b0c4e4083bc786e is using the Visual Studio 2022 C++ toolset. This PR updates the .vcxproj to use the same C++ toolset. 

If a user wants to continue using an older toolset like VS 2019 (v142) they can still do that by changing the toolset option in VS.


bcoconni:
Looks good to me. PR merged.

For the record: this PR is a follow up of PR #1292.

PR #1303, opened 2025-07-14
DRAFT: fix undefined casting behavior through new SGRawValue type

heshpdx:
We are seeing undefined behavior due to the casting of enums as integers, and that comes from calling the casted function pointers to tie property values.

This PR is a proof-of-concept of how to fix the issue using `std::function` with lambdas to wrap the casting. We should figure out if this is the best approach for changing the code. Right now this patch fixes the first error exposed by the LLVM optimizer, but there are 25+ more files of changes we would have to take on to make it clean.

#834

### DIFF
--- a/src/input_output/FGPropertyManager.h
+++ b/src/input_output/FGPropertyManager.h
@@ -270,7 +270,7 @@ class JSBSIM_API FGPropertyManager
       else {
         if (!setter) property->setAttribute(SGPropertyNode::WRITE, false);
         if (!getter) property->setAttribute(SGPropertyNode::READ, false);
-        tied_properties.push_back(property);
+        tied_properties.push_back(PropertyState(property, nullptr));
         if (FGJSBBase::debug_lvl & 0x20) std::cout << name << std::endl;
       }
     }


src/input_output/FGPropertyManager.h
```diff
@@ -253,6 +253,29 @@ class JSBSIM_API FGPropertyManager
...
+      }
+
+      if (!property->tie(SGRawValueCallable<T>(getter, setter),
+                                                       false))
+        std::cerr << "Failed to tie property " << name << " to indexed functions"
+                  << std::endl;
+      else {
+        if (!setter) property->setAttribute(SGPropertyNode::WRITE, false);
+        if (!getter) property->setAttribute(SGPropertyNode::READ, false);
+        tied_properties.push_back(PropertyState(property, nullptr));
```
heshpdx, line 273: This line is causing the segfault in the `TestPlanet` CI test.

The issue has to do with `BindingInstance` in `PropertyState` being set to `nullptr`.   In FGFDMExec.cpp line 796, a special 'Unbind()' assumes that all of the properties under 'atmosphere' have a `BindingInstance` set to that model. This is new code not in the SPEC CPU codebase.

bcoconni:
> Right now this patch fixes the first error exposed by the LLVM optimizer, but there are 25+ more files of changes we would have to take on to make it clean.

This is mostly addressed in the PR #1304 so that we can focus on the actual issue in this PR.

PR #1304, opened 2025-07-15
Fix illegal function pointer conversions

bcoconni:
In preparation to the resolution of the undefined behavior reported in https://github.com/JSBSim-Team/jsbsim/issues/834#issuecomment-3071158683, this PR removes most of the pointer conversions that are currently done to tie properties.

* In most cases, the compiler is able to automatically determine the function pointer types and the conversions have simply been removed.
* In cases where a `nullptr` is given for the getters, the type names in the template need to be explicitly specified and this removes the need for explicit conversions.
* In the remaining cases, this PR replaces the C cast by C++'s `reinterpret_cast<>`. These casts are precisely where the problem reported in issue #834 happens. A further PR will be needed to remove them. In the meantime they can easily be spotted in the source code by doing a simple search of the `reinterpret_cast`.


seanmcleod70:
Hmm, thought I could potentially use `std::is_enum_v<T>` to provide a generic `getValue()` function for all enum types.

```c++
template <typename T, typename = std::enable_if_t<std::is_enum_v<T>>>
inline T getValue<T>(const SGPropertyNode* node) {
  return reinterpret_cast<T>(node->getIntValue());
}
```

But hitting a compiler error.

```sh
1>FGWinds.cpp
1>C:\source\jsbsim\src\models\atmosphere\FGWinds.cpp(76,10): error C2768: 'getValue': illegal use of explicit template arguments
```

bcoconni:
@seanmcleod70 I've just pushed a new commit that implements your ideas about the specialization of `PropertyTraits`. It also includes some modifications from myself to shorten the code.

Let me know what you think about it.

seanmcleod70:
Hmm, so I was looking at your `getValue()` implementation and your comment above: 

> This is working because `props.hxx` does not define the general case

```c++
template<typename T> T getValue(const SGPropertyNode* node)
{
  static_assert(is_enum_v<T>); // Guard against misusing template instantiation.
  return static_cast<T>(node->getIntValue());
}
```

In Visual Studio it looks like `PROPS_STANDALONE` is defined:

image

So we end up with a declaration of:  

```c++
template<typename T> T getValue(const SGPropertyNode* node)
```

But no definition until your definition?

And it looks like in the case that `PROPS_STANDALONE` isn't defined then there is some `boost` specific checks to exclude `enum` types?

seanmcleod70:
Okay, back to `C++17` and using your idea of using `static_assert` I've just tested the following which compiles and links.

```c++
namespace simgear::props {
  //template<> struct PropertyTraits<JSBSim::FGWinds::tType> : public PropertyTraits<int> {};

  //template<typename T> requires std::is_enum_v<T> struct PropertyTraits<T> : public PropertyTraits<int> {};

  template<typename T> struct PropertyTraits : public PropertyTraits<int> {
    static_assert(std::is_enum_v<T>, "PropertyTraits specialization for enum types only");
  };
};
```

To be clear, all my testing so far has simply been to confirm that the code compiles and links successfully.

seanmcleod70:
I've put the following into `FGPropertyManager.h`, i.e. none of this code in `FGWinds.cpp` or `props.hxx` and it compiles and links.

```c++
namespace simgear::props {
  template<typename T> struct PropertyTraits : public PropertyTraits<int> {
    static_assert(std::is_enum_v<T>, "PropertyTraits specialization for enum types only");
  };
}

template<typename T> T getValue(const SGPropertyNode* node)
{
  static_assert(std::is_enum_v<T>, "getValue specialization for enum types only"); // Guard against misusing template instantiation.
  return static_cast<T>(node->getIntValue());
}
```

bcoconni:
> But I switched to putting the code into `FGPropertyManager.h` very soon afterwards I asked the original question.
> [...]
> Plus you even commented on my comment mentioning the move to FGPropertyManager.h a week or so ago 😉

Oops ! Sorry for the confusion.

PR #1309, opened 2025-07-26
add a cell to enable running on Google Colab and related readme change (#1305)

yuki-2000:
Following the discussion #1305 the example ipynb files are fixed to adapt Google colab and some explanation of readme is changed.

Note that
* The google colab link in readme will correctly work after this branch is merged to master, because the google colab icon links to master branch's file.
* Introduction's google colab icon links to the `python module examples` section, not the google colab. I also think  the text link like [this](), but i think the following icon  [Open in Colab]() may attract people more.
* I only add one cell in `Trim Envelope.ipynb`, but the binary image data change. There are no  change of image view.


bcoconni:
> 1. I don't know how this project handles version control, but it should be done when releasing the next version, 1.2.1. If you link to version 1.2 NOW in master branch, I think it will not work correctly on Google Colab because there are no changes to the program (like !git clone or !pip install) to run on Google Colab in v1.2.

Since the Python module is pip-install'ed then Google collab is guaranteed to be using the last stable version (currently v1.2.3). My concern is rather when the files `*.ipynb` in `master` are using features which have not yet been released to the last stable version from PyPI. Since the branch `release/v1.2` contains the code of the last stable version, I'd rather point to the files `*.ipynb` from the branch `release/v1.2`.
> 2. I do not think there are such ways. Rather, even if such a method existed, it would cause confusion or future bugs. In the discussion, we have added `if 'google.colab' in sys.modules:` so that it will not be executed locally.

Fair enough. I was just asking if that would be possible but since the code is documented, I guess it will fit.

yuki-2000:
>Since the Python module is pip-install'ed then Google collab is guaranteed to be using the last stable version (currently v1.2.3). My concern is rather when the files *.ipynb in master are using features which have not yet been released to the last stable version from PyPI. Since the branch release/v1.2 contains the code of the last stable version, I'd rather point to the files *.ipynb from the branch release/v1.2

I understand what you consider and I understand it.
Then, what should I do next? 
Should I make another merge pull request to release/v1.2, changing the link? (I do not know I can do it) or after change the code here and it is merged, the master branch will be merged to release/1.2?

Actually, this is my first pull request to official project, so please give me an instruction.

seanmcleod70:
> Is there a way to hide but yet run the first cell ? I mean it contains boiler plate code which is not of much interest for someone who would like to run a quick test of JSBSim in a sandbox.

Not sure off-hand if that's really possible with a Jupyter notebook. I'm pretty sure there are mechanisms to flag specific cells, such that when you run an export, e.g. to html or PDF of the Jupyter notebook that the export code will omit flagged cells, generally used to create slides, presentations etc. from the Jupyter notebook.

Personally I've actually switched to using [marimo](https://marimo.io/) notebooks which offers a fair bit more flexibility. In particular it makes use of a DAG to determine the order in which cells need to be run, and does so reactively. Which means that you can place these sorts of boiler plate cells at the end of the notebook, so the user only sees them once they scroll right down to the bottom. Plus you can tag code cells as 'hidden' in which case they're mostly hidden, just shows the first line of the code cell but with dimmed text, and a button to allow you to expand and show the code.

Here is an example from one of my marimo notebooks.

https://seanmcleod70.github.io/FlightDynamicsCalcs/ClimbPerformance.html

The marimo notebooks (plain python files) are available here if you want to take a look.

https://github.com/seanmcleod70/FlightDynamicsCalcs

Having plain python files as the basis of the marimo notebook without cached results in it is touted as an advantage. But a minor disadvantage I find is that it isn't as easy to provide a rendered version for someone who just wants to browse/read and doesn't want to have to run the notebook themselves.

Hence for each marimo notebook, when I commit it to my repo I also include an html exported version which is then available via Github pages.

agodemar:
The PR has been approved. Do we merge this?

seanmcleod70:
@yuki-2000 thanks for the submission which has now been merged into the master branch.

PR #1312, opened 2025-07-30
Add raw values enumeration to fix segfaults

heshpdx:
After the fine work done in #1304 and #1311, my colleague is still getting a segfault when building with `clang -flto=full -fvirtual-function-elimination -fvisibility=hidden`.

We believe the issue is from how the code is converting from a generic type to an Int type for Enum use. In this case, a casting occurs, changing the underlying object type, and thus swapping from an object type for which the vtable has been optimized away to a type that expects to have a vtable.

This PR shows one way to work around the issue. With this patch, we confirm that the JSBSim HEAD successfully runs on each of the CPUv8 inputs without segfaulting. Please let us know what you think. Thanks!

#834

### DIFF
--- a/src/input_output/FGPropertyManager.h
+++ b/src/input_output/FGPropertyManager.h
@@ -62,6 +62,37 @@ template<typename T> T getValue(const SGPropertyNode* node)
   return static_cast<T>(node->getIntValue());
 }
 
+template <class C, class T>
+class SGRawValueMethodsEnum : public SGRawValue<int>
+{
+public:
+  typedef T(C::* getter_t)() const;
+  typedef void (C::* setter_t)(T);
+  SGRawValueMethodsEnum(C& obj,
+    getter_t getter = 0, setter_t setter = 0)
+    : _obj(obj), _getter(getter), _setter(setter) {
+  }
+  virtual ~SGRawValueMethodsEnum() {}
+  virtual int getValue() const {
+    if (_getter) { return (int)(_obj.*_getter)(); }
+    else { return SGRawValue<int>::DefaultValue(); }
+  }
+  virtual bool setValue(int value) {
+      return this->setValue((T)value);
+  }
+  bool setValue(T value) {
+    if (_setter) { (_obj.*_setter)(value); return true; }
+    else return false;
+  }
+  virtual SGRaw* clone() const {
+    return new SGRawValueMethodsEnum(_obj, _getter, _setter);
+  }
+private:
+  C& _obj;
+  getter_t _getter;
+  setter_t _setter;
+};
+
 template <class C, class T, class U>
 class SGRawValueMethodsIndexedEnum : public SGRawValue<T>
 {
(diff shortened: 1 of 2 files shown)


seanmcleod70:
Actually not a member method returning an enum, rather it's a member field of type enum?

bramoore:
The segfault comes during `untie()`, where the value gets resolved.

```c++
inline int
SGPropertyNode::get_int () const
{
  if (_tied)
      return (static_Cast<SGRawValue<int>*>(_value.val)->getValue();
  else
    return _local_val.int_val;
}
```

From what I can tell, LLVM is optimizing away the vtable from the actual `SGRawValueMethods` type.  Casting to a `SGRawValue<int>` parent type is not the same, and expects a vtable.   Making an `SGRawValueMethodsEnum<>` that inherits from `SGRawValue<int>` gets the compiler into the place that it understands that these types are connected...  Deriving from `SGRawValue<Enum>` is not the same as `SGRawValue<int>`.

As for the other calls, I believe it is also coming back to the same root cause, but the cast of the pointer from `Enum*` to `int*` is sufficient to avoid the issue.

bcoconni:
I have to say that I am not comfortable with this PR.

Unless proven mistaken, the code in `SGPropertyNode::get_int` is legal C++ code. So we are in a completely different ballpark than PRs #1304 and #1311. And I am struggling to see why we should change JSBSim code to accomodate extremely aggressive optimizations from LLVM which make legal C++ code fail ?

src/input_output/FGPropertyManager.h
```diff
@@ -62,6 +62,37 @@ template<typename T> T getValue(const SGPropertyNode* node)
...
+    return new SGRawValueMethodsEnum(_obj, _getter, _setter);
+  }
+private:
+  C& _obj;
+  getter_t _getter;
+  setter_t _setter;
+};
+
 template <class C, class T, class U>
 class SGRawValueMethodsIndexedEnum : public SGRawValue<T>
```
bcoconni, line 97: Would not the class `SGRawValueMethodsIndexedEnum` need the same mechanism ?
```suggestion
class SGRawValueMethodsIndexedEnum : public SGRawValue<int>
```

bcoconni (commented):
Unless I have misunderstood the purpose of the patch, the class `SGRawValueMethodsIndexedEnum` also needs to inherit from `SGRawValue<int>` ?

Also I am unsure why `SGRawValueMethodsEnum::getValue` returns an `int` while `SGRawValueMethodsIndexedEnum::getValue` is returning the type `T`?

src/input_output/FGPropertyManager.h
bramoore: Only if the data being returned is an Enum type (type `T`). When using the Enum only as an index type (type `U`), the inheritance is still appropriate to come from `SGRawValue<T>`

bcoconni:
And I'm wondering if using `dynamic_cast<>` in `SGPropertyNode::get_int` and its sibling would not be some good practice.
https://github.com/JSBSim-Team/jsbsim/blob/5200260091fce99b5f7a124bbf1dc30f21bc4bfd/src/simgear/props/props.cxx#L567-L574
I guess that `dynamic_cast<SGRawValue<int>*>(_value.val)` would return `void` for the case at stake ? So something along the lines below would help intercepting any future similar occurrence ?
```c++
   if (_tied) {
     auto raw_value = dynamic_cast<SGRawValue<int>*>(_value.val);
     assert(raw_value); // Fails when compiled with NDEBUG.
     return raw_value->getValue();
   }
   else 
     return _local_val.int_val; 
```
This change would be implemented on our copy of SimGear only but this would help making sure that future changes are not introducing similar issues.

seanmcleod70:
Yep, `static_cast<>` is literally 3 `mov` operations, whereas 

```sh 
       auto raw_value = static_cast<SGRawValue<int>*>(_value.val);
00000001402AB0CA  mov         rax,qword ptr [this]  
00000001402AB0CF  mov         rax,qword ptr [rax+98h]  
00000001402AB0D6  mov         qword ptr [rsp+20h],rax 
        return raw_value->getValue();
```

Whereas `dynamic_cast` calls into the C++ runtime's dynamic cast function.

```sh
        auto raw_value = dynamic_cast<SGRawValue<int>*>(_value.val);
00000001402AB0CA  mov         dword ptr [rsp+20h],0  
00000001402AB0D2  lea         r9,[SGRawValue<int> `RTTI Type Descriptor' (01404AE258h)]  
00000001402AB0D9  lea         r8,[SGRaw `RTTI Type Descriptor' (01404AD8F8h)]  
00000001402AB0E0  xor         edx,edx  
00000001402AB0E2  mov         rax,qword ptr [this]  
00000001402AB0E7  mov         rcx,qword ptr [rax+98h]  
00000001402AB0EE  call        __RTDynamicCast (0140365CB2h)  
00000001402AB0F3  mov         qword ptr [rsp+30h],rax  
        return raw_value->getValue();
```

Which is at least another 2 function calls deep.

image

So maybe we should only ifdef the `dynamic_cast` for debug builds, and make sure we include some debug builds that run the existing test cases etc.

bramoore:
Checking in to see what the next steps are.  Is there anything more you need from us?

seanmcleod70:
I think we should go-ahead with this PR, but leave the `dynamic_cast` suggestion until later, to decide whether to use it at all, or only in debug builds, do some performance testing on it etc. @bcoconni your thoughts?

bcoconni:
> I think we should go-ahead with this PR, but leave the dynamic_cast suggestion until later, to decide whether to use it at all, or only in debug builds, do some performance testing on it etc. 

Yep ! I was about to say the same thing. So I 've pushed the "merge" button.

PR #1314, opened 2025-08-03
Add detailed comments for python example files (#1308)

yuki-2000:
Hello.

I added comments for beginners to the example files in  [examples/python](https://github.com/JSBSim-Team/jsbsim/tree/master/examples/python), except for [AoA vs CAS.ipynb](https://github.com/JSBSim-Team/jsbsim/blob/master/examples/python/AoA%20vs%20CAS.ipynb) because @agodemar just has improved it.

As a beginner of JSBSim, I added comments about what each variable I didn't understand meant, what the unit system was, jsbsim's unique settings, and the general flow of the whole process.
This time, I used the original comments as much as possible and did not change the code itself at all.

I apologize if @agodemar  has already working on these 3 files.

Also, I found a few bugs in the code, so I will post them to issues later.


examples/python/Rudder Kick.ipynb
```diff
@@ -1512,89 +1518,119 @@
...
+    "fdm['ic/beta-deg'] = 0    #sideslip angle (deg)\n",
     "\n",
     "# Initialize the aircraft with initial conditions\n",
     "fdm.run_ic() \n",
     "\n",
-    "# Trim\n",
+    "# Attempt to trim the aircraft.\n",
     "try:\n",
-    "    fdm['simulation/do_simple_trim'] = 1\n",
+    "    # 1 means zero accelerations and constant attitude, usually for straight-and-level flight\n",
```
seanmcleod70, line 1568: Mostly correct, but 'constant attitude' isn't a requirement. The angular accelerations, p-dot, q-dot, r-dot need to be 0, but the angular velocities, p, q, r don't need to be 0. So you can trim with a specific roll rate, e.g. p=5deg/s.

See for more details, 1 == tFull:

https://github.com/JSBSim-Team/jsbsim/blob/30a7261d0406ec152a5d81a2c35bb6ea61d9087c/src/initialization/FGTrim.h#L67-L92

And

https://github.com/JSBSim-Team/jsbsim/blob/30a7261d0406ec152a5d81a2c35bb6ea61d9087c/src/initialization/FGInitialCondition.h#L215-L217

examples/python/Thrust Vectoring Analysis.ipynb
```diff
@@ -3528,42 +3537,62 @@
     "\n",
     "\n",
     "def thrust_vector_range_test(altitude, speed, flight_path_angle, title):\n",
+    "    # altitude: altitude above sea level (ft)\n",
+    "    # speed: mach number of speed (<1)\n",
```
seanmcleod70, line 3541: Detail that speed can be Mach number if it's < 1 as you have, but also mention that > 1 specific KCAS (Knots Calibrated Air Speed).

examples/python/Thrust Vectoring Analysis.ipynb
```diff
@@ -3528,42 +3537,62 @@
...
+    "            # Record the simulation data.\n",
+    "            # Append the angle of attack to the result storage.\n",
     "            alphas.append(fdm[\"aero/alpha-deg\"])\n",
-    "\n",
-    "            thrust = fdm[\"propulsion/engine[0]/thrust-lbs\"]*2\n",
+    "            # Append the thrust to the result storage.\n",
+    "            thrust = fdm[\"propulsion/engine[0]/thrust-lbs\"]*2  # because there are two engines\n",
     "            thrusts.append(thrust)\n",
+    "\n",
+    "            # update the minimum thrust and thrust vectoring angles.\n",
```
seanmcleod70, line 3595: Start sentence with capital letter, i.e. # Update the....

examples/python/Trim Envelope.ipynb
```diff
@@ -77,67 +83,84 @@
...
+    "fdm['aero/alpha-max-rad'] = math.radians(12)   # Maximum angle of attack in radians.\n",
+    "fdm['aero/alpha-min-rad'] = math.radians(-4.0) # Minimum angle of attack in radians.\n",
     "\n",
     "# Set envelope limits\n",
-    "min_speed = 120\n",
-    "max_speed = 460\n",
-    "altitude = 15000\n",
-    "min_gamma = -10\n",
-    "max_gamma = 10\n",
+    "min_speed = 120   # Set the minimum airspeed (kts).\n",
```
seanmcleod70, line 95: Mention for the min and max speeds that it's CAS

seanmcleod70:
Thanks @yuki-2000, overall looks good, have added a couple of minor comments.

examples/python/Rudder Kick.ipynb
yuki-2000: Thank you for the comment. I misunderstand the trim.
I understand that trim itself means stay with initial conditions, and trim mode means which acceleration component  should be zero. Is this correct?

The comment should be `# 1 means zero recti-linear and angular accelerations, usually for straight-and-level flight`?
(while level have to be set by initial conditions.)

I am confused.
Someday, I would like to create examples of changeable property lists, trim mode lists, optimal usage situations, and so on.

seanmcleod70: I think there is often a common misconception that trim implies that the angular velocities need to all be 0. Just like the translational velocities don't need to be 0, if they did have to be 0 then the only trim solutions would be stationary trim solutions on the ground, there is no requirement for the angular velocities to be 0, they can be 0, but they don't have to be, the only constraint for both the translational velocities and angular velocities is that they be constant, i.e. that their accelerations are 0.

Hence my comment that your original comment claiming that trim **requires** a constant attitude isn't accurate.

Take a look at the following paper for some more details on trim.

[A General Solution to the Aircraft Trim Problem](http://wpage.unina.it/agodemar/DSV-DQV/AIAA-2007-6703-905_DeMarco_Duke_Berndt.pdf)

image

Note that there are multiple ways to solve the trim problem, the one described in the paper I linked to is not currently available in JSBSim. The current trim algorithm in JSBSim is based on a different algorithm.

seanmcleod70: You can thank @agodemar regarding the paper since he's one of the authors and also a JSBSim developer 😉 

> Also, I understand that level flight was achieved if the user set fdm['ic/h-sl-ft'].

No, what will indicate that you want level flight is setting gamma $\gamma$, flight path angle to 0.

`fdm['ic/gamma-deg'] = gamma`

You need to set `fdm[ic/h-sl-ft']` to some value even if you're requesting a trim that involves climbing or descending flight. So setting it doesn't request level (non changing altitude) trim. Gamma controls that, pretty sure it defaults to 0 if you don't explicitly set it.

1 - tFull - Longitudinal and lateral trim.

agodemar: @yuki-2000 yes, trimming a dynamical system is a subtle subject, especially when the system is a fliying vehicle in atmosphere.
One example of subtlelty is this: 
When you trim for a flight along a rectilinear trajectory with a nonzero flight path angle ($\gamma\ne0$), you want the true airspeed to be maintained constant with a constant regulation of aerodynamic control surfaces. This my be approximately true for a limited time interval. If you let the state propagate, you'll realize that the altitude is changing (because of nonzero $gamma$), hence air density $\rho$ is changing, hence external forces and moments are changing. 

So, rigourously, to keep a steady airspeed magnitude (and steady angle of attack $\alpha$ and a steady sideslip $\beta$) along a climbing or descending trajectory, one should continuously change flight control deflections and throttle setting. This is why sometimes you ear the term "instantaneous trim".
A similar reasoning is valid for a pull-up condition.

seanmcleod70: You run into this issue of the trim only being valid in the near term even for level trim cases, i.e. $\gamma = 0$. For example given fuel burn both the mass and the moments of inertia will be changing after the trim has been computed.

Now 'near term' is larger for this case than the climbing/descending case.

JSBSim offers a property to prevent the fuel burn affecting the mass and moments of inertia.

`propulsion/fuel_freeze`

yuki-2000:
I fix the comments based of above discussions.
Also the small bug #1315 is also fixed.

examples/python/Rudder Kick.ipynb
seanmcleod70: In effect yes, either the pilot if they are hand-flying or the autopilot when it's engaged, which for commercial airliners is for a large percentage of the overall flight.

So for example the autopilot would be configured to maintain some altitude, e.g. 33,000ft at some speed, e.g. Mach 0.82 and it will then continuously modify the control surfaces and the engine throttles to maintain that.

Out of interest, what is your background and what are you using JSBSim for?

seanmcleod70:
Yep, don't worry about that 1 particular failure. As you said you only modified .ipynb files, but by default any commit kicks off the full CI in terms of rebuilding everything and testing it etc.

If you know for sure your commits don't modify any files that require a build to be kicked off, e.g. editing documentation like the README or these .ipynb files then you can add a [skip ci] label to the commit description to prevent the rebuild etc.

yuki-2000:
Thank you very much.

dershow:
One of my main jobs as a flight instruction is to keep repeating “pitch, power, trim” over and over and over again.  ;-)

Followed by “aviate, navigate, communicate”.



--Adam

seanmcleod70:
> Well, this is redundant information from myself 😄 I've read your message too fast @seanmcleod70 

No problem, I've done that myself a couple of times with some of your comments 😉

PR #1317, opened 2025-08-09
Clean up the properties code and `Tie` methods.

bcoconni:
This PR is a follow up of the PRs #1304, #1311 and #1312.

* The function `getValue` in `FGPropertyManager.h` as well as the `PropertyTraits` conversions in `FGWinds.cpp` and `FGPropagate.cpp` are no longer needed due to the introduction of the class `SGRawValueMethodsEnum`. This PR removes the corresponding code.
* Some clean up in the classes `SGRawValueMethodsEnum` and `SGRawValueMethodsIndexedEnum`:
  * Remove the empty destructor
  * Use `override` instead of repeating the `virtual` keyword from the mother class.
  * Replace `0` by `nullptr` for pointers.
* Use the shorter syntax `std::enable_if_t<>` instead of `std::enable_if<>::type`
* In the method `Tie<C, T, U>`  replace the `static_assert` by the template `std::enable_if_t<>`, similarly to what has been done by the PR #1312
* Use `dynamic_cast` in the method `SGPropertyNode::get_int` and it siblings to check that no illegal conversions are made, as per https://github.com/JSBSim-Team/jsbsim/pull/1312#issuecomment-3144705913
  * The usage of `dynamic_cast<>` is limited to compilations where `NDEBUG` is not defined. Otherwise `static_cast<>` is used for performance reasons.
* Fixed the order in which the members are initialized in the `SGPropertyNode` constructors to avoid some warnings from the compilers (with gcc flag `-Wreorder`).

### DIFF
--- a/src/input_output/FGPropertyManager.h
+++ b/src/input_output/FGPropertyManager.h
@@ -56,12 +56,6 @@ INCLUDES
 FORWARD DECLARATIONS
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
-template<typename T> T getValue(const SGPropertyNode* node)
-{
-  static_assert(std::is_enum_v<T>, "PropertyTraits specialization for enum types only");
-  return static_cast<T>(node->getIntValue());
-}
-
 template <class C, class T>
 class SGRawValueMethodsEnum : public SGRawValue<int>
 {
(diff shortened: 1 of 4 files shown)


seanmcleod70:
Looks good.

The `NDEBUG` define with it's negative logic always make me double check it's use 😉

bcoconni:
> The `NDEBUG` define with it's negative logic always make me double check it's use 😉

The same with me :smile:

BTW does MSVC define the flag `NDEBUG` when compiling with `RelWithDebInfo` ? I am asking because the Python wheels that are uploaded to PyPI are currently compiled with `RelWithDebInfo`:
https://github.com/JSBSim-Team/jsbsim/blob/b005a91d3151624eb26ae5595570bd188ef49702/.github/workflows/cpp-python-build.yml#L673
I am concerned that if it does not define `NDEBUG` then the Windows version of the Python wheels will use `dynamic_cast<>` instead of `static_cast<>`.

seanmcleod70:
Hmm, so this details CMAKE's Build Types.

https://gist.github.com/MangaD/475b8b413aff7682b803fb007083fb5c

Now it only lists compiler flages for GCC/Clang, and interestingly it explicitly defines `NDEBUG` for the release builds, i.e. it doesn't appear to rely on the compiler itself to define it?

So we need to figure out what compiler flags are passed to MSVC for CMAKE's `RelWithDebInfo`.

bcoconni:
So are we ready to merge this PR ?

I will address the `NDEBUG` flag for Windows in a separate PR.

seanmcleod70:
> So are we ready to merge this PR ?

Yep, I thought the only outstanding thing was the potential issue with `NDEBUG` on MSVC builds. Which you mention now you'll address in a separate PR.

bcoconni:
> Yep, I thought the only outstanding thing was the potential issue with `NDEBUG` on MSVC builds. Which you mention now you'll address in a separate PR.

Indeed, as you mentioned, we are already setting it up correctly in the MSVC project files and, according to GitHub Copilot, `NDEBUG` is defined by default for CMake when buidling with `RelWithDebInfo` and `Release`. But well that's AI which may or may not be trusted. So I guess we need to specify explicitly the flag in `CMakeLists.txt` for non-debug builds to be on the safe side, and also to protect ourselves against a change of CMake defaults.

seanmcleod70:
Yep, setting `NDEBUG` ourselves explicitly in `CMakeLists.txt` is the safest option, at worst it will be defined twice in the arguments passed to the compiler.

When I initially took a look I thought I'd be able to fairly easily find a reference to some config file in the CMake installation listing the compiler options per compiler per build type, but wasn't able to at the time until I ran out of steam.

I did come across this link which mentions how you can get CMake to output the compiler options.

https://stackoverflow.com/questions/24767450/using-cmake-how-do-you-determine-the-default-compiler-flags-per-build-type

PR #1321, opened 2025-08-15
Add NEU (North East Up) properties for distance from IC start

seanmcleod70:
Based on discussion https://github.com/JSBSim-Team/jsbsim/discussions/1318 in terms of a request to be able to easily plot the coordinates of an aircraft in the local frame to show a 2D or 3D plot of the aircraft's trajectory I've added 3 new properties in `FGAuxiliary`.

```
position/neu-n-ft
position/neu-e-ft
position/neu-u-ft
```

They output the current offset of the aircraft in a local frame with an origin based on the initial condition's lat, lon and 0 alt relative to the geoid ellipsoid.

### DIFF
--- a/src/models/FGAuxiliary.cpp
+++ b/src/models/FGAuxiliary.cpp
@@ -185,14 +185,14 @@ bool FGAuxiliary::Run(bool Holding)
   vMachUVW(eW) = vAeroUVW(eW) / in.SoundSpeed;
 
   // Position tracking in local frame with local frame origin at lat, lon of initial condition
-  // and at 0 altitude relative to the geoid ellipsoid. Position is NEU (North, East, UP) in feet.
+  // and at 0 altitude relative to the reference ellipsoid. Position is NEU (North, East, UP) in feet.
   if (!NEUFromStartInitialized) {
     NEUStartLocation = FDMExec->GetIC()->GetPosition();
     NEUStartLocation.SetPositionGeodetic(NEUStartLocation.GetLongitude(), NEUStartLocation.GetGeodLatitudeRad(), 0.0);
     NEUFromStartInitialized = true;
   }
   vNEUFromStart = NEUStartLocation.LocationToLocal(in.vLocation);
-  vNEUFromStart(3) *= -1.0;  // Flip sign for Up, so + for altitude above geoid
+  vNEUFromStart(3) *= -1.0;  // Flip sign for Up, so + for altitude above reference ellipsoid
 
   Vground = sqrt( in.vVel(eNorth)*in.vVel(eNorth) + in.vVel(eEast)*in.vVel(eEast) );
 
@@ -367,7 +367,7 @@ double FGAuxiliary::GetNlf(void) const
 double FGAuxiliary::GetLongitudeRelativePosition(void) const
 {
   return in.vLocation.GetDistanceTo(FDMExec->GetIC()->GetLongitudeRadIC(),
-                                    in.vLocation.GetGeodLatitudeRad())* fttom;
+                                    in.vLocation.GetGeodLatitudeRad());
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -375,7 +375,7 @@ double FGAuxiliary::GetLongitudeRelativePosition(void) const
 double FGAuxiliary::GetLatitudeRelativePosition(void) const
 {
   return in.vLocation.GetDistanceTo(in.vLocation.GetLongitude(),
-                                    FDMExec->GetIC()->GetGeodLatitudeRadIC())* fttom;
+                                    FDMExec->GetIC()->GetGeodLatitudeRadIC());
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -384,7 +384,7 @@ double FGAuxiliary::GetDistanceRelativePosition(void) const
 {
   auto ic = FDMExec->GetIC();
   return in.vLocation.GetDistanceTo(ic->GetLongitudeRadIC(),
-                                    ic->GetGeodLatitudeRadIC())* fttom;
+                                    ic->GetGeodLatitudeRadIC());
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/src/models/FGAuxiliary.cpp
+++ b/src/models/FGAuxiliary.cpp
@@ -83,6 +83,8 @@ FGAuxiliary::FGAuxiliary(FGFDMExec* fdmex) : FGModel(fdmex)
   vAeroPQR.InitMatrix();
   vMachUVW.InitMatrix();
   vEulerRates.InitMatrix();
+  vNEUFromStart.InitMatrix();
+  NEUFromStartInitialized = false;
 
   bind();
 
@@ -116,6 +118,8 @@ bool FGAuxiliary::InitModel(void)
   vAeroPQR.InitMatrix();
   vMachUVW.InitMatrix();
   vEulerRates.InitMatrix();
+  vNEUFromStart.InitMatrix();
+  NEUFromStartInitialized = false;
 
   return true;
 }
@@ -180,7 +184,15 @@ bool FGAuxiliary::Run(bool Holding)
   vMachUVW(eV) = vAeroUVW(eV) / in.SoundSpeed;
   vMachUVW(eW) = vAeroUVW(eW) / in.SoundSpeed;
 
-  // Position
+  // Position tracking in local frame with local frame origin at lat, lon of initial condition
+  // and at 0 altitude relative to the geoid ellipsoid. Position is NEU (North, East, UP) in feet.
+  if (!NEUFromStartInitialized) {
+    NEUStartLocation = FDMExec->GetIC()->GetPosition();
+    NEUStartLocation.SetPositionGeodetic(NEUStartLocation.GetLongitude(), NEUStartLocation.GetGeodLatitudeRad(), 0.0);
+    NEUFromStartInitialized = true;
+  }
+  vNEUFromStart = NEUStartLocation.LocationToLocal(in.vLocation);
+  vNEUFromStart(3) *= -1.0;  // Flip sign for Up, so + for altitude above geoid
 
   Vground = sqrt( in.vVel(eNorth)*in.vVel(eNorth) + in.vVel(eEast)*in.vVel(eEast) );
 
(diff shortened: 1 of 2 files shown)


seanmcleod70:
In terms of a naming convention for these properties I did notice that @agodemar had in some code he referenced in the issue used the naming convention of NEA - North East Altitude, so Altitude instead of Up. May be a better option in terms of naming?

bcoconni:
This is looking redundant with the properties `position/distance-from-start-lon-mt` (corresponding to the East direction), `position/distance-from-start-lat-mt` (corresponding to the North direction) and `position/h-sl-meters` (Up direction). In addition, the computations that are made in this PR are only making sense when you stay in the vicinity of the IC start (i.e. the Earth curvature is not accounted for so you may for example obtain negative altitudes while flying well above the ground).

I have seen from #1318 that the properties that I have listed above have been chosen to display the 3D trajectory so not sure what the point of this PR is ?

seanmcleod70:
In terms of some more context, often textbooks on flight dynamics include linearized models like $\dot{x} = Ax + Bu$ and for example as shown in @yuki-2000's notebook:

https://colab.research.google.com/drive/1qv5Hll_v4UWcC6YVJDv2KYifhyz7vU5Q?usp=sharing

Most of these linearized models assume a flat earth and generate trajectory output relative to a starting point which is then plotted in 2D or 3D to show the trajectory. 

So the idea was to provide the same trajectory coordinates from a JSBSim run that could be used in the same plot or a comparison plot etc.

@agodemar mentions the following in https://github.com/agodemar/WiReS/blob/master/calc/02_Tutorial02_ComplexManeuver-UnitConversion-Trajectory.ipynb 

> In order to correctly visualize the trajectory, we must first convert the geocentric position time history to a more intuitive reference frame. A classic choice would be the NED (North, East, Down) reference frame, but since the aircraft is of course flying above the ground, it would result in negative coordinates along the Z axis. Therefore, we decided to represent the trajectory in terms of North, East coordinates with respect to the starting point of the simulation, and altitude above the ground.

> Converts the time histories of geocentric latitude, longitude and altitude above Sea Level
    to the time histories of North, East and altitude. The latter are expressed in a reference 
    frame whose origin is in the initial position (L,l) of the simulation, but at Mean Sea Level altitude.
    Earth is considered flat.

The `position/distance-from-start-lon-mt` and `position/distance-from-start-lat-mt` properties output a geodesic based distance. 

So after @yuki-2000 asked about 3D cartesian coordinates to use, and I noticed that the `position/distance-from-start-` properties were geodesic based and noticed that @agodemar was post-processing ECEF coordinates to create local frame coordinates I thought it would be useful to provide them directly from JSBSim.

Now, I guess it may be the case that given the small distances for these sorts of trajectories that there isn't really much difference versus the geodesic based properties.

You mentioned:

> the computations that are made in this PR are only making sense when you stay in the vicinity of the IC start

Correct, but anyone using a local tangential frame to the earth should know this :wink: The trajectories typically rendered using a flat local frame aren't long distance and definitely not intercontinental trajectories from say Paris to Tokyo. They're typically used to render short term maneuvers.

seanmcleod70:
@agodemar I meant to ask, was there a particular reason you didn't make use of the `position/distance-from-start-lon-mt` and `position/distance-from-start-lat-mt` properties?

> Now, I guess it may be the case that given the small distances for these sorts of trajectories that there isn't really much difference versus the geodesic based properties.

One quick test, using the 737 cruise script, after 100s, note I converted the `position/distance-from-start-lon-mt` and `position/distance-from-start-lat-mt` properties to feet.

```sh
    position/neu-n-ft = 64761.025060
    position/distance-from-start-lat-mt = 64667.549213

    position/neu-e-ft = 37390.086006
    position/distance-from-start-lon-mt = 37336.438030
```

seanmcleod70:
@yuki-2000 no problem with you jumping in and commenting. It was your original question/discussion that started this.

So, just so that we're on the same plane/page. 

image

So, this PR adds properties so that a trajectory relative to a local NEU plane like the local NED plane in green shown above can be easily retrieved from JSBSim.

Now there are existing properties, i.e. `position/distance-from-start-lon-mt` and `position/distance-from-start-lat-mt` that can be used, so the following diagram tries to show some of the differences.

$NEU_0$ is the origin of a local NEU plane with $P0$ representing the starting/initial conditions location of the aircraft and $P1$ indicating the aircraft's position some time later. In this example the aircraft starts off at a (lat, lon) of (0,0) and flies due north.

image

So `position/distance-from-start-lat-mt` would return the distance of the geodesic from $NEU_0$ to $LL1$. Whereas the new property `position/neu-n-ft` would be the distance from $NEU_0$ to $N1$.

In the $NEU$ plane the altitude at $P1$ would be the distance from $N1$ to $P1$, whereas the geod altitude would be the distance from $LL1$ to $P1$.

Using the 737 cruise script as a basis, here are some figures on what the differences look like. So I've started the 737 off at a (lat, lon) of (0,0) and have it flying due north. Note, all values are in feet.

```sh
Time: 0
    position/neu-n-ft = 0.000000
    position/neu-e-ft = 0.000000
    position/neu-u-ft = 30000.000000
    position/distance-from-start-lat-mt = 0.000000
    position/distance-from-start-lon-mt = 0.000000
    position/geod-alt-ft = 30000.000000
    position/h-sl-ft = 30000.000000

Time: 5
    position/neu-n-ft = 3674.827387
    position/neu-e-ft = -0.000005
    position/neu-u-ft = 30000.032050
    position/distance-from-start-lat-mt = 3669.531074
    position/distance-from-start-lon-mt = 0.000005
    position/geod-alt-ft = 30000.356431
    position/h-sl-ft = 30000.356431

Time 10:
    position/neu-n-ft = 7424.477390
    position/neu-e-ft = -0.000095
    position/neu-u-ft = 30000.291574
    position/distance-from-start-lat-mt = 7413.776597
    position/distance-from-start-lon-mt = 0.000095
    position/geod-alt-ft = 30001.615652
    position/h-sl-ft = 30001.615653

Time 50:
    position/neu-n-ft = 37392.207156
    position/neu-e-ft = -0.018114
    position/neu-u-ft = 30010.842574
    position/distance-from-start-lat-mt = 37338.256797
    position/distance-from-start-lon-mt = 0.018088
    position/geod-alt-ft = 30044.427430
    position/h-sl-ft = 30044.427432

Time 100:
    position/neu-n-ft = 74779.571828
    position/neu-e-ft = 0.064755
    position/neu-u-ft = 29944.511870
    position/distance-from-start-lat-mt = 74671.675133
    position/distance-from-start-lon-mt = 0.064662
    position/geod-alt-ft = 30078.833996
    position/h-sl-ft = 30078.834005
```

Note that there is no autopilot engaged, the aircraft is simply trimmed at 30,000ft and then left to fly for 100s. If there was an autopilot engaged to maintain an altitude, then `position/h-sl-ft` would've remained constant at 30,000ft.

bcoconni:
The flat Earth assumption is not limited to using a fixed coordinate frame, it is also assuming constant gravity (i.e. g = 9,81 m/s²), no Coriolis effect, no curvature to the ground (i.e. the ground is an infinite plane), and some others I may be omitting. So I am a bit concerned that this feature may be advertised as such.

>> the computations that are made in this PR are only making sense when you stay in the vicinity of the IC start

> Correct, but anyone using a local tangential frame to the earth should know this 😉

I am not sure we can reasonably expect that from anyone using this feature. Not that I am against this PR but I think we must be wary about what this feature is doing and what it can be used for.

IMHO the following points should be properly documented:
1. The NEU frame is fixed in the ECEF frame (i.e. following the Earth rotation)
2. It is fixed at the aircraft initial position (similar to the

seanmcleod70:
@bcoconni no rush from my side. Happy to create another PR to address your comments/suggestions.

bcoconni:
> > I find that the property names could be more explicit
> 
> @bcoconni so something like the following? Which makes it clear what the origin of the NEU coordinates is, i.e. to prevent confusion where someone could think that the NEU frame is attached the aircraft?
> 
> `position/distance-from-start-neu-n-ft`
> 
> So encoding the origin, `distance-from-start`, then the frame, `neu`, then the axis, `n`, then the distance units, `ft`.
> 
> Roughly matching the existing property - `position/distance-from-start-lat-mt`.

Yeah, that was my idea although I am not sure `distance` is the proper word since a distance is supposed to be positive. I was also thinking `position/position-from-start-neu-n-ft` but the word `position` appears twice. May be `position/from-start-neu-n-ft` ?

src/models/FGAuxiliary.cpp
```diff
@@ -180,7 +184,15 @@ bool FGAuxiliary::Run(bool Holding)
...
   vMachUVW(eW) = vAeroUVW(eW) / in.SoundSpeed;
 
-  // Position
+  // Position tracking in local frame with local frame origin at lat, lon of initial condition
+  // and at 0 altitude relative to the reference ellipsoid. Position is NEU (North, East, UP) in feet.
+  if (!NEUFromStartInitialized) {
+    NEUStartLocation = FDMExec->GetIC()->GetPosition();
+    NEUStartLocation.SetPositionGeodetic(NEUStartLocation.GetLongitude(), NEUStartLocation.GetGeodLatitudeRad(), 0.0);
+    NEUFromStartInitialized = true;
+  }
```
bcoconni, line 193: I'd rather initialize `NEUStartLocation` in a method that would be called by `FGFDMExec::Initialize`:
https://github.com/JSBSim-Team/jsbsim/blob/83e3df2c49c13eae78ab7853848cdd9e02c6c5ac/src/FGFDMExec.cpp#L676-L681
Since `FGFDMExec::Initialize` is called itself by `FGFDMExec::RunIC` that would ensure that `NEUStartLocation` would be initialized at the same time than the rest of the simulation. And that would save te usage of the flag `NEUFromStartInitialized`.

src/models/FGAuxiliary.cpp
```diff
@@ -355,15 +367,15 @@ double FGAuxiliary::GetNlf(void) const
 double FGAuxiliary::GetLongitudeRelativePosition(void) const
 {
   return in.vLocation.GetDistanceTo(FDMExec->GetIC()->GetLongitudeRadIC(),
-                                    in.vLocation.GetGeodLatitudeRad())* fttom;
+                                    in.vLocation.GetGeodLatitudeRad());
```
bcoconni, line 370: It is not clear why the unit conversion from feet to meters is removed ? This method is used to compute the property `position/distance-from-start-lon-mt` which is expressed in meters.
https://github.com/JSBSim-Team/jsbsim/blob/83e3df2c49c13eae78ab7853848cdd9e02c6c5ac/src/models/FGAuxiliary.cpp#L446

src/models/FGAuxiliary.cpp
```diff
@@ -355,15 +367,15 @@ double FGAuxiliary::GetNlf(void) const
...
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 double FGAuxiliary::GetLatitudeRelativePosition(void) const
 {
   return in.vLocation.GetDistanceTo(in.vLocation.GetLongitude(),
-                                    FDMExec->GetIC()->GetGeodLatitudeRadIC())* fttom;
+                                    FDMExec->GetIC()->GetGeodLatitudeRadIC());
```
bcoconni, line 378: Same as above. This method is used to compute the property `position/distance-from-start-lat-mt` which is expressed in meters.
https://github.com/JSBSim-Team/jsbsim/blob/83e3df2c49c13eae78ab7853848cdd9e02c6c5ac/src/models/FGAuxiliary.cpp#L447

src/models/FGAuxiliary.cpp
```diff
@@ -372,7 +384,7 @@ double FGAuxiliary::GetDistanceRelativePosition(void) const
 {
   auto ic = FDMExec->GetIC();
   return in.vLocation.GetDistanceTo(ic->GetLongitudeRadIC(),
-                                    ic->GetGeodLatitudeRadIC())* fttom;
+                                    ic->GetGeodLatitudeRadIC());
```
bcoconni, line 387: Same as above. This method is used to compute the property `position/distance-from-start-mag-mt` which is expressed in meters.
https://github.com/JSBSim-Team/jsbsim/blob/83e3df2c49c13eae78ab7853848cdd9e02c6c5ac/src/models/FGAuxiliary.cpp#L448

bcoconni (commented):
The way `NEUStartLocation` is initialized is not idiomatic to JSBSim.

And I do not understand why the conversion from feet to meters has been removed while the results are used by properties which are returning their value in meters ?

PR #1323, opened 2025-08-25
Fixed f450 example: engine rpm is zero when low fps

scropions:
image
image
image

### DIFF
--- a/aircraft/F450/Effectors.xml
+++ b/aircraft/F450/Effectors.xml
@@ -8,7 +8,7 @@
         <min>0.0</min>
         <max>1.0</max>
       </clipto>
-      <output>fcs/throttle-cmd-norm[0]</output>
+      <output>fcs/throttle-pos-norm[0]</output>
     </actuator>
 
     <actuator name="Aft Left">
@@ -17,7 +17,7 @@
         <min>0.0</min>
         <max>1.0</max>
       </clipto>
-      <output>fcs/throttle-cmd-norm[1]</output>
+      <output>fcs/throttle-pos-norm[1]</output>
     </actuator>
 
     <actuator name="Front Left">
@@ -26,7 +26,7 @@
         <min>0.0</min>
         <max>1.0</max>
       </clipto>
-      <output>fcs/throttle-cmd-norm[2]</output>
+      <output>fcs/throttle-pos-norm[2]</output>
     </actuator>
 
     <actuator name="Aft Right">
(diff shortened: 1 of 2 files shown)


seanmcleod70:
@scropions we need more context, at a minimum link to the original issue, since you can't assume everyone reviewing this PR would've seen and read the original issue that was submitted.

Secondly, can you explain how or why the UE frame rate effects the original properties used in `Effectors.xml` and `FlightControl.xml` and how changing the properties to the ones in your commit avoid the issue.

scropions:
https://github.com/JSBSim-Team/jsbsim/tree/83e3df2c49c13eae78ab7853848cdd9e02c6c5ac its forked branch

seanmcleod70:
> I'll take a look at your changes to `FlightControl.xml` next.

In my comments above, my references to what I saw in `FlightControl.xml` were based on what I saw in the original `FlightControl.xml`. 

> It looks like instead there is a `fcs/cmdHeave_nd` but I don't see a direct real-time pilot input for it? I only see it set in `scripts\Test_F450_Launch.xml`?

So, I see now looking at your modified `FlightControl.xml` that you addressed this, replacing `fcs/cmdHeave_nd` with `fcs/throttle-cmd-norm[x]`.

However, it looks like you've replaced a single heave command with 4 separate ones. I'm not sure that's typical for a quadcopter. I would expect to only have a single pilot input for heave.

seanmcleod70:
> And we have 4 engines, which means the input power must be separate for each.

You don't need to have 4 user inputs because there are 4 engines. The user can input a single heave/vertical velocity command etc. and the flight control system will then work out how much power to request from each of the 4 engines. The flight control system also looks at the other user inputs/commands like pitch rate, roll rate, yaw rate and then mixes all those with the single heave command to come up with particular power requests for each of the 4 engines.

scropions:
image
image
column G - /fdm/jsbsim/propulsion/engine/propeller-rpm <br>
column H - /fdm/jsbsim/propulsion/engine/thrust-lbs <br>
column I - /fdm/jsbsim/propulsion/engine/pow <br>
[before changes](https://github.com/user-attachments/files/22019594/quad_log.csv) <br>
[after changes](https://github.com/user-attachments/files/22019596/quad_log1.csv) <br>

scropions:
Its just input. In unreal engine we take only one input too

seanmcleod70:
@scropions thanks, your PR has now been merged.

PR #1327, opened 2025-08-28
Improve NEU properties

seanmcleod70:
- Fix the inadvertent removal of the ft to m conversion for the `position/distance-from-start-lat/lon-mt` properties in the original PR for NEU properties.
- Improve the naming of the NEU properties, i.e. `position/from-start-neu-n-ft`
- Initialize NEU starting position in a more JSBSim idiomatic way

### DIFF
--- a/src/models/FGAuxiliary.cpp
+++ b/src/models/FGAuxiliary.cpp
@@ -84,6 +84,7 @@ FGAuxiliary::FGAuxiliary(FGFDMExec* fdmex) : FGModel(fdmex)
   vMachUVW.InitMatrix();
   vEulerRates.InitMatrix();
   vNEUFromStart.InitMatrix();
+  NEUCalcValid = false;
 
   bind();
 
@@ -118,6 +119,7 @@ bool FGAuxiliary::InitModel(void)
   vMachUVW.InitMatrix();
   vEulerRates.InitMatrix();
   vNEUFromStart.InitMatrix();
+  NEUCalcValid = false;
 
   return true;
 }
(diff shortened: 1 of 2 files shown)


src/models/FGAuxiliary.cpp
```diff
@@ -186,11 +192,6 @@ bool FGAuxiliary::Run(bool Holding)
 
   // Position tracking in local frame with local frame origin at lat, lon of initial condition
   // and at 0 altitude relative to the reference ellipsoid. Position is NEU (North, East, UP) in feet.
-  if (!NEUFromStartInitialized) {
-    NEUStartLocation = FDMExec->GetIC()->GetPosition();
-    NEUStartLocation.SetPositionGeodetic(NEUStartLocation.GetLongitude(), NEUStartLocation.GetGeodLatitudeRad(), 0.0);
-    NEUFromStartInitialized = true;
-  }
   vNEUFromStart = NEUStartLocation.LocationToLocal(in.vLocation);
   vNEUFromStart(3) *= -1.0;  // Flip sign for Up, so + for altitude above reference ellipsoid
```
bcoconni, line 196: Don't we want this value to be computed on demand ? i.e. if and only of the property `position/from-start-neu-*ft` is read ?

That would mean the vector `vNEUFromStart` would be updated by the method `FGAuxiliary::GetNEUPositionFromStart`.

bcoconni (commented):
Looks good to me 👍

Just some food for thought: we might want to avoid computing the NEU position only when needed rather than forcing it being updated every time step.

seanmcleod70:
One workaround for the const issue, make the methods const again and declare `vNEUFromStart` and `NEUValid` as mutable.

```c++
+  double GetNEUPositionFromStart(int idx) const { return (GetNEUPositionFromStart())(idx); }
+  const FGColumnVector3& GetNEUPositionFromStart() const;

  FGLocation NEUStartLocation;
  mutable FGColumnVector3 vNEUFromStart;
  mutable bool NEUValid;
```

PR #1330, opened 2025-09-06
Add `NDEBUG` to MSVC compile definitions for release builds.

bcoconni:
Following PR #1317 which is introducing some code conditionally compiled with the flag `NDEBUG`, this PR modifies the CMake files so that `NDEBUG` is set when building JSBSim with MSVC for the `Release` and `RelWithDebInfo` builds.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -100,9 +100,15 @@ if(WIN32)
     # The flag JSBSIM_EXPORT must be declared PRIVATE to avoid propagating it
     # to other targets such as the executable, the Python module and the unit
     # tests. Otherwise the linking of these targets will fail.
-    target_compile_definitions(libJSBSim PRIVATE JSBSIM_EXPORT)
+    target_compile_definitions(libJSBSim PRIVATE
+                               JSBSIM_EXPORT
+                               $<$<CONFIG:Release>:NDEBUG>
+                               $<$<CONFIG:RelWithDebInfo>:NDEBUG>)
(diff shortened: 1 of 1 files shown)


seanmcleod70:
I was 95% confident previously that `NDEBUG` was being set for release builds by CMake for MSVC, but wasn't able to find a reference to a CMake config file that I could double-check for the CMake version being used by Github for our builds.

So I tried the following on my PC, which has CMake version 3.31.6

```sh
C:\source\jsbsim\src>cmake -LAH
...
...
-- The CXX compiler identification is MSVC 19.44.35213.0
....
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
....
// Flags used by the CXX compiler during MINSIZEREL builds.
CMAKE_CXX_FLAGS_MINSIZEREL:STRING=/MD /O1 /Ob1 /DNDEBUG

// Flags used by the CXX compiler during RELEASE builds.
CMAKE_CXX_FLAGS_RELEASE:STRING=/MD /O2 /Ob2 /DNDEBUG

// Flags used by the CXX compiler during RELWITHDEBINFO builds.
CMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING=/MD /Zi /O2 /Ob1 /DNDEBUG
```

So I'm 99% confident now that `NDEBUG` is being set by CMake for our Github builds for MSVC 😉

PR #1331, opened 2025-09-07
[skip ci] Notebooks improvements

bcoconni:
In continuation to the PR #1314 which allowed to upload the example notebooks on the Google Colab platform, this PR improves the code a bit:
* The GitHub repo is no longer cloned since the aircraft data is included in the Python package (using `pip install jsbsim`)
* Avoid using terminal commands (the ones with the bang `!`) except `pip install`. These are not portable ( and can be replaced by portable Python code (i.e. using `os.get_cwd()` instead of `!pwd`).
* The commands `%matplotlib inline` and `jsbsim.FGJSBBase().debug_lvl = 0` need not be issued on each notebook cells.

And sorry, my IDE cleans up trailing spaces before saving files so there are a few white spaces modifications that are polluting the diff.


seanmcleod70:
Ah, and so the following works since I guess in this case there must be logic in either the Python wrapper for FGFDMExec or within FGFDMExec itself to figure out the path to the JSBSim files relative to it's installed location?

```python
PATH_TO_JSBSIM_FILES=None
```

bcoconni:
Looks good! Go for it

seanmcleod70:
Hmm, I was pretty certain that when I looked up `[skip ci]` a while ago that I saw documentation mentioning that it could appear anywhere in the line.

But doing a quick search on it and looking at documentation from Github:

https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

> If any commit message in your push or the HEAD commit of your PR **contains** the strings [skip ci], [ci skip], [no ci], [skip actions], or [actions skip] workflows triggered on the push or pull_request events will be skipped.

https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs

> Workflows that would otherwise be triggered using on: push or on: pull_request won't be triggered if you **add** any of the following strings to the commit message in a push, or the HEAD commit of a pull request:

I see the Github documentation is ambiguous.  

And regular Bing Copilot (as opposed to Copilot Pro :wink:) has an example with `[skip ci]` at the end of the line.

image

Now having said all that I hadn't looked at `jsbsim/.github/workflows/cpp-python-build.yml` partly because I thought (without knowing too much about how the CI is setup) that this level of filter checking happened at a higher level by Github before our custom CI build actions were invoked.

Actually reading: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs

> If a workflow is skipped due to [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) or a commit message (see below), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

So does that not imply that our workflow `cpp-python-build` will be skipped and therefore the following is superfluous? 

https://github.com/JSBSim-Team/jsbsim/blob/09f9169adc4e1495c6104b6028d6d327621d90c1/.github/workflows/cpp-python-build.yml#L14

And maybe was only required before 2021 before Github added support?

bcoconni:
@seanmcleod70 I've just realized that one of your question has remained unanswered.
> so the following works since I guess in this case there must be logic in either the Python wrapper for FGFDMExec or within FGFDMExec itself to figure out the path to the JSBSim files relative to it's installed location?

Yes, the logic is in the Python wrapper. When `None` is passed to the constructor of `FGFDMExec`, the `else` branch is taken in the code below and JSBSim root is set to the "*default root directory*" i.e. the directory where the standard aircraft have been installed by `pip`.
https://github.com/JSBSim-Team/jsbsim/blob/ac3ec7cb357aa789f1af31516a7fc3b1b3319605/python/jsbsim.pyx.in#L729-L736

The function `get_default_root_dir` scans all the standard directories where the Python packages are installed. It is looking for a directory `jsbsim/aircraft` in one of these directories and return the corresponding path or raise an exception on failure.

https://github.com/JSBSim-Team/jsbsim/blob/ac3ec7cb357aa789f1af31516a7fc3b1b3319605/python/jsbsim.pyx.in#L65-L72

PR #1335, opened 2025-09-13
Synchronize CMake min version with the overall project.

bcoconni:
With the release of CMake 4.0, [compatibility with versions earlier than 3.5 are no longer supported](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features). This PR updates the CMake minimum version requirement to 3.15 for the MSIS test that is run by our CI workflow.

Version 3.15 is already the minimum version required for the overall *JSBSim* project so this will not imply an upgrade for the developers or anyone compiling JSBSim from scratch:
https://github.com/JSBSim-Team/jsbsim/blob/405c00d830cf4dd912908085d606c057d9c415d8/CMakeLists.txt#L1

### DIFF
--- a/src/models/atmosphere/MSIS/CMakeLists.txt
+++ b/src/models/atmosphere/MSIS/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.1)
+cmake_minimum_required (VERSION 3.15)
 
 project(NRLMSIS)
 


bcoconni:
For the record, the files `CMakeLists.txt` and `src/models/atmosphere/MSIS/CMakeLists.txt` are the only CMake files who are calling the `cmake_minimum_required` command.

bcoconni:
Merged because the macos-latest platform appears to have migrated to CMake 4.0, which causes [the CI workflow for PR #1336](https://github.com/JSBSim-Team/jsbsim/actions/runs/17710448568/job/50328380674) to fail:

```
Run mkdir build && cd build
  mkdir build && cd build
  cmake ../src/models/atmosphere/MSIS
  shell: /bin/bash -e {0}
  env:
    release: false
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
-- Configuring incomplete, errors occurred!
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


Error: Process completed with exit code 1.
```

PR #1375, opened 2026-01-13
Add function GetAltitudeDependentConditions to API

AirshipSim:
New API function to access atmospheric conditions in one go.
Related to #1194 , which needed split.

### DIFF
--- a/src/models/FGAtmosphere.cpp
+++ b/src/models/FGAtmosphere.cpp
@@ -195,6 +195,17 @@ double FGAtmosphere::GetSoundSpeed(double altitude) const
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+// Get all atmospheric conditions which depend on the altitude
+
+void FGAtmosphere::GetAltitudeDependentConditions(AltitudeDependentConditions& conditions, double altitude) const
+{
+  conditions.Temperature = GetTemperature(altitude);
+  conditions.Pressure = GetPressure(altitude);
+  conditions.Density = (conditions.Pressure / (Reng * conditions.Temperature));
+  conditions.SoundSpeed = sqrt(SHRatio * Reng * conditions.Temperature);
+}
+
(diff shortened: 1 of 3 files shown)


agodemar:
@bcoconni @seanmcleod70 are you fine with this PR? Looks a good one to me

seanmcleod70:
Just refreshing my memory since https://github.com/JSBSim-Team/jsbsim/pull/1194 is back from Nov/Dec 2024, i.e. more than a year ago.

So a client could implement their own:

```c++
void GetAltitudeDependentValues(AltitudeDependentValues& values, double altitude) const
{
  values.Temperature = pAtmosphere->GetTemperature(altitude);
  values.Pressure = pAtmosphere->GetPressure(altitude);
  values.Density = pAtmosphere->GetDensity(altitude);
  values.SoundSpeed = pAtmosphere->GetSoundSpeed(altitude);
}
```

But you're saying there is a noticeable performance improvement to be had by not having `GetTemperature()` called 3 times and `GetPressure()` called twice? By re-implementing the density and speed of sound calculations using cached values for temperature and pressure and implementing the formula calculation in multiple places now?

```c++
double FGAtmosphere::GetDensity(double altitude) const
{
  return GetPressure(altitude)/(Reng * GetTemperature(altitude));
}

double FGAtmosphere::GetSoundSpeed(double altitude) const
{
  return sqrt(SHRatio * Reng * GetTemperature(altitude));
}
```

Now maybe I'm missing some additional context, but doesn't the atmosphere model already provide cached calculated values for `Density` and `Soundspeed`? Plus `Temperature` and `Pressure`.

https://github.com/JSBSim-Team/jsbsim/blob/9ea4925a68cdf1ee5a4c0c33e591c744f624cc03/src/models/FGAtmosphere.cpp#L142-L170

With the cached values for the current timestep accessed via:

https://github.com/JSBSim-Team/jsbsim/blob/9ea4925a68cdf1ee5a4c0c33e591c744f624cc03/src/models/FGAtmosphere.h#L166-L167

https://github.com/JSBSim-Team/jsbsim/blob/9ea4925a68cdf1ee5a4c0c33e591c744f624cc03/src/models/FGAtmosphere.h#L182-L183

Unless you're specifically wanting to use a different altitude compared to the current altitude for the aircraft, for the current timestep?

PR #1377, opened 2026-01-21
Edit J246 aircraft guidance properties

heshpdx:
Use different property domain for eccentricity so that apogee and perigee are calculated correctly.

### DIFF
--- a/aircraft/J246/Systems/J246GuidanceExecutive.xml
+++ b/aircraft/J246/Systems/J246GuidanceExecutive.xml
@@ -162,60 +162,6 @@
 
   <channel name="Orbital Parameters">
 
-    <fcs_function name="guidance/executive/semimajor-axis-length-ft">
-      <description>
-        This equation calculates the semimajor-axis length in feet. a = mu/(2*mu/r - v^2) 
-      </description>
-      <function>
-        <quotient>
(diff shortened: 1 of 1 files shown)


bcoconni:
Ping ! Any comments about my suggestion ?

heshpdx:
I just tried your suggestion 1 and then added suggestion 2. These all work also. The outputs are just slightly different from the original (due to rounding error) although it grows over time. In both cases O3 does equal Ofast, so that problem is not there. If we change `scripts/J2461.xml` then we should also change `scripts/J2460.xml` which has the same property.

Here are the outputs from J2461.xml. 

[O3-J2461-bc.xml.txt](https://github.com/user-attachments/files/24998385/O3-J2461-bc.xml.txt)
[O3-J2461-bc2.xml.txt](https://github.com/user-attachments/files/24998382/O3-J2461-bc2.xml.txt)
[Ofast-J2461-bc.xml.txt](https://github.com/user-attachments/files/24998383/Ofast-J2461-bc.xml.txt)
[Ofast-J2461-bc2.xml.txt](https://github.com/user-attachments/files/24998384/Ofast-J2461-bc2.xml.txt)

This is the difference in O3 when replacing the apogee/perigee computations (within FP rounding errors).

image

PR #1382, opened 2026-02-01
Optimize calls to `SGPropertyNode::getNode` in `FGAtmosphere::Calculate`

bcoconni:
### The problem

As I mentioned in https://github.com/JSBSim-Team/jsbsim/pull/1375#issuecomment-3829000981, the method `FGAtmosphere::Calculate` is spending a significant amount of time resolving property names into property nodes (i.e. calling `SGPropertyNode::getNode`):

image

Actually the logic is based on the existence of the nodes `atmosphere/override/temperature`, `atmosphere/override/pressure` and `atmosphere/override/density`.

https://github.com/JSBSim-Team/jsbsim/blob/5988bb46d02b55f019fef7ee6390b01fd3e38984/src/models/FGAtmosphere.cpp#L142-L170

If one of these nodes is existing then the temperature/pressure/density is extracted from its corresponding node. This feature is designed to use an external atmospheric model that would provide its data through the `atmosphere/override` properties.

The problem is that interrogating `SGPropertyNode::getNode` at each iteration is costly: 72012 calls to `SGPropertyNode::getNode` translates into 3,940,012 calls to `strncmp` (in the screenshot above it is an SSE2 optimized version of it).

### The proposed solution

Since all the nodes are located below `atmosphere/override` then the new code is only testing if `atmosphere/override` is existing at each iteration. If it does not exist then there is no point in interrogating the other nodes for the temperature/pressure/density. This is dividing the number of calls to `SGPropertyNode::getNode` by 3 and the time spent in `FGAtmosphere::Calculate` by roughly the same order of magnitude.

image

In addition to that, once the existence of a node is confirmed, it is registered in a variable of type `SGPropertyNode_ptr` so that `FGAtmosphere::Calculate` will no longer need to resolve the name into a node thus avoiding further calls to `SGPropertyNode::getNode`.

### DIFF
--- a/src/models/FGAtmosphere.cpp
+++ b/src/models/FGAtmosphere.cpp
@@ -61,6 +61,10 @@ FGAtmosphere::FGAtmosphere(FGFDMExec* fdmex)
   Name = "FGAtmosphere";
 
   bind();
+  SGPropertyNode* root = PropertyManager->GetNode();
+  atmosphere_node = root->getNode("atmosphere");
+
+  assert(atmosphere_node);
   Debug(0);
 }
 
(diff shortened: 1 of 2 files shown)


seanmcleod70:
Yep, I wasn't aware of this atmosphere override option, so wasn't really sure if there was a standard for when the override property needed to be created etc. So yes, safest option for now in terms of backwards compatibility is to go with your current implementation. A 3%-4% increase is still a pretty useful improvement.

PR #1408, opened 2026-03-19
FGInitialCondition::Load throws instead of returning false

AirshipSim:
Following discussion #1407 
_Hint: Hide white space differences when checking Files changes_

The function `FGInitialCondition::Load` was obviously patched up along the years. Upon error, it could
- throw a LogException (unhandled)
- throw XMLLogException  (unhandled)
- return false  

It is now unified to return false in all cases. Functions inside Load( ) which could throw are now guarded by a try / catch.

Other corrected minor issue: the reading of engine configuration was done directly in FGInitialCondition::Load( ), instead of being done in the functions Load_v1 or Load_v2 like any other tag.

### Current usage in the codebase
In the codebase, Load( ) was called in two different places:

https://github.com/JSBSim-Team/jsbsim/blob/98c03332e75ac5749fb2f699700154130e2a3a38/src/input_output/FGScript.cpp#L195-L200

https://github.com/JSBSim-Team/jsbsim/blob/98c03332e75ac5749fb2f699700154130e2a3a38/src/JSBSim.cpp#L477-L482

In both cases, the return value ( true or false ) was correctly checked, nothing to do there.

### Testing
Testing in command line, the errors are caught correctly and JSBSim terminates with the proper error message, instead of terminate suddenly with FATAL ERROR and less log:

```
Could not open file: Path "aircraft/c172r/reset0.xml"
Init file: Path "aircraft/c172r/reset0" could not be loaded.
Initialization unsuccessful
Script file Path "scripts/c1721.xml" was not successfully loaded

D:\DEV\jsbsim>
```
```
In file aircraft/c172r/reset00.xml: line 16
XML parse error: mismatched tag
Error parsing Path "aircraft/c172r/reset00"
In file aircraft/c172r/reset00.xml: line 16
XML parse error: mismatched tagInitialization unsuccessful
Script file Path "scripts/c1721.xml" was not successfully loaded

D:\DEV\jsbsim\Debug\JSBSim.exe (process 2552) exited with code -1 (0xffffffff).
```

### DIFF
--- a/src/initialization/FGInitialCondition.cpp
+++ b/src/initialization/FGInitialCondition.cpp
@@ -1020,15 +1020,15 @@ bool FGInitialCondition::Load(const SGPath& rstfile, bool useAircraftPath)
 
   // Make sure that the document is valid
   if (!document) {
-    LogException err;
-    err << "File: " << init_file_name << " could not be read.\n";
-    throw err;
+    FGLogging log(LogLevel::ERROR);
+    log << "Init file: " << init_file_name << " could not be loaded.\n";
+    return false;
   }
 
   if (document->GetName() != "initialize") {
-    LogException err;
-    err << "File: " << init_file_name << " is not a reset file.\n";
-    throw err;
+    FGLogging log(LogLevel::ERROR);
+    log << "File: " << init_file_name << " is not a valid reset file.\n";
+    return false;
   }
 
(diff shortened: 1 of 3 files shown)


bcoconni:
PR dropped since there has been no update from the author.

PR #1424, opened 2026-04-10
File tree reorganization

bcoconni:
The code in `src/utilities` has not received any significant contribution for 10+ years. It contains the code of the legacy `aeromatic` tool as well as some code to plot results either with gnuplot or [dislin](https://www.dislin.de/).

I am not aware of anyone using these plot tools since nowadays people are mostly plotting curves using our Python module with Jupyter notebooks and matplotlib.

Also the legacy `aeromatic` is superseded by `aeromatic++`.

I am suggesting to migrate the code located in `src/utilities` to `utils` for the following reasons:
* Even if no one seems to be using this code any longer, it will be kept in the repo for anyone to use or contribute.
* Currently, the code in `src/utilities` is incorrectly included in the Python source package (i.e. `jsbsim-1.3.0.tar.gz`). This is due to the way file filters have been written in `pyproject.toml`. I could certainly modify the filters to exclude explicitly `src/utilities` but I find it easier to move the code out of reach :smile:

This PR is creating a new folder `utils/simplot` to host the code of the plot tools currently located in `src/utilities`. The name `simplot` is coming from the build instructions in `src/utilities/main.cpp`:
https://github.com/JSBSim-Team/jsbsim/blob/0a779a94ec2e8a9cc72282fc6f66cec9468b948f/src/utilities/main.cpp#L10
The folder `data_plot` is also moved to `utils/simplot/data_plot` since it contains XML data files that are read by `plotXMLVisitor.cpp`.

The file `JSBSimPlot.xsd` which contains the XML schema for the files in `data_plot` is also moved to `utils/simplot` alongside with the folder `data_plot` itself.

And finally this PR is also moving the legacy `aeromatic` from `src/utilities/aeromatic` to `utils/aeromatic`.

The expected result is to have a cleaner file tree without dropping any legacy code.


agodemar:
Looks good to me.

jonsberndt:
Looks good to me, too.

seanmcleod70:
Looks good.

bcoconni:
Thanks PR merged.

PR #1431, opened 2026-04-13
Update Python DefaultLogger to perform filtering

seanmcleod70:
As per discussion https://github.com/JSBSim-Team/jsbsim/discussions/1427.

### DIFF
--- a/python/jsbsim.pyx.in
+++ b/python/jsbsim.pyx.in
@@ -172,10 +172,16 @@ def _inherit_docstring_from_super_class(cls):
 @_inherit_docstring_from_super_class
 class DefaultLogger(FGLogger):
     """Default logger: print messages to stdout without formatting."""
+    def __init__(self, min_level=LogLevel.BULK):
+        self.min_level = min_level
+    def set_min_level(self, level: LogLevel) -> None:
+        self.min_level = level
     def file_location(self, filename: str, line: int) -> None:
-        print(f"\nIn file {filename}: line {line}")
+        if int(self.log_level) >= int(self.min_level):
+            print(f"\nIn file {filename}: line {line}")
     def message(self, message: str) -> None:
-        print(f"{message}", end='')
+        if int(self.log_level) >= int(self.min_level):
+            print(f"{message}", end='')
 
 
 # Reset the logger before Python shutdown to prevent crashes.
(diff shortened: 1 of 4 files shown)


bcoconni (approved):
Looks good to me.

bcoconni:
Oh, I've just realized that `pyxstubgen.py` is failing. Your PR uses some syntax that the tool is not recognizing. I will investigate that.

seanmcleod70:
Hmm, maybe missing type info for `min_level` argument to `__init__()`?

```python
class DefaultLogger(FGLogger):
    """Default logger: print messages to stdout without formatting."""
    def __init__(self, min_level=LogLevel.BULK):
```
Line 316 `pname, ptype = self.visit(parameter.children[0])` in `pyxstubgen.py` where it's failing, more than likely getting `pname` back but no `ptype`?

I thought type hints were hints and not compulsory 😉 Also what confused me was that only the MacOSX build was failing.

bcoconni:
> Line 316 `pname, ptype = self.visit(parameter.children[0])` in `pyxstubgen.py` where it's failing, more than likely getting `pname` back but no `ptype`?

Most likely, yes.
> I thought type hints were hints and not compulsory 😉 Also what confused me was that only the MacOSX build was failing.

Type hints are indeed optional in Python but `pyxstubgen.py` fails when one is missing. This is an unexpected feature of the tool 😉 because we do want the Python module to have type hints. However the code in `pyxstubgen.py` could be modified so that it issues a meaningful error message rather than crashes.

seanmcleod70:
I've pushed a commit to add the type hint. But the MacOSX build is now failing, and one of the Linux builds with network issues in terms of downloading Julia stuff.

seanmcleod70:
I re-ran the MacOSX build and it's still failing in `pyxstubgen.py`, but now in a different place.

```sh
  File "/Users/runner/work/jsbsim/jsbsim/python/pyxstubgen.py", line 325, in funcdef
    pvalue = get_constant(value)
  File "/Users/runner/work/jsbsim/jsbsim/python/pyxstubgen.py", line 90, in get_constant
    assert tree.children == []
AssertionError
```

bcoconni:
Yeah, `pyxstubgen.py` cannot handle Cython code where the default value of a function parameter is a "dotted" name i.e. `LogLevel.BULK`.

I have pushed a commit to your PR to fix that. The tests are now passing.

Issue #497, opened 2021-08-27
<pure_gain>: no error messages


mike402:
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => **Please do not submit support request here**, see note at the top of this template.

**Describe the issue**
I noticed that `<pure_gain>` does not report any errors in its configuration.

**What is the current behavior?**
1. Without `<input>` crashes FlightGear (see below the backtrace).

2. Without `<gain>` it silently assumes the gain of 1.

3. With two `<input>`s or `<gain>`s it silently uses the first ones.

**What is the expected behavior?**
The expected behaviour would be that in case "1." it would print an error and stop FlightGear, and in the other 2 cases print a warning, with enough information to find the misconfigured `<gain>`.

**What is the motivation / use case for changing the behavior?**
The warnings would reduce the rate of silent errors occuring in FlgihtGear aircraft.

**Please tell us about your environment:**
FlightGear next as of 2021-08-26.

**Other information**
Backtrace for case "1.".
```
Thread 1 "fgfs" received signal SIGSEGV, Segmentation fault.

<...>

#0  0x000055555623d9c5 in JSBSim::FGGain::Run() ()
#1  0x00005555561f12a9 in JSBSim::FGFCS::Run(bool) ()
#2  0x000055555628f386 in JSBSim::FGFDMExec::Run() ()
#3  0x000055555628f556 in JSBSim::FGFDMExec::RunIC() ()
#4  0x00005555561a4018 in FGJSBsim::init() ()
#5  0x0000555555f42c00 in FDMShell::doInitAndBind() ()
#6  0x0000555555f4381a in FDMShell::update(double) ()
#7  0x00005555564f784a in SGSubsystemGroup::Member::update(double) ()
#8  0x00005555564fc107 in SGSubsystemGroup::update(double) ()
#9  0x00005555564f2eba in SGSubsystemMgr::update(double) ()
#10 0x000055555613f512 in fgMainLoop() ()
#11 0x00005555560926b2 in fgOSMainLoop() ()
#12 0x0000555556144377 in fgMainInit(int, char**) ()
#13 0x00005555558ea7fe in main ()
#0  0x000055555623d9c5 in JSBSim::FGGain::Run() ()
#1  0x00005555561f12a9 in JSBSim::FGFCS::Run(bool) ()
#2  0x000055555628f386 in JSBSim::FGFDMExec::Run() ()
[...]

PR #1433, opened 2026-04-15
Add warning for duplicate <gain> elements in FGGain (#497)

SaVi456:
When multiple <gain> elements are specified in a pure_gain, scheduled_gain, or aerosurface_scale component, only the first is used. This adds an error-level warning so users are notified of the misconfiguration, following the existing pattern from CheckInputNodes() in FGFCSComponent.cpp.

Addresses the remaining duplicate element detection issue from #497.

Note: This code was written entirely by Claude Opus 4 (Anthropic AI Assistant) via Claude Code. The human contributor directed the project selection and submitted the PR.

### DIFF
--- a/src/models/flight_control/FGGain.cpp
+++ b/src/models/flight_control/FGGain.cpp
@@ -70,10 +70,18 @@ FGGain::FGGain(FGFCS* fcs, Element* element) : FGFCSComponent(fcs, element)
 
   auto PropertyManager = fcs->GetPropertyManager();
   Element* gain_element = element->FindElement("gain");
-  if (gain_element)
+  if (gain_element) {
     Gain = new FGParameterValue(gain_element, PropertyManager);
-  else
+    // Warn if duplicate <gain> elements are provided
+    if (element->FindNextElement("gain")) {
+      FGXMLLogging log(element, LogLevel::ERROR);
+      log << "      Multiple <gain> elements specified.\n"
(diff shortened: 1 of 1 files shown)


bcoconni:
Thanks for your contribution.

However regarding the detection of redundant `<gain>` it has been told in the issue #497 that C++ was not the way to proceed (see https://github.com/JSBSim-Team/jsbsim/issues/497#issuecomment-974849486). You should use XML schemas for that, more specifically the file `JSBSimSystem.xsd` for the particular case of `<gain>`. The XML schema is up to date since PR #1038.
https://github.com/JSBSim-Team/jsbsim/blob/027c1414f7b8c048b4fa6f0c2258261637f6869e/JSBSimCommon.xsd#L635-L652

SaVi456:
@bcoconni 

Thanks for the feedback. I see that the XSD schema at JSBSimCommon.xsd (lines 635-652) already constrains <gain> to maxOccurs="1" by default, so XML validation would catch duplicates.

Should I close this PR, or would you prefer I redirect the effort elsewhere on #497 — for example, ensuring the remaining issue (duplicate elements being silently accepted when schema validation is not used) is documented?

This comment was drafted by Claude Opus 4 (Anthropic AI Assistant) via [Claude Code](https://github.com/anthropics/claude-code).

bcoconni:
I have closed issue #497.
Thanks for your contribution.
PR closed as "Won't fix"

PR #1441, opened 2026-04-22
JOSS paper

seanmcleod70:
A WIP (Work In Progress) PR for the joint editing of the JOSS paper.

PR #1452, opened 2026-05-07
Ignore Python notebooks.

bcoconni:
In addition to MarkDown files, this PR modifies the CI workflow to ignore Python notebooks modifications.

**Reminder:** commits containing **only** some modifications of one or several Python notebooks will be ignored. If the commit modifies some other files that do not match the `paths-ignore:` filter then the `push`/`pull_request` event will trigger the workflow.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -11,12 +11,14 @@ on:
     paths-ignore:
       - "**/*.md"
       - "joss_paper/**"
+      - "**/*.ipynb"
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [master]
     paths-ignore:
       - "**/*.md"
       - "joss_paper/**"
+      - "**/*.ipynb"
   schedule:
     - cron: '0 9 * * 2'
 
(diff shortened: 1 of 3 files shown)


agodemar:
Yes, this setting is so convenient. Please, merge

seanmcleod70:
> **Reminder:** commits containing **only** some modifications of one or several Python notebooks will be ignored. If the commit modifies some other files that do not match the `paths-ignore:` filter then the `push`/`pull_request` event will trigger the workflow.

Also, remember it's cumulative, so if a PR has an initial commit which only modifies Jupyter notebooks `.ipynb` and some additional commit is added to the PR which does modify other files, then from that point onwards all commits will kick-off the workflow tasks. In that case you need to start adding `[skip ci]` to subsequent commits to the PR **if** these other files that are modified shouldn't kick off the workflows.

bcoconni:
> Also, remember it's cumulative, so if a PR has an initial commit which only modifies Jupyter notebooks `.ipynb` and some additional commit is added to the PR which does modify other files, then from that point onwards all commits will kick-off the workflow tasks. In that case you need to start adding `[skip ci]` to subsequent commits to the PR **if** these other files that are modified shouldn't kick off the workflows.

Yes, I saw your explanation in the PR #1441 about this unexpected behavior. Thanks for the reminder 👍

PR #1454, opened 2026-05-12
Throw an exception for unknown table type in GetValue()

seanmcleod70:
While doing a build with MSVC, via Visual Studio I noticed a single compiler warning:

> 1>C:\source\jsbsim\src\math\FGTable.cpp(546,1): warning C4715: 'JSBSim::FGTable::GetValue': not all control paths return a value

Based on there being no return value for the `default` case:

https://github.com/JSBSim-Team/jsbsim/blob/1ecd08f228b68e44c4ecafef2b5979f750223edf/src/math/FGTable.cpp#L520-L543

Not that serious for a debug build, which will assert, but a release build means more than likely a crash soon afterwards.

### DIFF
--- a/src/math/FGTable.cpp
+++ b/src/math/FGTable.cpp
@@ -542,6 +542,9 @@ double FGTable::GetValue(void) const
   }
   default:
     assert(false);
+    LogException err;
+    err << "Unknown table type.\n";
+    throw err;
   }
 }
 
--- a/src/math/FGTable.cpp
+++ b/src/math/FGTable.cpp
@@ -45,6 +45,35 @@ INCLUDES
 #include "input_output/FGXMLElement.h"
 #include "input_output/string_utilities.h"
 
+/******************************************************************************
+UNREACHABLE(msg) macro is designed to optimize impossible code branches away
+in release optimized builds and to assert in debug builds with a custom message.
+
+Often used for unreachable default in fully covered switches.
+
+Only use for invariants that can be proven from the structure of the code, not
+for assumptions about external inputs like user input, deserialized data etc.
+
+c++23 defines std::unreachable(), however JSBSim currently assumes c++17 as
+the minimum, so if being compiled with c++ < 23 the macro makes use of a
+compiler specific intrinsic.
+******************************************************************************/
+
+#ifdef NDEBUG
+#if __cplusplus >= 202302L
+#include <utility>
+#define UNREACHABLE(msg) std::unreachable();
+#elif defined(__GNUC__) || defined(__clang__)
+#define UNREACHABLE(msg) __builtin_unreachable();
+#elif defined(_MSC_VER)
+#define UNREACHABLE(msg) __assume(false);
+#else
+#define UNREACHABLE(msg) do { } while(0); // fallback: no-op
+#endif
+#else
+#define UNREACHABLE(msg) assert(false && msg); std::abort();
+#endif
+
 using namespace std;
 
 namespace JSBSim {
@@ -377,7 +406,7 @@ FGTable::FGTable(std::shared_ptr<FGPropertyManager> pm, Element* el,
       *this << buf;
       break;
     default:
-      assert(false); // Should never be called
+      UNREACHABLE("invalid table type") // Should never be called
       break;
     }
   }
@@ -461,7 +490,7 @@ FGTable::FGTable(std::shared_ptr<FGPropertyManager> pm, Element* el,
     if (Data.size() != nRows+1) missingData(el, nRows, Data.size()-1);
     break;
   default:
-    assert(false);  // Should never be called
+    UNREACHABLE("invalid table type") // Should never be called
     break;
   }
 
@@ -541,10 +570,7 @@ double FGTable::GetValue(void) const
     return GetValue(lookupPropertyValues.data());
   }
   default:
-    assert(false);
-    LogException err;
-    err << "Unknown table type.\n";
-    throw err;
+    UNREACHABLE("invalid table type") // Should never be called
   }
 }
 


seanmcleod70:
Talking of `assert`, there are quite a number of them in `FGTable.cp`, but when checking the C/C++ build and test workflow it doesn't look like we ever build a debug version?

So we never get a chance to test all the asserts during our CI?

bcoconni:
Since the member `FGTable::Type` is private, this `assert` is supposed never to be reached. And if it is reached it means there is an internal error in the C++ code (i.e. an invalid value is attributed to `Type`). Also if the `default` clause is removed in the `switch` then the compilers will complain with some warnings.

IMHO throwing an exception here would not make much sense since there is nothing the final user could do to fix the problem since that exception would be the result of a bug in JSBSim. And basically, the result will be the same: the program will halt.

bcoconni:
> Talking of `assert`, there are quite a number of them in `FGTable.cp`, but when checking the C/C++ build and test workflow it doesn't look like we ever build a debug version?

At least the Linux and macOS targets are built with the `-g` flag and without `NDEBUG` defined. So as far as I can tell, their tests are run with the `assert`'s enabled.

https://github.com/JSBSim-Team/jsbsim/blob/1ecd08f228b68e44c4ecafef2b5979f750223edf/.github/workflows/cpp-python-build.yml#L146
https://github.com/JSBSim-Team/jsbsim/blob/1ecd08f228b68e44c4ecafef2b5979f750223edf/.github/workflows/cpp-python-build.yml#L499

seanmcleod70:
> ...but a release build means more than likely a crash soon afterwards...

Hmm, taking a closer look at the standard calling convention for x64 at least on Unix systems and Windows I'm not so sure that in a build without asserts, typically release builds, will result in a crash soon afterwards.

Calling convention used in conjunction with the `call` instruction:

> Calling Conventions (SysV & Microsoft x64)
>
> SysV (Linux/Unix): First six integer/pointer args in rdi, rsi, rdx, rcx, r8, r9.
>
> Microsoft x64: First four integer/pointer args in rcx, rdx, r8, r9.
>
> Remaining arguments go on the stack.
>
> Return values ≤ 64 bits are in rax; **floating-point in xmm0**.

Setting a breakpoint in `FGTable::GetValue()` the first hit of the breakpoint is from a call in `FGGain::Run()`:

```c++
FGGain::Run() {
  double SchedGain = Table->GetValue();
```
Disassembly showing the `xmm0` register value, the return value, being copied into the stack variable `SchedGain`.

```asm
    double SchedGain = Table->GetValue();
0000000140185DC5  mov         rax,qword ptr [this]  
0000000140185DCA  mov         rax,qword ptr [rax+120h]  
0000000140185DD1  mov         rcx,qword ptr [this]  
0000000140185DD6  mov         rcx,qword ptr [rcx+120h]  
0000000140185DDD  mov         rax,qword ptr [rax]  
0000000140185DE0  call        qword ptr [rax+8]  
0000000140185DE3  movsd       mmword ptr [rsp+20h],xmm0 
```

So, in the case where there is some internal corruption of the `Type` instance in `FGTable` it means that the register `xmm0` doesn't get updated with any value. So, in effect the caller will be receiving whatever the last value was that was put into the `xmm0` register, more than likely from some earlier call to a function that returned a float, or some other floating point code/calculation that set `xmm0` to some value. Which means the program could continue running for quite a while until these semi-random values potentially result in a nan during some calculation.

So I guess some of our options are?

#### Option 1
Do nothing, and live with the compiler warning (level 1 in MSVC)

#### Option 2
Throw an exception effectively only for non-assert builds, gets rid of compiler warning.

#### Option 3
Create an assert for this case which isn't removed during non-assert builds, but also return some value, which will never execute but will remove the compiler warning.

#### Option 4
Selectively disable this compiler warning in this function/file across the different compilers.

#### Option 5
Return some default value, maybe `nan` after the existing assert. Will remove the compiler warning, and for non-assert builds will more than likely result in a much earlier crash.

seanmcleod70:
Hmm, so not long after posting the above, I happened to come across `std::unreachable()` and a very good writeup on it at - https://towardsdev.com/cpp23-std-unreachable-compiler-optimization-1ab0251be605

So my new suggestion/option 6 is to make use of `unreachable` for this case, and other `default` cases for `Type` in the `FGTable` class.

Given that the `Type` `enum` is a type that we own and exhaustively cover.

Given we're not targeting `c++23` yet we would make use of the polyfill from the writeup:

```c++
#if __cplusplus >= 202302L
    #include <utility>
    #define MY_UNREACHABLE() std::unreachable()
#elif defined(__GNUC__) || defined(__clang__)
    #define MY_UNREACHABLE() __builtin_unreachable()
#elif defined(_MSC_VER)
    #define MY_UNREACHABLE() __assume(false)
#else
    #define MY_UNREACHABLE() do { } while(0) // fallback: no-op
#endif
```

And I'd suggest using his pattern which also caters for debug builds.

```c++
#include <utility>
#include <cassert>

[[noreturn]] inline void truly_unreachable(
    [[maybe_unused]] const char* msg = "unreachable code reached"
) {
    #ifndef NDEBUG
        // Debug: crash loudly with a message
        assert(false && "unreachable code was reached");
        // If assert is somehow disabled, abort anyway
        std::abort();
    #else
        // Release: trust the programmer, optimize aggressively
        //std::unreachable();
        MY_UNREACHABLE()
    #endif
}
```

This approach does mean that in builds without `assert()` the behavior of the program is undefined if the `Type` `enum` does get corrupted.

bcoconni:
Makes sense but the code of `truly_unreachable` has 2 weirdness's:
1. The argument `[[maybe_unused]]` is actually *never used*. The parameter `msg` is not used anywhere whatever `#ifdef` branch is taken.
2. Adding `std::abort` after `assert` seems a bit over the top, no ?!?

bcoconni:
Other than the 2 points above, looks good to me to replace the occurrences of `assert(false)` in `FGTable` by calls to `truly_unreachable`.

seanmcleod70:
Hmm, on closer inspection and testing there are a couple of issues with the `truly_unreachable()` function.

I initially thought he had a typo in:
`assert(false && "unreachable code was reached");` 
and he meant to type:
`assert(false && msg);`

But the issue with that is that `assert()` is `#defined`, e.g.

```c++
    _ACRTIMP void __cdecl _wassert(
        _In_z_ wchar_t const* _Message,
        _In_z_ wchar_t const* _File,
        _In_   unsigned       _Line
        );

    #define assert(expression) ((void)(                                                       \
            (!!(expression)) ||                                                               \
            (_wassert(_CRT_WIDE(#expression), _CRT_WIDE(__FILE__), (unsigned)(__LINE__)), 0)) \
        )
```

So at the time that the compiler's pre-processor processes `assert(false && msg);` it literarily turns that into the string: `false && msg`, so you don't get the runtime version of the contents of `msg` printed out during the assertion, instead you get:

`Assertion failed: false && msg, file C:\source\jsbsim\src\math\FGTable.cpp, line 72` 

Which also brings up the next issues with the call to `assert()` in the `truly_unreachable()` function.

Take a look at the line number printed above during the assertion, `line 72`, compared to the source code line numbers for my testing.

image

Which is the line number in the `truly_unreachable()` function, instead of being line `line 581` which is what we want to display to the user as the source of the assertion.

So instead of the function `truly_unreachable()` my proposed solution is to define another macro, `UNREACHABLE`, which you can see defined on lines 61-65, and used on line 580.

So now, with `NDEBUG` not defined we get an assertion which both contains our actual message we supplied on line 580, and the line number printed out for the assertion is the correct source line number of 580.

`Assertion failed: false && "invalid table type", file C:\source\jsbsim\src\math\FGTable.cpp, line 580`

If `NDEBUG` is defined then the macro `UNREACHABLE` is replaced with the macro `MY_UNREACHABLE` which in the MSVC case results in being replaced with the call to `__assume(false)`, which the optimizer then optimizes out of the generated code.

Lastly, in terms of the call to `std::abort()` after the call to `assert()`, that does seem to be some really extra paranoia in terms of `assert` potentially being disabled via some additional #define over and above `NDEBUG`.

bcoconni:
> So instead of the function `truly_unreachable()` my proposed solution is to define another macro, `UNREACHABLE`, which you can see defined on lines 61-65, and used on line 580.

Looks good to me with one small detail: a better name than `MY_UNREACHABLE` could be picked :smile: Or even better, directly inlined `MY_UNREACHABLE` into `UNREACHABLE` ?
```c++
#ifdef NDEBUG
    #if __cplusplus >= 202302L
        #include <utility>
        #define UNREACHABLE(msg) std::unreachable()
    #elif defined(__GNUC__) || defined(__clang__)
        #define UNREACHABLE(msg) __builtin_unreachable()
    #elif defined(_MSC_VER)
        #define UNREACHABLE(msg) __assume(false)
    #else
        #define UNREACHABLE(msg) do { } while(0) // fallback: no-op
    #endif
#else
    #define UNREACHABLE(msg) assert(false && msg); std::abort;
#endif
```
Code using `#ifdef`'s is ugly anyway.

seanmcleod70:
Yep, coming up with one name is easier/better than 2 😉

So I've gone with your one big #ifdef suggestion.

For now I've defined the macro in `FGTable.cpp`. If we do come across other parts of the source code that could make use of it then we can move it to some common utility header file.

bcoconni (approved):
Looks good to me.

Issue #1455, opened 2026-05-12
MSVC static CMake build broken due to missing expat linkage


AbrilRBS:
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

**Describe the issue**

As part of the effort to add the 1.3.0 release to Conan Center ([Linked PR](https://github.com/conan-io/conan-center-index/pull/29914)), we found an issue when compiling the MSVC static build using CMake and system libraries, where we got errors like (`SYSTEM_EXPAT=ON`)

```
JSBSim.lib(easyxml.cxx.obj) : error LNK2019: unresolved external symbol __imp_XML_SetElementHandler referenced in function "void __cdecl readXML(class std::basic_istream<char,struct std::char_traits<char> > &,class XMLVisitor &,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (?readXML@@YAXAEAV?$basic_istream@DU?$char_traits@D@std@@@std@@AEAVXMLVisitor@@AEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@2@@Z)
```

([See full logs here](https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-29914/5/package_build_logs/build_log_jsbsim_1_3_0_e0c514411fa6ff6080c7bb6804d905e7_5fa1880d14c85bcf0d8ef3f8e6ca83faa63be4e5.txt))

The contributor for this PR suggested adding the following patch to solve this issue, and while this looks good on my end, it would be great to have your insight on this, thanks! 

**Please tell us about your environment:**

Windows MSVC 194, static C++ CMake build, `SYSTEM_EXPAT=ON`

**Other information**

While the linked report uses Conan to drive the build, the command looks something like
```
$ cmake -G "Ninja" -DCMAKE_TOOLCHAIN_FILE="generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="C:/j/29914-5/b/jsbsi148f89378073f/p" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" -DCMAKE_C_COMPILER="cl" -DCMAKE_CXX_COMPILER="cl" "C:/j/29914-5/b/jsbsi148f89378073f/b/src"

$ cmake --build "C:\j\29914-5\b\jsbsi148f89378073f\b\build\Release" -- -j8
```

[...]

PR #1458, opened 2026-05-15
Fix builds using the system expat library

bcoconni:
Following the issue #1455, this PR fixes builds that use the system version of the expat library:
* All the logic related to the expat library is moved to `src/simgear/xml/CMakeLists.txt`. Previously the logic was spread across multiple CMake files (`CMakeLists.txt`, `src/CMakeLists.txt` and `src/simgear/xml/CMakeLists.txt`)
* The main target now links against object libraries instead of target objects `$<TARGET_OBJECTS:XXX>`. The latter did not allow to propagate the compilation flags and linked libraries from the target `Xml` to the target `libJSBSim`.
* The file `src/simgear/xml/easyxml.cxx` now imports `<expat.h>` (with brackets) when JSBSim is built using the system expat library. This guarantees that the header will match the library. Previously `easyxml.cxx` was importing `"expat.h"` (with double quotes) meaning that the header `src/simgear/xml/expat.h` was used whether JSBSim was built with the embedded version of expat or not. This is meant to avoid a potential version mismatch between the system expat library and the version embedded in JSBSim.
* Building with a system expat library is now also tested with MSVC (previously was only tested on Linux).
* Builds using `pkg-config` have been fixed (previously `pkg_check_modules` was not called).
* Took this opportunity to replace all occurrences of `PKG_CONFIG_FOUND` by `PkgConfig_FOUND` since [the former is deprecated since CMake 4.2](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#deprecated-variables).

This PR fixes the issue #1455, since JSBSim is now built with the CMake imported target `EXPAT::EXPAT` (or `PkgConfig::PC_EXPAT`), thus automatically taking into account the required flags for compilation such as `XML_STATIC` on Windows.


seanmcleod70:
Looks good.

agodemar:
Good.

bcoconni:
Thanks! The PR is merged.

PR #1459, opened 2026-05-15
Expose Input enable/disable and implement for additional cases

seanmcleod70:
As per discussion - https://github.com/JSBSim-Team/jsbsim/discussions/1428 this PR adds `DisableInput` and `EnableInput` to `FGFDMExec` to allow clients to control whether any inputs are enabled or disabled.

- Add `FGFDMExec::DisableInput()` and `FGFDMExec::EnableInput()`
- Add python wrappers `FGFDMExec.disable_input()` and `FGFDMExec.enable_input()`
- Use of `std::unique_ptr<FGfdmSocket>` in place of original raw pointer for `socket` in `FGInputSocket`.

`FGFDMExec` contains a single `FGInput` instance which in turn has a `std::vector<FGInputType*>` containing each of the input types that are defined in the aircraft's FDM etc.

There was already `Enable() Disable()` logic in `FGInput`, updating an `enable` field, which is then used within `FGInput::Run()` to potentially not call `FGInputType::Run()` for all of the input instances .

```c++
bool FGInput::Run(bool Holding)
{
  if (FDMExec->GetTrimStatus()) return true;
  if (FGModel::Run(Holding)) return true;
  if (!enabled) return true;

  vector<FGInputType*>::iterator it;
  for (it = InputTypes.begin(); it != InputTypes.end(); ++it)
    (*it)->Run(Holding);

  return false;
}
```

So this would then prevent a socket based input type from reading from a socket. However there was no mechanism to pass on this change in enabled state to the input types for a global change in enabled state for the `FGInput` instance. Which means the input types couldn't close their socket.

Also there was no way to disable an input type so that between creation and it's `InitModel()` method being called it would know not to bother creating a socket.

So the current enabled state of `FGInput` is now passed along to the ctors of the `FGInputType` classes.

```c++
  if (type.empty() || type == "SOCKET") {
    Input = new FGInputSocket(FDMExec, enabled);
  } else if (type == "QTJSBSIM") {
    Input = new FGUDPInputSocket(FDMExec, enabled);
  } 
```

So there was a bit more logic that needed to be added compared to what I was expecting when I started.

So now a user, especially one concerned about running multiple instances at the same time of an aircraft with socket based inputs defined in it's FDM can use the following logic to prevent any sockets being created by any instance so that there is no concurrency clash with regards to creating a listening UDP/TCP socket on the same port.

```python
fdm = jsbsim.FGFDMExec(None)
fdm.disable_input()
fdm.load_model('737')
```

A user can also toggle the enabled state during execution, e.g.

```python
fdm = jsbsim.FGFDMExec(None)
fdm.load_model('737')
while true:
   if some_condition:
      fdm.disable_input()
   if some_other_condition:
      fdm.enable_input()
   fdm.run()
```

### DIFF
--- a/src/input_output/FGInputSocket.cpp
+++ b/src/input_output/FGInputSocket.cpp
@@ -98,7 +98,6 @@ bool FGInputSocket::InitModel(void)
 
 bool FGInputSocket::CreateSocket()
 {
-  socket.reset(); // close any existing socket
   socket = std::make_unique<FGfdmSocket>(SockPort, SockProtocol);
 
   if (!socket) return false;
--- a/src/input_output/FGInputSocket.cpp
+++ b/src/input_output/FGInputSocket.cpp
@@ -53,8 +53,8 @@ namespace JSBSim {
 CLASS IMPLEMENTATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
-FGInputSocket::FGInputSocket(FGFDMExec* fdmex) :
-  FGInputType(fdmex), socket(0), SockProtocol(FGfdmSocket::ptTCP),
+FGInputSocket::FGInputSocket(FGFDMExec* fdmex, bool isEnabled) :
+  FGInputType(fdmex, isEnabled), SockProtocol(FGfdmSocket::ptTCP),
   BlockingInput(false)
 {
 }
@@ -63,7 +63,7 @@ FGInputSocket::FGInputSocket(FGFDMExec* fdmex) :
 
 FGInputSocket::~FGInputSocket()
 {
-  delete socket;
+  socket.reset();
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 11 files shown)


src/input_output/FGInputSocket.cpp
```diff
@@ -63,7 +63,7 @@ FGInputSocket::FGInputSocket(FGFDMExec* fdmex) :
 
 FGInputSocket::~FGInputSocket()
 {
-  delete socket;
+  socket.reset();
 }
```
bcoconni: Not sure why the destructor has been kept? When the member `socket` will be destroyed, the destructor of `FGfdmSocket` will be called anyway, thus closing the socket.

I would have completely removed the destructor.

src/input_output/FGInputSocket.cpp
```diff
@@ -92,17 +92,56 @@ bool FGInputSocket::Load(Element* el)
...
-  }
+  return true;
+}
+
+//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+bool FGInputSocket::CreateSocket()
+{
+  socket.reset(); // close any existing socket
+  socket = std::make_unique<FGfdmSocket>(SockPort, SockProtocol);
```
bcoconni: I think calling `socket.reset()` is useless in this situation as it is the whole point of `std::unique_ptr<>` to destroy upon release. See [unique_ptr: what is the effect of calling reset before assignment](https://stackoverflow.com/questions/60224528/unique-ptr-what-is-the-effect-of-calling-reset-before-assignment).

src/input_output/FGUDPInputSocket.cpp
bcoconni, line 1: Shouldn't `socket == 0` be replaced by `!socket` since `socket` is now a `std::unique_ptr<>` ?
https://github.com/JSBSim-Team/jsbsim/blob/d6fee6bb02c95a6b8ea1954cfc7254f840a82be6/src/input_output/FGUDPInputSocket.cpp#L91-L93

bcoconni (changes requested):
Two pedantic remarks about calls to `unique_ptr<>.reset()` that are IMHO useless.

Another remark about a comparison `socket == 0` in `FGUDPInputSocket.cpp` that is most likely resulting in the construction of a `std::unique_ptr<FGfdmSocket>(nullptr)` before making the comparison.

src/input_output/FGInputSocket.cpp
seanmcleod70: Yep, the explicit `FGInputSocket` destructor isn't necessary now. Was a bit of a basic mechanical update for the change to `std::unique_ptr`. I'll push a commit to remove it.

src/input_output/FGUDPInputSocket.cpp
seanmcleod70: > Another remark about a comparison `socket == 0` in `FGUDPInputSocket.cpp` that is most likely resulting in the construction of a `std::unique_ptr<FGfdmSocket>(nullptr)` before making the comparison.

There is an explicit `==` operator that takes a `unique_ptr` and a `nullptr_t`, which is what I confirmed MSVC is calling in this case.

```c++
_EXPORT_STD template <class _Ty, class _Dx>
_NODISCARD _CONSTEXPR23 bool operator==(const c<_Ty, _Dx>& _Left, nullptr_t) noexcept {
    return !_Left;
}
```

Just to double-check that it isn't an additional operator added by MSVC I checked the C++ standard:

https://en.cppreference.com/cpp/memory/unique_ptr/operator_cmp

Turns out I had actually changed this sort of comparison in `FGInputSocket::CreateSocket()`.

```diff
- if (socket == 0) return false;
+ if (!socket) return false;
```

But in the derived class `FGUDPInputSocket` I hadn't actually really looked at the socket code, just the enable state via the constructor.

So I'll push a commit to change this instance in the derived class to match.

src/input_output/FGInputSocket.cpp
seanmcleod70: The original reason I had the `socket.reset()` call in `CreateSocket()` was because in my testing I tested the case of calling `Input->Enable()` multiple times without matching `Input->Disable()` calls in-between, in which case the 2nd call to `Input->Enable()` would result in a socket error since a socket with the same port etc. already existed.

```sh
Creating input TCP socket on port 5137
Could not bind to TCP input socket, error = 0
```

I then added the `if (enabled) return;` check in the `Enable()` method instead, but forgot to remove the `socket.reset()` call in `CreateSocket()`.

```c++
void FGInputSocket::Enable()
{
  if (enabled) return; // already enabled
  FGInputType::Enable();
  CreateSocket();
}
```

So I'll remove the `socket.reset()` call in `CreateSocket()`.

seanmcleod70:
@bcoconni I've pushed 3 commits to address your comments.

bcoconni (approved):
Looks good to me.

PR #1462, opened 2026-05-18
Add FGNullAtmosphere and extend LoadPlanet atmosphere factory

AlexanderRex:
Adds atmosphere support so JSBSim can simulate vehicles on Mars and the Moon.

**What changes**

* `<atmosphere model="Mars"/>` and `<atmosphere model="Null"/>` are now valid in `LoadPlanet`, alongside the existing `"MSIS"`.
* New `FGNullAtmosphere` class: a vacuum. Pressure and density are zero; temperature is set to the cosmic microwave background (~2.7 K) just so reads return a finite value. Use it for the Moon, Mercury, asteroids.

**Bugs found while wiring this up**

* `LoadPlanet` was unbinding the old atmosphere *after* constructing the new one. But `FGAtmosphere`'s constructor calls `bind()`, and at that point the old model still owned `atmosphere/T-R`, `atmosphere/P-psf`, etc., so the new model's ties silently failed, and the FDM kept reading stale values from the old instance. Fixed by unbinding first.
* `FGMars::Calculate` was returning temperature in °F, but `Density = P / (Reng·T)` needs absolute temperature. Surface T came out around −26 °F → negative absolute → negative density → NaN on the first step. Added the +459.67 F→R offset to both lapse branches.
* `FGMars` and `FGNullAtmosphere` constructors were calling `bind()` themselves, but `FGAtmosphere`'s base constructor already does. The duplicate was harmless at FDM startup but produced a flood of `Failed to tie property` warnings once the factory started swapping atmosphere models at runtime. Removed.

**Tests**

Three new cases in `tests/TestPlanet.py`:
* `test_mars_atmosphere`: via the planet `<include>` path; checks T = 433.99 R, P = 14.62 psf, ρ ≈ 1.43e-5 slugs/ft³, plus Mars radius and surface gravity.
* `test_load_Mars_atmosphere`: same expected values via runtime `LoadPlanet`.
* `test_load_Null_atmosphere`: vacuum, P = ρ = 0, T = 4.905 R.

`tests/moon.xml` now declares `<atmosphere model="Null"/>` (it's the Moon, no atmosphere). New `tests/mars.xml`. Full `TestPlanet` suite (8/8) passes.


src/FGFDMExec.cpp
```diff
@@ -788,10 +790,27 @@ bool FGFDMExec::LoadPlanet(Element* element)
...
-      if (model == "MSIS") {
-        // Replace the existing atmosphere model
+
+      // Unbind the existing atmosphere BEFORE constructing the replacement.
+      // FGAtmosphere's constructor calls bind(), which would silently fail to
+      // tie any property already owned by the previous model and leave the new
+      // model's properties dangling (reads return stale values from the old
+      // instance).
+      bool willReplace = (model == "MSIS" || model == "Mars" || model == "Null");
+      if (willReplace) {
```
bcoconni: I am not a huge fan of this construct. The variable `willReplace` is not used anywhere else and the naming is not particularly informative (what will replace what ?) I would prefer a one liner:
```suggestion
      if (model == "msis" || model == "mars" || model == "null") {
```

src/FGFDMExec.cpp
```diff
@@ -788,10 +790,27 @@ bool FGFDMExec::LoadPlanet(Element* element)
     Element* atm_element = element->FindElement("atmosphere");
     if (atm_element && atm_element->HasAttribute("model")) {
       string model = atm_element->GetAttributeValue("model");
```
bcoconni: I know this was not done initially but your patch is using camel casing while `MSIS` is all capitals (because it is an acronym). Let's make sure the user will not struggle with the casing and compare lowercase names using the function `to_lower` from `string_utilities.h`.
```suggestion
      string model = to_lower(atm_element->GetAttributeValue("model"));
```

src/FGFDMExec.cpp
```diff
@@ -788,10 +790,27 @@ bool FGFDMExec::LoadPlanet(Element* element)
...
-        Models[eAtmosphere] = std::make_shared<FGMSIS>(this);
+      }
+
+      std::shared_ptr<FGAtmosphere> newAtmosphere;
+      if (model == "MSIS")
+        newAtmosphere = std::make_shared<FGMSIS>(this);
+      else if (model == "Mars")
+        newAtmosphere = std::make_shared<FGMars>(this);
+      else if (model == "Null")
+        newAtmosphere = std::make_shared<FGNullAtmosphere>(this);
```
bcoconni: Following the comment I made above, use lowercase names:
```suggestion
      if (model == "miss")
        newAtmosphere = std::make_shared<FGMSIS>(this);
      else if (model == "mars")
        newAtmosphere = std::make_shared<FGMars>(this);
      else if (model == "null")
        newAtmosphere = std::make_shared<FGNullAtmosphere>(this);
```

bcoconni:
Thanks for the PR @AlexanderRex.

Even though it makes sense to restore the Mars atmosphere, I am not so sure about adding the "null" atmosphere. In JSBSim, an empty `<aerodynamics/>` section makes the vehicle behave as if there were no atmosphere. And even with the null atmosphere, you would not specify an `<aerodynamics/>` section anyway so I am not sure about the added value of the null atmosphere model beyond being explicit ? But maybe I am missing something ?

And also I am bit worried about setting the pressure and density to zero. Has the code been reviewed to make sure that this would not lead to divisions by zero or any other similar problems ?

bcoconni (commented):
I would suggest a few minor changes in addition to the points I have mentioned in another comment.

bcoconni:
> **Bugs found while wiring this up**
>
> `LoadPlanet` was unbinding the old atmosphere *after* constructing the new one. But `FGAtmosphere`'s constructor calls `bind()`, and at that point the old model still owned `atmosphere/T-R`, `atmosphere/P-psf`, etc., so the new model's ties silently failed, and the FDM kept reading stale values from the old instance. Fixed by unbinding first.

I am not sure about what you exactly mean by that ? Your statement is worded as if you had found an error in JSBSim, but the code in `master` branch is already meeting this requirement (unbinding before constructing):
https://github.com/JSBSim-Team/jsbsim/blob/5e23ead05e4bca6b125c9a7a3bdd12f7cd19b9c2/src/FGFDMExec.cpp#L792-L795

And would the readings be frozen then the test `test_load_MSIS_atmosphere` would fail
https://github.com/JSBSim-Team/jsbsim/blob/5e23ead05e4bca6b125c9a7a3bdd12f7cd19b9c2/tests/TestPlanet.py#L64-L76

AlexanderRex:
> Thanks for the PR @AlexanderRex.
> 
> Even though it makes sense to restore the Mars atmosphere, I am not so sure about adding the "null" atmosphere. In JSBSim, an empty `<aerodynamics/>` section makes the vehicle behave as if there were no atmosphere. And even with the null atmosphere, you would not specify an `<aerodynamics/>` section anyway so I am not sure about the added value of the null atmosphere model beyond being explicit ? But maybe I am missing something ?
> 
> And also I am bit worried about setting the pressure and density to zero. Has the code been reviewed to make sure that this would not lead to divisions by zero or any other similar problems ?

Hi @bcoconni, thanks for your review!

The unbind/construct order on master is already correct. What I described
as a "bug" was actually a regression introduced by my own refactor (I
pulled `make_shared` out of the `if (model == "MSIS")` block and ran it
before `Unbind`); I caught it during testing and reverted to master's
order, but the PR description framed the reversal as a JSBSim fix.
Sorry for the confusion, I will update the wording.

I also check NullAtmosphere again and yeah it will trigger zero division, so I think to remove it from this PR and and keep this change focused on restoring Mars.
We can discuss a vacuum model separately if there's interest :) (I thought to bring moon demo with Apollo model)

Will implement your suggestions and remove NullAtmosphere

seanmcleod70:
> We can discuss a vacuum model separately if there's interest :) (I thought to bring moon demo with Apollo model)

If you mean just the lunar module descending to and taking off from the Moon then as @bcoconni mentioned you don't necessarily require a vacuum atmosphere model. You can set the planet to the Moon and leave the default earth atmosphere  model and simply have no force/moment entries in the `<aerodynamics/>` section, so JSBSim would just be modelling Moon's gravity plus the lunar module's thrust.

If you're talking about modelling the full Apollo mission from earth to the moon and back, then the issue is that there isn't a mechanism to swap/fade out/fade in one planet to another and one atmosphere to another.

The closest I've seen I guess is the JSBSim/FlightGear Space Shuttle model, which needs aerodynamic forces/moments for take-off and landing, but needs the aerodynamic forces/moments to pretty much go to 0 while in orbit. Actually not sure if they use the MSIS atmosphere model, given it's defined from 0-1000km. Space Shuttle's orbit was between 320km - 620km.

Or whether they use the standard ISA, and even though it's not defined above 87km, I'm guessing the density drops close enough to 0 that the Space Shuttle ends up with close to 0 aerodynamic forces/moments?

bcoconni (approved):
Looks good to me.

AlexanderRex:
> > We can discuss a vacuum model separately if there's interest :) (I thought to bring moon demo with Apollo model)
> 
> If you mean just the lunar module descending to and taking off from the Moon then as @bcoconni mentioned you don't necessarily require a vacuum atmosphere model. You can set the planet to the Moon and leave the default earth atmosphere model and simply have no force/moment entries in the `<aerodynamics/>` section, so JSBSim would just be modelling Moon's gravity plus the lunar module's thrust.
> 
> If you're talking about modelling the full Apollo mission from earth to the moon and back, then the issue is that there isn't a mechanism to swap/fade out/fade in one planet to another and one atmosphere to another.
> 
> The closest I've seen I guess is the JSBSim/FlightGear Space Shuttle model, which needs aerodynamic forces/moments for take-off and landing, but needs the aerodynamic forces/moments to pretty much go to 0 while in orbit. Actually not sure if they use the MSIS atmosphere model, given it's defined from 0-1000km. Space Shuttle's orbit was between 320km - 620km.
> 
> Or whether they use the standard ISA, and even though it's not defined above 87km, I'm guessing the density drops close enough to 0 that the Space Shuttle ends up with close to 0 aerodynamic forces/moments?

To be honest I thought to split it into the stages like: 
1) Land on the Lunar surface with Appolo first using custom controller or RL policy
2) Simulate process starting from Earth

But at the moment Mars atmosphere is more necessary for me, so thats why I start fixing it :)
And thank you so much for merging my PR!

PR #1463, opened 2026-05-18
Don't assume STDOUT LogLevel for all calls to FGTable::Print()

seanmcleod70:
While testing various combinations of `jsbsim.FGJSBBase().debug_lvl` and `jsbsim.get_logger().set_min_level(jsbsim.LogLevel.XXX)` I noticed something odd with the output of FGTable data.

With `jsbsim.FGJSBBase().debug_lvl = 1` and `jsbsim.get_logger().set_min_level(jsbsim.LogLevel.DEBUG)` I would see output along these lines, i.e. function name with any associated table for the function.

```sh
    Function: aero/coefficient/Clr
    1 dimensional table with 2 rows.
        0       0.1
        2       0.033
    Function: aero/coefficient/Clda
    Function: aero/coefficient/Cldr
    Function: aero/coefficient/Cmalpha
    1 dimensional table with 2 rows.
        0       -1.2
        2       -0.3
    Function: aero/coefficient/Cmde
```

But with `jsbsim.get_logger().set_min_level(jsbsim.LogLevel.FATAL)` then the function names would disappear, and other aircraft config output, e.g. mass details, FCS details etc., would also disappear, but there would be table data output with no context, e.g.

```sh
    1 dimensional table with 2 rows.
        0       0.1
        2       0.033
    1 dimensional table with 2 rows.
        0       -1.2
        2       -0.3
```

That's because all the other data/details were being output at `LogLevel::DEBUG`, e.g.

```c++
  if (debug_lvl & 1) { // Standard console startup message output
    if (from == 0) { // Constructor
      if (!Name.empty()) {
        FGLogging log(LogLevel::DEBUG);
        log << "    Function: " << Name << endl;
      }
    }

  if (debug_lvl > 0) {
    MassBalance->GetMassPropertiesReport(0);
```

But `FGTable::Print()` was set to output at `LogLevel::STDOUT`:

```c++
FGTable::FGTable()
  if (debug_lvl & 1) Print();

void FGTable::Print(void)
{
  FGLogging out(LogLevel::STDOUT);
  out << std::setprecision(4);

  switch(Type) {
    case tt1D:
      out << "    1 dimensional table with " << nRows << " rows.\n";
      break;
    case tt2D:
```

So I've modified `FGTable::Print()` to take a reference to a `FGLogging` instance, and defaulted the constructor to assume `LogLevel::DEBUG`. Then modified the handful of call sites that explicitly call `FGTable::Print()` to pass a reference to a `FGLogging` instance. 

```c++
FGTable::FGTable()
  if (debug_lvl & 1) {
    FGLogging out(LogLevel::DEBUG);
    Print(out);
  }

void FGTable::Print(FGLogging& out)
{
  out << std::setprecision(4);
```

So at some point if there was a command line argument, or property tree entry allowing a user to request a `FGTable` to print itself, it could pass a `FGLogging` instance with `LogLevel::STDOUT`.

### DIFF
--- a/src/math/FGTable.h
+++ b/src/math/FGTable.h
@@ -411,7 +411,8 @@ public:
 
   unsigned int GetNumRows() const {return nRows;}
 
-  void Print(FGLogging& out = FGLogging(LogLevel::STDOUT));
+  void Print();
+  void Print(FGLogging& out);
 
   std::string GetName(void) const {return Name;}
 
--- a/src/math/FGTable.h
+++ b/src/math/FGTable.h
@@ -411,7 +411,7 @@ public:
 
   unsigned int GetNumRows() const {return nRows;}
 
-  void Print(FGLogging& out);
+  void Print(FGLogging& out = FGLogging(LogLevel::STDOUT));
 
   std::string GetName(void) const {return Name;}
 
--- a/src/math/FGTable.h
+++ b/src/math/FGTable.h
@@ -411,7 +411,7 @@ public:
 
   unsigned int GetNumRows() const {return nRows;}
 
-  void Print(void);
+  void Print(FGLogging& out);
 
   std::string GetName(void) const {return Name;}
 
--- a/src/math/FGTable.cpp
+++ b/src/math/FGTable.cpp
@@ -749,6 +749,12 @@ FGTable& FGTable::operator<<(const double x)
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+void FGTable::Print()
+{
+  FGLogging out(LogLevel::STDOUT);
+  Print(out);
+}
+
 void FGTable::Print(FGLogging& out)
 {
   out << std::setprecision(4);
--- a/src/math/FGTable.cpp
+++ b/src/math/FGTable.cpp
@@ -469,7 +469,10 @@ FGTable::FGTable(std::shared_ptr<FGPropertyManager> pm, Element* el,
 
   bind(el, Prefix);
 
-  if (debug_lvl & 1) Print();
+  if (debug_lvl & 1) {
+    FGLogging out(LogLevel::DEBUG);
+    Print(out);
+  }
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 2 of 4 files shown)


src/math/FGTable.h
```diff
@@ -411,7 +411,7 @@ class JSBSIM_API FGTable : public FGParameter, public FGJSBBase
 
   unsigned int GetNumRows() const {return nRows;}
 
-  void Print(void);
+  void Print(FGLogging& out);
```
bcoconni: Maybe we could provide a default value to the parameter so that one could still call `Print()` (with no parameters) and get their table displayed on the console as the method currently does ?

```suggestion
  void Print(FGLogging& out = FGLogging(LogLevel::STDOUT));
```

bcoconni (commented):
Looks good.

I was just wondering if that would make sense to provide a default value to `FGTable::Print` so that it keeps the API backward compatible and the legacy behavior (i.e. printing to the console) ?

src/math/FGTable.h
seanmcleod70: Good suggestion, I've just pushed a commit to add it.

seanmcleod70:
Hmm, not as trivial as simply:

```diff
-  void Print(FGLogging& out);
+  void Print(FGLogging& out = FGLogging(LogLevel::STDOUT));
```

So, the change above compiled fine, with no warnings and ran fine locally with MSVC, however the non-MSVC compilers during the CI produced a compiler error.

```sh
FGTable.h:414:31: error: cannot bind non-const lvalue reference of type ‘JSBSim::FGLogging&’ to an rvalue of type ‘JSBSim::FGLogging’
  414 |   void Print(FGLogging& out = FGLogging(LogLevel::STDOUT));
```

https://www.stroustrup.com/C++11FAQ.html#rval

> In C++, non-const references can bind to lvalues and const references can bind to lvalues or rvalues, but there is nothing that can bind to a non-const rvalue. **That's to protect people from changing the values of temporaries that are destroyed before their new value can be used.**

In this particular case it wasn't an issue that the temporary was changed and destroyed, it flushed it's output when the temporary's destructor was called at the end of the the `Print()` method

Now it turns out that MSVC has a non-standard extension which allows a non-const reference to bind to an rvalue, and will display compiler warning C4239 when it's used, but the warning is only displayed if you enable warning level 4, which I didn't.

https://github.com/MicrosoftDocs/cpp-docs/blob/main/docs/error-messages/compiler-warnings/compiler-warning-level-4-c4239.md

> This type conversion is not allowed by the C++ standard, but it is permitted here as an extension. This warning is always followed by at least one line of explanation describing the language rule being violated.

So, I see 2 main options, both of which involve having 2 overloaded versions of `Print()`.

### Option 1

```c++
void Print()
{
   FGLogging out(LogLevel::STDOUT);
   Print(out);
}

void Print(FGLogging& out)
{
// Printing code
}
```

### Option 2

```c++
void Print(FGLogging&& out = FGLogging(LogLevel::STDOUT))  // rvalue reference
{
   Print(out);
}

void Print(FGLogging& out)
{
// Printing code
}
```

bcoconni:
Thanks for the explanation @seanmcleod70

My favorite is option 1, it is simple and the method using `void` can be inlined.

I like option 2 because it is smart but we prefer simple over smart, don't we ? 😄

seanmcleod70:
Yep, Option 1, how many developers are going to be familiar with an rvalue reference, took me a bit of time to look it up 😉

bcoconni (approved):
Looks good to me.

PR #1464, opened 2026-05-25
Add pitch SAS implementation with Jupyter notebook

ANF1S:
This PR adds a Jupyter notebook example demonstrating a pitch Stability Augmentation System (SAS) implementation for the A-4 Skyhawk aircraft using JSBSim python module. 

The work was originated from the Show and Tell discussion: https://github.com/JSBSim-Team/jsbsim/discussions/1448 , where it was suggested submitting a PR after completing the Jupiter notebook example.

**Changes:**
* Added a new `aircraft/A4x` aircraft based on the original `aircraft/A4` to implement the pitch SAS example from the baseline aircraft model.
* Added a new `pitch_sas.xml` system file, where the pitch SAS logic is located.
* Integrated the pich SAS output in the `pitch <channel>` of the modified aircraft configuration.
* Added a first order lag `<actuator>` model in the `pitch <channel>` to introduce response delay between control command and elevator deflection.
* Added an A-4 image in `examples/python/images`
* Added a Jupter notebook `examples/python/test_pitch_sas.ipynb` demonstrating:
   * The use of pitch SAS design.
   * Elevator doublet test.
   * Aircraft responses visualization.
   * Analytical verification showing changes in damping ratio and natural frequency.

The goal is to provide educational example showing how a SAS can be implemented through xml system file and exercised from python.

**Notes:**
* In the first cell of the notebook, `PATH_TO_JSBSIM_FILES=` has  `"--/--"` value due to the relative path to the directory where the `FDM` is located.


agodemar:
@ANF1S thank you for submitting this well written example!

seanmcleod70:
@ANF1S thanks for the Jupyter notebook writeup, it looks good. Will still take a closer look, but one initial thing that we need to think about is the location of the aircraft model.

I don't think it's a good idea and scalable to add a slightly modified version of an existing aircraft in the repo for potentially each example notebook that needs a modified FDM.

I think it would be better to have an `aircraft` sub-directory within `examples\python`, e.g. `examples\python\aircraft\A4SAS`.

May require a bit of extra work to get it to work with Google Colab.

agodemar:
@seanmcleod70 
> I think it would be better to have an `aircraft` sub-directory within `examples\python`, e.g. `examples\python\aircraft\A4SAS`.

I agree on this: having a local `aircraft` folder in examples will not pollute the official collection of FDMs.

@ANF1S could you please relocate your input files 

- `A4x.xml` renamed `A4SAS.xml` 
- `pitch_sas.xml`

in `examples/python/aircraft/A4SAS`?

A note on the notebook `test_pitch_sas.ipynb`: in the bottom plot of the stacked plots produced by `cell [6]`, please add a legend saying that the orange curve is the elevator command.

ANF1S:
@agodemar , @seanmcleod70 Thanks so much for your feedback. I will make those changes.

agodemar:
@ANF1S thanks for providing corrected files.
@seanmcleod70 for me the changes are OK. Waiting your considerations.

ANF1S:
Hello @agodemar , @seanmcleod70 ,  I have updated the code and pushed the requested modifications.

While testing:
**Problem:**
* `<engine>` and `<thruster>`  path configuration.  Obtained  `Could not open file: J52` when loading aircraft.

**Solution:**
* Created a folder  `A4SAS/engine` and copied the files  `direct.xml` and  `J52.xml` into it.

Please let me know if this implementation looks good to you or if you would prefer any further adjustments.

agodemar:
@ANF1S are you sure the `engine` shouldn't be at the same level of `examples/python`? Besides `examples/python/aircraft`

Moreover, what about:
```Python
PATH_TO_JSBSIM_FILES= "."
```

seanmcleod70:
@agodemar the `engine` folder can be in either, i.e. at the same level as `aircraft`, which makes the engine available to all aircraft, or privately as a sub-directory of the specific aircraft. JSBSim has logic to look in both locations.

ANF1S:
@agodemar , @seanmcleod70  thanks for the feedback. Locally , I tested

* PATH_TO_JSBSIM_FILES= "." and works fine in the current directory

* For `engine`, the path should be  `examples/python/engine`

I will update those modifications

seanmcleod70:
Yep, I just downloaded the notebook, the FDM, engine etc. and tested locally.

I tested both of the following and both worked.

```python
  PATH_TO_JSBSIM_FILES= "./"
  PATH_TO_JSBSIM_FILES=os.getcwd()
```

I'd suggest leaving the `engine` directory as a sub-directory of `aircraft\A4SAS` in terms of it being somewhat self-contained.

seanmcleod70:
I haven't tried Google Colab yet. That's going to be a bit trickier in terms of getting the `aircraft` directory copied over into Colab.

seanmcleod70:
Looks like 2 main options to get the aircraft data files into Colab, clone the whole repo, seems a bit wasteful, or add some manual directory creation and remote file copying along these lines: 

```
!mkdir aircraft
!cd aircraft
!mkdir A4SAS
!cd A4SAS
!wget https://github.com/JSBSim-Team/jsbsim/blob/master/aircraft/A4/A4.xml
```

seanmcleod70:
Hmm, after some further testing, it looks like I was having issues trying to `cd` within Colab.

So without `cd`, make use of the `-P` option of `wget`.

```
!mkdir aircraft
!mkdir aircraft/A4SAS
!wget -P aircraft/A4SAS/ https://github.com/JSBSim-Team/jsbsim/blob/master/aircraft/A4/A4.xml
```

seanmcleod70:
@ANF1S in terms of the filename, currently `test_pitch_sas.ipynb` I'd suggest giving it a more friendly looking name along the lines of some of the other examples, so maybe `Pitch SAS Implementation.ipynb`.

There seem to be a couple of statements that have been copied over from an existing example that aren't relevant for this example:

```py
from pathlib import Path

# IC references TAS, comment mentions CAS
fdm['ic/vt-kts']    = 250  #calibrated airspeed (kts)

print("Trim failed, continuing rudder kick in an untrimmed state.")
```

ANF1S:
Thanks for the review @seanmcleod70 . I fixed those lines and also changed the name of the notebook. About `engine` , it stays as subdirectory of `aircraft/A4SAS` as suggested.

seanmcleod70:
@ANF1S thanks. So from my side the only outstanding thing is the Google Colab support. If the notebook is going to support being launched from the repo into Colab then we need to make sure it works, or we need to remove the Colab support.

ANF1S:
Hello @seanmcleod70 

I do not have too much experience with Google Colab, but I tried this to test locally: Like other examples in Colab I established the first cell with JSBSim installation. Both `PATH_TO_JSBSIM_FILES="./"` because of the local path in Colab
```python
import os

try:
  import jsbsim
  print('JSBSim is installed')
  # Relative path to the directory where the flight model is stored
  # Note - Aircraft directory needs to be writeable in order to modify the cg
  PATH_TO_JSBSIM_FILES= "./"
  print(f'Current working directory: {os.getcwd()}')
except ImportError:
  import sys
  if 'google.colab' in sys.modules:
      print('Installing JSBSim...')
      %pip install jsbsim
      # Use the standard aircraft supplied by the JSBSim Python package
      PATH_TO_JSBSIM_FILES="./"
```
In the second cell, it continues with the other requiered libraries and when I add the following commands
```python
(...)
AIRCRAFT_NAME="A4SAS"

!wget -P aircraft/A4SAS/ https://raw.githubusercontent.com/ANF1S/jsbsim/fcs-design/examples/python/aircraft/A4SAS/A4SAS.xml
!wget -P aircraft/A4SAS/ https://raw.githubusercontent.com/ANF1S/jsbsim/fcs-design/examples/python/aircraft/A4SAS/pitch_sas.xml

!wget -P aircraft/A4SAS/engine https://raw.githubusercontent.com/ANF1S/jsbsim/fcs-design/examples/python/aircraft/A4SAS/engine/J52.xml
!wget -P aircraft/A4SAS/engine https://raw.githubusercontent.com/ANF1S/jsbsim/fcs-design/examples/python/aircraft/A4SAS/engine/direct.xml

(...)
```
I obtained:

Folders_Colab

And could continue with the simulation well.

With this command I could download the `xml` files and create the folders. This will look different because it was obtained from my branch:
```
!wget -P aircraft/A4SAS/ https://raw.githubusercontent.com/[MY-USER]/jsbsim/[MY-BRANCH]/examples/python/aircraft/A4SAS/A4SAS.xml
```

When I tried this command:
```
!wget -P aircraft/A4SAS/ https://github.com/JSBSim-Team/jsbsim/blob/master/aircraft/A4/A4.xml
```
I think I downloaded `html` code and not the `xml` file

agodemar:
@ANF1S Thanks for the update. If you confirm that the Colab works on your side, I'll merge your example with the official JSBSim master branch

seanmcleod70:
In terms of getting the raw file contents as opposed to the html page I copied the URL from the `Raw` button.

image

Which is:

[https://raw.githubusercontent.com/JSBSim-Team/jsbsim/refs/heads/master/aircraft/A4/A4.xml](https://raw.githubusercontent.com/JSBSim-Team/jsbsim/refs/heads/master/aircraft/A4/A4.xml)

So we would need to just update that URL to match the path in the `examples` directory etc.

> Note - Aircraft directory needs to be writeable in order to modify the cg

You can remove that comment which comes from a specific example that needed to write and update the aircraft's FDM while running the notebook.

Does the `wget` command create the directory tree, so you don't need explicit `mkdir` commands before using `wget`?

ANF1S:
@agodemar , @seanmcleod70 Thanks for your comments. Yes, the Colab works on my side while testing. I will do the updates. 

With respect to `!wget`, it can create directory tree using with the flag `!wget -P` . So it can download the files and create destination folder.

ANF1S:
I applied the requested updates. Let me know if everything looks good on your end.

seanmcleod70:
Hmm, bit of a pain, Github is giving errors when trying to render any Jupyter notebooks - https://github.com/orgs/community/discussions/197350

So I'm glancing through the text of the diff.

Looks like 99% there. 

> Use the standard aircraft supplied by the JSBSim Python package

This comment isn't true for this notebook, we're explicitly using an aircraft outside the JSBSim Python package, hence the `wget` copies we need for Colab.

I was expecting to see the `wget` calls within the `google.colab` if statement, rather than in the next cell.

There are 2 main scenarios.

1. User has cloned/copied the repo locally to their system and they open this Jupyter notebook. In this case there is no need to perform the `wget` commands, since the required aircraft files are already there from the clone/copy. What does need to be checked is that the Python module is installed, since a simple clone/copy doesn't install the Python module.

2. User opens this Jupyter notebook via Google Colab, manually or via a Colab link we provide etc. In this case, in addition to ensuring that the JSBSim Python module is installed we also need to perform the `wget` commands to retrieve the aircraft files. 

So I'd suggest something along these lines.

```python
# Confirm that JSBSim is installed. If not and running on Google Colab
# then install JSBSim into the Google Colab instance.

import os

try:
  import jsbsim
  print('JSBSim Python module is installed')
  # Relative path to the directory where the flight model is stored
  PATH_TO_JSBSIM_FILES= "./"
  print(f'Current working directory: {os.getcwd()}')
except ImportError:
  import sys
  if 'google.colab' in sys.modules:
    print('Installing JSBSim...')
    %pip install jsbsim
    # Download aircraft files
    !wget -P aircraft/A4SAS/ https://raw.githubusercontent.com/JSBSim-Team/jsbsim/refs/heads/master/examples/python/aircraft/A4SAS/A4SAS.xml
    !wget -P aircraft/A4SAS/ https://raw.githubusercontent.com/JSBSim-Team/jsbsim/refs/heads/master/examples/python/aircraft/A4SAS/pitch_sas.xml
    !wget -P aircraft/A4SAS/engine https://raw.githubusercontent.com/JSBSim-Team/jsbsim/refs/heads/master/examples/python/aircraft/A4SAS/engine/J52.xml
    !wget -P aircraft/A4SAS/engine https://raw.githubusercontent.com/JSBSim-Team/jsbsim/refs/heads/master/examples/python/aircraft/A4SAS/engine/direct.xml
  else:
    print("Error: JSBSim Python module is not installed")
```

ANF1S:
Hi @seanmcleod70. Thanks for the feedback. I understood about the scenarios, my mistake. 

For the second scenario, does `PATH_TO_JSBSIM_FILES= "./"` still need to be defined to create the instance `fdm = jsbsim.FGFDMExec(PATH_TO_JSBSIM_FILES)`, right?  Do you think it should go after downloading the aircraft files?

seanmcleod70:
> For the second scenario, does `PATH_TO_JSBSIM_FILES= "./"` still need to be defined to create

I had removed it from inside the `except` block in my suggestion above since I saw it being set in the `try` block above, but forgot that it's the very first statement in the `try` block, i.e. `import jsbsim` that can throw the exception, in which it wouldn't be set.

It needs to be set to the same value for both scenarios, so there is no point setting it twice, which is what I was trying to avoid. We should either set it before the `try` block, or as the first statement inside the `try` block, or after the `except` block.

ANF1S:
Hi @seanmcleod70 . Thanks for your message. I removed the redundant path constant. Let me know if it looks good to you.

seanmcleod70:
Thanks @ANF1S for the contribution, I've merged it into the JSBSim repo.

Hopefully Github will resolve their current Jupyter notebook rendering issue soon.

seanmcleod70:
@ANF1S @agodemar I've also just pushed a commit to JSBSim's `README.md` to add an entry and link to your Jupyter notebook example plus a link to open it into Google Colab, which I've just tested and it all worked.

agodemar:
@ANF1S @seanmcleod70 great job!
Great adddition to the examples collection.
This is yet another good example of open source best practice: involving users or developers in the public review space

ANF1S:
@seanmcleod70 @agodemar Thanks for your guidance during the review. I'm glad I could contribute to JSBSim!

PR #1465, opened 2026-05-31
Reject duplicate input sockets with the same port and protocol

Zaretto:
FGInput::Load registers each <input> socket unconditionally, so defining an input socket twice tries to bind the same port again and fails. 

Expose the port and protocol on FGInputSocket and skip a new socket whose port and protocol match one already registered, warning and discarding the duplicate.

### DIFF
--- a/src/input_output/FGInputSocket.h
+++ b/src/input_output/FGInputSocket.h
@@ -70,6 +70,11 @@ public:
   */
   bool Load(Element* el) override;
 
+  /// @return the port this input socket is bound to (0 if not yet loaded).
+  unsigned int GetPort(void) const { return SockPort; }
+  /// @return the protocol (TCP/UDP) this input socket uses.
+  FGfdmSocket::ProtocolType GetProtocol(void) const { return SockProtocol; }
+
   /** Initializes the instance. This method basically opens the socket to which
       inputs will be directed.
       @result true if the execution succeeded.
--- a/src/models/FGInput.cpp
+++ b/src/models/FGInput.cpp
@@ -40,6 +40,7 @@ INCLUDES
 
 #include "FGInput.h"
 #include "FGFDMExec.h"
+#include "input_output/FGInputSocket.h"
 #include "input_output/FGUDPInputSocket.h"
 #include "input_output/FGXMLFileRead.h"
 #include "input_output/FGModelLoader.h"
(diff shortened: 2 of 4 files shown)


seanmcleod70:
Isn't this the tip of the iceberg?

What exactly was your load/re-load procedure? It appears that the previous set of inputs weren't discarded/cleaned up in-between. But surely a number of other models will end up with similar duplicate instance issues then as well? For example a duplicate output writing to the same .csv file. Duplicate aerodynamic forces and moments?

seanmcleod70:
Just to be clear, I'm referring to your 'reload' comment.

Which is somewhat separate to the case where the user accidently adds a duplicate entry in the FDM XML, e.g. 2 input socket types with the same protocol/port.

Zaretto:
Thanks for the feedback.

Sorry, the comment was misleading.

What I actually meant was a duplicate which in my case was caused by having an input in a XML script and the same input also in the main XML. 

What happens then:
- two FGInputSocket instances are created
- the second bind() fails in InitModel with only a low-level "could not bind" errno
- that input is silently dead

It may be the tip of an iceberg but this one genuinely caused me a problem that required debugging and cost me time so I thought it was worth fixing. This isn't a pre-emptive rework of code that might fail - this did fail.

seanmcleod70:
No problem. So, for your particular case, duplicate FGInputSocket across aircraft FDM and script XML I see much less of an iceberg compared to a reload resulting in duplicate entries in lots of models.

I'm a bit confused how in your particular case the duplicate FGInputSocket actually caused an issue. Didn't you still end up with 1 good working FGInputSocket instance, so any packets sent to that particular protocol-port by an external program would still get through and be handled by the 1 good working instance?

Zaretto:
You're right and I'm sorry for wasting your time.

With the duplicate you still get one working FGInputSocket, and anything sent to that port and protocol is handled by it. The second one fails to bind and sits there dead, but it doesn't affect the working one, so nothing actually breaks.

I just tried to reproduce the problems I had (which I remembered causing a heap corruption) with the FGScript scenario and I can't get it to break. Neither can I break it with code that creates an `<input>`.

So I don't think there's anything worth fixing here, and I'll withdraw it.

Thanks for taking the time to walk through it.

seanmcleod70:
No problem. I had recently spent some time in this code for this PR - https://github.com/JSBSim-Team/jsbsim/pull/1459

PR #1469, opened 2026-06-04
Sr75 jsbsim baseline

saipoojith-creator:


agodemar:
@saipoojith-creator thank you for your request. I think that, before your PR is approved, you could provide an example of use. 
Before flowing into the official repository, data should undergo at least some basic tests. Trim, for instance; or turns, or controlled maneuvers. Throwing data into the `aircraft` folder with no examples of use is not appropriate, in my opinion.

@bcoconni @seanmcleod70 correct?

seanmcleod70:
@agodemar I'm 90% certain that this type of PR, which we've had in the past, is the result of an error by the user of their git tool, i.e. they didn't mean to issue a PR, rather than were trying to merge branches in their repo, create a PR for their own repo etc.

But yes, at a minimum we won't even look or consider a PR until it includes a description of what the PR is about.

bcoconni:
Closing the PR since the author is not responding.

@saipoojith-creator you are welcome to re-open it when you think you are in a position to address the comments that @agodemar and @seanmcleod70 rightfully made above.

Issue #1466, opened 2026-06-01
FGBrushlessDCMotor calculates too much output power


Scavanger:
**I'm submitting a ...**
  - [X] bug report

I’m currently experimenting with fixed-wing UAV/RC models and have noticed that FGBrushlessDCMotor calculates the output power far too high.

For example, for this motor with an 18x8 propeller
````xml
<?xml version="1.0"?>
  <brushless_dc_motor name="Hacker A60-5XS">
  <velocityconstant> 420 </velocityconstant>
  <coilresistance>0.012</coilresistance>
  <noloadcurrent>2.40</noloadcurrent>
  <maxvolts> 25.2</maxvolts>
</brushless_dc_motor>
````

According to JSBSim, this results in a maximum power output of just under 6 kW, with a current draw of almost 300 A, which is almost three times higher than the manufacturer’s specification (2090 W, at under 100 A).

**What is the expected behavior?**
Depending on the propeller, the shaft power should be a good 1500 - 2100 W; JSBSim is a long way off that.

**What could be causing the error?**
I can’t see any major errors in FGBrushlessMotor, but what could be causing the problem is that no Reynolds number correction is applied in FGPropeller, so with such small propellers, the RPM calculation is incorrect; see https://m-selig.ae.illinois.edu/pubs/DetersAnandaSelig-2014-AIAA-2014-2151.pdf

**Proposed solution**
Two optional tables in the Propeller XML that contain a Reynolds number factor for thrust and power.

bcoconni:
It is most likely related to the fidelity of the first order model that JSBSim is using for brushless DC motors. According to the [Hacker A60-5XS specs](https://www.hacker-motor-shop.com/Brushless-Motors/Hacker-Outrunner/Hacker-A60/A60-motor-planes/A60-5XS-V4-28-Pole-kv420.htm?shop=hacker_e&a=article&ProdNr=33726633&p=7013):

| Propeller | APC 18x8| APC 17x8|
|-|-|-|
| Voltage [V] | 22 | 22 |
| Current [A] | 95 | 85 |
| Power [W] | 2090 | 1870 |
| RPM | 7600 | 7800 |

I could match the above value with Kv=436 RPM/V and R=0.048 Ohms.

Note that it needs the resistance to be quadrupled while the velocity constant is only slightly modified. I guess that departing from the official specifications to match the actual performance of the engine is a way to work around the simplicity of the 1st order model that JSBSim uses for electric motors.

[...]

PR #1470, opened 2026-06-08
Reynolds number correction for propellers + Script for converting APC PER3 files

Scavanger:
Adds the missing Reynolds numbers correction in FGPropeller. This applies to small UAV/RC propellers with low Reynolds numbers and should result in a 10–15% improvement.

In the propeller XML, two additional tables, CT_RPM_FACTOR and CP_RPM_FACTOR, have been added, these are  normalised to 1.0 at the RPM for which the baseline C_THRUST/C_POWER data were measured.

I have also added a script that converts APC PER3 data files into clean propeller XML files and calculates the corrected Reynolds numbers and an example XML for a APC 18x8E Propeller.

Unfortunately, XML files for small propeller XML files generated by Aero-matic are incorrect, as Aero-matic is not designed for such small propellers and calculates a shaft power that is far too high. 

https://github.com/JSBSim-Team/jsbsim/issues/1466


bcoconni:
Thanks for your contribution.

It looks good to me.

Using RPMs to get the Reynolds number correction makes perfect sense because it is assumed that:
* The air viscosity is constant which I think is a reasonable assumption for UAVs/MAVs which are designed to fly at low altitudes.
* The propeller blade velocity is mainly driven by its circumferential velocity due to the vehicle velocity being negligible in respect. 
 
These 2 assumptions stand because we are dealing with UAV/MAV with small propellers and because the Reynolds number correction only applies to (very) small propellers. So I am wondering if this should be documented ?

In particular, tables `CT_RPM_FACTOR` and `CP_RPM_FACTOR` would need to be tuned if one propeller would be used to model [Ingenuity](https://en.wikipedia.org/wiki/Ingenuity_(helicopter)) flying in the Mars atmosphere and for another UAV on Earth. That is now a possibility following the PR #1462 😉

bcoconni:
@seanmcleod70 any comments on that PR?

seanmcleod70:
Looks good to me. 

> These 2 assumptions stand because we are dealing with UAV/MAV with small propellers and because the Reynolds number correction only applies to (very) small propellers. So I am wondering if this should be documented ?

Yes I think it would be useful to add the additional documentation.

PR #1472, opened 2026-06-12
Fix pkgconfig libs so they don't include CMake object/interface libraries

blazingpretzel:
A recent update to the CMake configuration made it where the `libJSBSim` receives the contents of the various object libraries using the `target_link_libraries()` command, instead of the previously used generator statements to pull the object files directly into the library.

However, because CMake now sees multiple "libraries" being linked to `libJSBSim`, this caused script that generates the pkg-config file to fail, as it simply adds the name of each library to the "libs" section, regardless of whether it's a real static/shared library, vs CMake's virtual object/interface libraries.

This change attempts to fix that by checking if a library is a static or shared library before adding it to the list of pkg-config libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,17 @@ if (PkgConfig_FOUND)
   if (NOT BUILD_SHARED_LIBS)
     get_target_property(libJSBSim_LINK_LIBRARIES libJSBSim LINK_LIBRARIES)
     foreach(_LIB ${libJSBSim_LINK_LIBRARIES})
+
+      # Ensure this library is not a CMake object library or interface library
+      if (TARGET ${_LIB}) # Non-CMake libraries are ok (specifically libm)
+        get_target_property(LINK_LIB_TYPE ${_LIB} TYPE)
+        if (NOT ((${LINK_LIB_TYPE} STREQUAL "STATIC_LIBRARY") OR (${LINK_LIB_TYPE} STREQUAL "SHARED_LIBRARY")))
+          continue()
+        endif()
+      endif()
+
       set(JSBSIM_PKG_CONFIG_LINK_LIBRARIES "${JSBSIM_PKG_CONFIG_LINK_LIBRARIES} -l${_LIB}")
+
     endforeach(_LIB)
   endif(NOT BUILD_SHARED_LIBS)
   configure_file(${CMAKE_CURRENT_SOURCE_DIR}/JSBSim.pc.in ${CMAKE_CURRENT_BINARY_DIR}/JSBSim.pc @ONLY)


CMakeLists.txt
```diff
@@ -135,7 +135,20 @@ if (PkgConfig_FOUND)
...
     get_target_property(libJSBSim_LINK_LIBRARIES libJSBSim LINK_LIBRARIES)
     foreach(_LIB ${libJSBSim_LINK_LIBRARIES})
-      set(JSBSIM_PKG_CONFIG_LINK_LIBRARIES "${JSBSIM_PKG_CONFIG_LINK_LIBRARIES} -l${_LIB}")
+
+      # Ensure this library is not a CMake object library or interface library
+      set(AddTarget true)
+      if (TARGET ${_LIB}) # Non-CMake libraries are ok (specifically libm)
+        get_target_property(LinkLib_TYPE ${_LIB} TYPE)
+        if (NOT ((${LinkLib_TYPE} STREQUAL "STATIC_LIBRARY") OR (${LinkLib_TYPE} STREQUAL "SHARED_LIBRARY")))
+          set(AddTarget false)
```
bcoconni:
```suggestion
          continue()
```
Use the [CMake command `continue()`](https://cmake.org/cmake/help/v4.3/command/continue.html) to skip the occurrences of `${_LIB}` that are static or shared libraries. This will allow to remove the variable `AddTarget` and all the logic using it.

CMakeLists.txt
```diff
@@ -135,7 +135,20 @@ if (PkgConfig_FOUND)
   if (NOT BUILD_SHARED_LIBS)
     get_target_property(libJSBSim_LINK_LIBRARIES libJSBSim LINK_LIBRARIES)
     foreach(_LIB ${libJSBSim_LINK_LIBRARIES})
-      set(JSBSIM_PKG_CONFIG_LINK_LIBRARIES "${JSBSIM_PKG_CONFIG_LINK_LIBRARIES} -l${_LIB}")
+
+      # Ensure this library is not a CMake object library or interface library
+      set(AddTarget true)
+      if (TARGET ${_LIB}) # Non-CMake libraries are ok (specifically libm)
+        get_target_property(LinkLib_TYPE ${_LIB} TYPE)
```
bcoconni: Please use `LINK_LIB_TYPE` instead of `LinkLib_TYPE` as the convention for variable naming is to use `SNAKE_CASE_WITH_CAPITAL_LETTERS` except when using target names which allow using `CamelCase`. In this case, `LinkLib` does not refer to a target name hence you should use `LINK_LIB_TYPE` for the variable name.

CMakeLists.txt
```diff
@@ -135,7 +135,20 @@ if (PkgConfig_FOUND)
...
+        if (NOT ((${LinkLib_TYPE} STREQUAL "STATIC_LIBRARY") OR (${LinkLib_TYPE} STREQUAL "SHARED_LIBRARY")))
+          set(AddTarget false)
+        endif()
+      endif()
+
+      if (${AddTarget})
```
bcoconni:
```suggestion
```
Using `continue()` above allows to remove the `if()` here.

bcoconni (changes requested):
Thanks for your contribution.

I suggest a couple of changes to simplify the code and make it consistent with the rest of the CMake file.

bcoconni:
Also for the record, the recent update that is mentioned is the PR #1458.

CMakeLists.txt
blazingpretzel: Oh cool, today I learned CMake added `continue()`. Thanks!

blazingpretzel:
Added and pushed suggested changes.

bcoconni (approved):
Thanks for the quick update @blazingpretzel.
Looks good to me.

bcoconni:
The PR is now merged. Thanks @blazingpretzel for your contribution.

PR #1474, opened 2026-06-16
Include the offending path in the property-name parse exception

Zaretto:
The property-name parser throws when a name contains a disallowed character but does not indicate which property path triggered it. Append the path to the exception message so the offending property can be identified.


seanmcleod70:
So it looks like elsewhere in the same function some identifying information is included as part of the exception of the message. Any reason for not being consistent in terms of how the string is built for the exception message?

https://github.com/JSBSim-Team/jsbsim/blob/0d44e1253435780acddac8e9a05dd8b6c08ff6a4/src/simgear/props/props.cxx#L191

What about this case lower down in the same function, should we not also include the path name in the exception to help users figure out which path name is problematic?

https://github.com/JSBSim-Team/jsbsim/blob/0d44e1253435780acddac8e9a05dd8b6c08ff6a4/src/simgear/props/props.cxx#L215

bcoconni:
I agree with @seanmcleod70 that all the `throw` instances in that function shall be treated equally.

Zaretto:
I also agree with you both so I'll make the fixes.

Zaretto:
I have made all the exceptions in the active parser consistent:
- Each message is now built the same way, inside the string constructor.
- Each appends the offending path, quoted so it stays readable if the path contains a space or a colon.
- This covers the three throws in parse_name and the neighbouring "unterminated index" throw in parse_index, so a
bad path is reported wherever parsing fails.

The change is confined to the parser compiled here (PROPS_STANDALONE).

PR #1475, opened 2026-06-16
Do not prepend the output path to an already-absolute output name

Zaretto:
Fix issue when FGOutputFile::SetOutputName joins the configured output path with the supplied name, which produces a malformed path when the name is itself absolute. 

i.e.; Take an absolute name as-is; only relative names are resolved under the output path.

### DIFF
--- a/src/input_output/FGOutputFile.h
+++ b/src/input_output/FGOutputFile.h
@@ -100,7 +100,10 @@ public:
       the next call to SetStartNewOutput().
       @param name new name */
   void SetOutputName(const std::string& fname) override {
-    Name = (FDMExec->GetOutputPath()/fname).utf8Str();
+    // An absolute name is used as-is; only a relative name is taken under the
+    // output path.
+    SGPath p(fname);
+    Name = p.isAbsolute() ? fname : (FDMExec->GetOutputPath()/fname).utf8Str();
     runID_postfix = -1;
     Filename = SGPath();
   }
--- a/src/input_output/FGOutputFile.h
+++ b/src/input_output/FGOutputFile.h
@@ -95,13 +95,13 @@ public:
   */
   void SetStartNewOutput(void) override;
   /** Overwrites the name identifier under which the output will be logged.
+      If fname is a relative filepath then prepend the configured output path,
+      otherwise use the absolute filepath.
       For this method to take effect, it must be called prior to
       FGFDMExec::RunIC(). If it is called after, it will not take effect before
       the next call to SetStartNewOutput().
-      @param name new name */
+      @param fname new name */
   void SetOutputName(const std::string& fname) override {
-    // An absolute name is used as-is; only a relative name is taken under the
-    // output path.
     SGPath p(fname);
     Name = p.isAbsolute() ? fname : (FDMExec->GetOutputPath()/fname).utf8Str();
     runID_postfix = -1;


bcoconni:
I am in two minds with regard to this PR, do we really want to bypass the output path using absolute paths?
* @Zaretto could you please describe the use case where you need to specify an absolute path rather than setting the output path to the required location?
* @seanmcleod70 @agodemar any thoughts?

Zaretto:
On Linux and similar operating systems you can already do this implicitly by setting `name="/folder/file"` however that doesn't work on Windows where I want to do `name="c:/folder/file"`. 

I had problems using scripts on Windows to create output where it needed to be rather than in the aircraft (or script folder) that prevented me from using the scripts in a single library for testing variants of models.

I guess this is more of a bugfix than a new feature.

bcoconni:
> I had problems using scripts on Windows to create output where it needed to be rather than in the aircraft (or script folder)

There is the method `FGFDMexec::SetOutputPath` that allows to avoid exactly that:

https://github.com/JSBSim-Team/jsbsim/blob/32aa2af6b4da611d4f92cc0e44139f3d39efc16f/src/FGFDMExec.h#L342-L351

Are you using it ? It's a much more portable and versatile way to prevent your aircraft and script folders from being *polluted* by output files.

seanmcleod70:
But there doesn't seem to be a command line argument, unless I'm missing it, to set the output path?

So it looks like JSBSim defaults to the current directory?

https://github.com/JSBSim-Team/jsbsim/blob/32aa2af6b4da611d4f92cc0e44139f3d39efc16f/src/JSBSim.cpp#L397

And if you're writing a C++ wrapper or using the Python wrapper then you can call `FGFDMExec::SetOutputPath()`, but if you're running the standard JSBSim executable then you're stuck with the current directory?

https://github.com/JSBSim-Team/jsbsim/blob/32aa2af6b4da611d4f92cc0e44139f3d39efc16f/src/JSBSim.cpp#L842-L863

My understanding is that @Zaretto had script files/aircraft files with `<output/>` elements and wanted the ability to specify a full path in the `<output/>` element. Although I'm not sure how he's running things, i.e. just the regular JSBSim executable or not.

```xml
<output name="B737_datalog.csv" type="CSV" rate="20">
   <property> velocities/vc-kts </property>
   <velocities> ON </velocities>
</output>
```

@Zaretto I'm not sure why you don't think this is an issue in Linux since your example `name="/folder/file"` surely will only work once the concatenation is done resulting in `./folder/file` if `folder` is a sub-directory of the current working directory or the current working directory is the root of the file system.

But if you want an absolute path from the root then you will also run into this issue on Linux.

bcoconni:
> but if you're running the standard JSBSim executable then you're stuck with the current directory?

True. But that will no longer be the case once the PR #1478 will be merged 😉

Note also that as you mentioned, the output files are by default written to the current directory when using the executable `JSBSim.exe`. And in that case, the output files would not be in the aircraft or scripts folder.

seanmcleod70:
Hmm, in terms of 'current' directory, I had thought of mentioning above that @Zaretto could potentially change the current directory to where he wanted his output file to go, but I only just tried that for interest now with:

```sh
C:\source\temp\FlightTests>C:\source\jsbsim\Debug\JSBSim.exe --root=\source\jsbsim\ C:\source\jsbsim\scripts\737_cruise.xml
```

Where `737_cruise.xml` has `<output name="737_cruise_output.csv" type="CSV" rate="10"....`

However I noticed that `737_cruise_output.csv` didn't appear in `C:\source\temp\FlightTests`, instead it appeared in the root directory that I had specified.

I guess the `.\737_cruise_output.csv` is being re-directed relative to the root via this code, i.e. so it's not relative to the current directory?

https://github.com/JSBSim-Team/jsbsim/blob/d89e357b77594fb464090ddf465da3e56e523c6f/src/FGFDMExec.h#L715-L720

Zaretto:
> Are you using it? It's a much more portable and versatile way to prevent your aircraft and script folders from being _polluted_ by output files.

Yes, I am, and keeping the aircraft and script folders clean was exactly part of what I needed to solve. So we agree on the goal.

My problem was twofold.

First, setting the output to an absolute path on Windows (including the drive specification) produced mangled paths. An absolute name only works while the output path is unset. Once an output path is set, the name is appended to it:

- POSIX: `/abs/file.csv` is silently appended to the output path as `<outdir>/abs/file.csv`
- Windows: the result is `i:/outdir/i:/abs/file.csv`, a drive letter mid-path, which cannot be opened

That is why I see this as a bug fix rather than a feature change. Absolute output names are already accepted, they are just silently corrupted once an output path is also set.

Second, I genuinely needed the output in two different places on disk, and a single output path only gives me one. For example, the parameter log from the test script goes to one location and the aircraft debug files to another. At the time I could not build JSBSim from source and had to use a prebuilt executable, so I worked around it with a batch file that ran JSBSim and then shuffled the output into the right places afterwards. Honouring absolute names lets JSBSim write each output to its final location directly and removes that step.

bcoconni (approved):
OK. Thanks for the explanations @Zaretto.
Looks good to me.

bcoconni:
> I guess the `.\737_cruise_output.csv` is being re-directed relative to the root via this code, i.e. so it's not relative to the current directory?
> 
> https://github.com/JSBSim-Team/jsbsim/blob/d89e357b77594fb464090ddf465da3e56e523c6f/src/FGFDMExec.h#L715-L720

Indeed. I had forgotten this subtlety.

seanmcleod70:
I'd suggest a slight rewording of the comment:

> // An absolute name is used as-is; only a relative name is taken under the
    // output path.

```c++
// If fname is a relative filepath then prepend the configured output path, 
// otherwise use the absolute filepath.
```

bcoconni:
> I'd suggest a slight rewording of the comment:
> 
> > // An absolute name is used as-is; only a relative name is taken under the
> > // output path.
> 
> ```c++
> // If fname is a relative filepath then prepend the configured output path, 
> // otherwise use the absolute filepath.
> ```

Yep! And move it in the Doxygen comments/docs just above:
https://github.com/JSBSim-Team/jsbsim/blob/c9d0f18f2929ec626ba5f973e5352abbb5f37c7a/src/input_output/FGOutputFile.h#L97-L102

agodemar:
Great catch @Zaretto, I also had similar problems. The PR looks good to me too.

seanmcleod70:
Looks good.

PR #1478, opened 2026-06-21
Add an option `--outputpath` to the JSBSim scripts

bcoconni:
This PR adds an option `--outputpath` to the binary executable `JSBSim` and the Python script `JSBSim.py`. This allows to specify from the command line the output path where all the output files will be written.

I took this opportunity to:
* change the order the arguments are displayed in and group them by themes (input, output, miscellaneous)
* Fix the Python module name (was `_jsbsim` but should be `jsbsim`). For some reason, no one complained about that, most likely because `JSBSim.py` is not known or used.

### DIFF
--- a/python/JSBSim.py
+++ b/python/JSBSim.py
@@ -48,7 +48,7 @@ def main():
                         help="sets (overrides) the name of a data output file")
     parser.add_argument("--logdirectivefile", action="append", metavar="<filename>",
                         help="specifies the name of a data logging directives file")
-    parser.add_argument("--outputpath", metavar="<filename>",
+    parser.add_argument("--outputpath", metavar="<path>",
                         help="specifies the directory where the output files will be written.")
     parser.add_argument("--realtime", default=False, action="store_true",
                         help="specifies to run in real world time")
--- a/python/JSBSim.py
+++ b/python/JSBSim.py
@@ -27,23 +27,25 @@ import sys
 import time
 import xml.etree.ElementTree as et
 
-import _jsbsim
+import jsbsim
 
 def main():
     parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter)
     parser.add_argument("input", nargs='?', help="script file name")
     parser.add_argument("--version", action="version",
-                        version="%(prog)s {}".format(_jsbsim.FGJSBBase().get_version()))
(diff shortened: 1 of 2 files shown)


src/JSBSim.cpp
```diff
@@ -842,18 +851,19 @@ void PrintHelp(void)
...
-    cout << "                                   (can appear multiple times)" << endl;
     cout << "    --root=<path>  specifies the JSBSim root directory (where aircraft/, engine/, etc. reside)" << endl;
     cout << "    --aircraft=<filename>  specifies the name of the aircraft to be modeled" << endl;
     cout << "    --script=<filename>  specifies a script to run" << endl;
+    cout << "    --initfile=<filename>  specifies an initialization file" << endl;
+    cout << "    --planet=<filename>  specifies a planet definition file" << endl;
+    cout << "    --outputlogfile=<filename>  sets (overrides) the name of a data output file" << endl;
+    cout << "    --logdirectivefile=<filename>  specifies the name of a data logging directives file" << endl;
+    cout << "                                   (can appear multiple times)" << endl;
+    cout << "    --outputpath=<filename> specifies the directory where the output files will be written." << endl;
```
seanmcleod70: `<path>` instead of `<filename>`

python/JSBSim.py
```diff
@@ -27,23 +27,29 @@
...
                         help="specifies a script to run")
+    parser.add_argument("--initfile", metavar="<filename>",
+                        help="specifies an initialization file")
+    parser.add_argument("--planet", metavar="<filename>",
+                        help="specifies a planet definition file")
+    parser.add_argument("--outputlogfile", action="append", metavar="<filename>",
+                        help="sets (overrides) the name of a data output file")
+    parser.add_argument("--logdirectivefile", action="append", metavar="<filename>",
+                        help="specifies the name of a data logging directives file")
+    parser.add_argument("--outputpath", metavar="<filename>",
```
seanmcleod70: `<path>` instead of `<filename>`.

seanmcleod70:
Looks good.

src/JSBSim.cpp
bcoconni: Done.

python/JSBSim.py
bcoconni: Done.

bcoconni:
Thanks for the review.
PR merged.

Issue #1390, opened 2026-02-11
Requesting a getter function in Python interface: fdm.get_propagate().get_state()


agodemar:
**I'm submitting a feature request**

**Describe the issue**
I am using the Python interface to JSBSim. When I instantiate the object `FGFDMExec` like so:

```Python
PATH_TO_JSBSIM_FILES="../.." # I'm in jsbsim_root/examples/python
AIRCRAFT_NAME="c172x"

# Instantiate the FDMExec object and load the aircraft
fdm = jsbsim.FGFDMExec(PATH_TO_JSBSIM_FILES)

load_status = fdm.load_model(AIRCRAFT_NAME)
```
I do get access to the low-level FGPropagate object:
```Python
fg_propagate = fdm.get_propagate()
fg_propagate.get_uvw()
fg_propagate.get_Tec2b()
fg_propagate.get_Tl2b()
# fg_propagate.get_state() # not defined yet
```
but at the moment only the 3 above getter functions are available.

How to add the new getter function `get_state()` that will return a Python object corresponding to the public C++ struct `FGPropagate::VehicleState`?

In particular, I'm interested into getting the objects containing the state derivatives: `dqPQRidot`, `dqUVWidot`, `dqInertialVelocity`, `dqQtrndot` (stored as `std::deque`-s).

seanmcleod70:
I'm no expert, but take a look at the following 2 snippets.

https://github.com/JSBSim-Team/jsbsim/blob/3003423070d7c2791ec28d7ab867cf39b4f57e26/python/jsbsim.pxd#L168-L173

https://github.com/JSBSim-Team/jsbsim/blob/3003423070d7c2791ec28d7ab867cf39b4f57e26/python/jsbsim.pyx.in#L91-L125

agodemar:
@seanmcleod70 thanks for the pointers. I intend to develop a Python function that returns the time derivatives of all state variables. This will help to integrate JSBSim python module within the PathView project, once the Flask backend will be released.

As I see, after a `fdm.run()`, from the property manager we can get all the variables I'm interested in.

bcoconni:
[...]

Issue #86, opened 2018-07-02
FlightGear: atmosphere/P-sl-psf stuck.


mike402:
Since around FG 2018.1, "atmosphere/P-sl-psf" seems to no longer reflect weather changes.

I had been using it in a hack to keep abrupt jumps in atmo pressure from disturbing my autopilot (as discussed in this thread: https://sourceforge.net/p/flightgear/mailman/message/36299722/), but now I had to throw on more epicycles, see history of the channel "Aero/atmosphere kludge" in this file:

https://gitlab.com/mdanil/Tu-144/blob/master/Tu-144-jsbsim/functions.xml

mike402:
Yes, in fact I have been using flightgear -next branch for a while. The property is always "2116.228" and does not change with METAR or manual QNH setting in weather dialog.

It is just one channel, not the whole file, but I should have summarized to make it clear.

Before I could smooth the pressure by doing something like this:

1. Calculate altitude offset (Hsl-abrupt) from atmosphere/P-sl-psf, apply lag_filter.
2. Sum Hsl-abrupt with position/h-sl-meters to produce indicated altitude above sea level (Habs-m)
3. Either use this altitude directly in systems, or calculate the smoothed pressure from it.

Now I have to:
1. Calculate indicated altitude above sea level (Habs-abrupt-m) from "jerky" pressure atmosphere/P-psf.
2. Subtract position/h-sl-meters to get Hsl-abrupt, apply lag_filter.
3. Do the same as in steps 2 and 3 above.

seanmcleod:
@mike402 what Jsbsim atmosphere property or properties do the METAR and QNH settings in Flight Gear modify?

mike402:
For sure atmosphere/P-psf and atmosphere/pressure-altitude, there may be others. I think they indirectly affect aero/qbar-psf etc as well.

seanmcleod:
@mike402 could you confirm, I don't have a copy of FlightGear handy.

If you take a look at the JSBSim commits over the last month or so you'll notice that a lot of them were around updates to the standard atmosphere model.

Now if you take a look at `atmosphere/P-psf` and `atmosphere/pressure-altitude` you'll notice that there are only getters provided by JSBSim, so FlightGear's METAR and QNH weather settings can't be modifying them directly.

```c++
PropertyManager->Tie("atmosphere/P-psf", this, &FGAtmosphere::GetPressure);
[...]

PR #1481, opened 2026-06-23
Add a per-model execution-enable property

Zaretto:
Implements per-model enable proposed by @bcoconni in #1390.

The suggested solution (https://github.com/JSBSim-Team/jsbsim/issues/1390#issuecomment-3901613611):
- add a property to each model, managed by FGModel
- when false, FGModel::Run() returns true and the model is skipped
- name it simulation/models/{Name}/enabled

This branch:
- adds simulation/models/<name>/enabled for every scheduled model, default true
- checks it in FGModel::Run() via the existing "skip this frame" return that every model already honours
- binds one property per model in FGFDMExec, in eModels order

Behaviour is unchanged unless a model is explicitly disabled. A disabled model's body does not run, but its last state and property bindings are preserved, so an external system can take over that model in place, for example an external propagator owning FGPropagate.

Two notes:
- the names are fixed in eModels order rather than read from FGModel::Name, because some models rename themselves when their config loads. That is the only deviation from the {Name} form in the original suggestion.
- a static_assert keeps the name table in step with the eModels enum

### DIFF
--- a/src/models/FGModel.cpp
+++ b/src/models/FGModel.cpp
@@ -56,7 +56,8 @@ GLOBAL DECLARATIONS
 CLASS IMPLEMENTATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
-FGModel::FGModel(FGFDMExec* fdmex, const std::string& enableName)
+FGModel::FGModel(FGFDMExec* fdmex, const std::string& name)
+ : Name(name)
 {
   FDMExec     = fdmex;
 
@@ -68,7 +69,14 @@ FGModel::FGModel(FGFDMExec* fdmex, const std::string& enableName)
   exe_ctr     = 1;
   rate        = 1;
 
-  if (!enableName.empty()) BindModelEnabled(enableName);
+  // A model that is scheduled by FGFDMExec is named and gets an enable property.
+  // The input and output types are not scheduled: they are instantiated one per
+  // <input> and <output> element, so a shared property path would be tied more
+  // than once. They pass an empty name and are enabled through the per-instance
+  // property that FGOutputType::SetIdx already binds, or through Enable() and
+  // Disable().
+  if (!Name.empty())
+    PropertyManager->Tie("simulation/models/" + Name + "/enabled", &enabled);
 
   Debug(0);
 }
@@ -90,26 +98,11 @@ bool FGModel::InitModel(void)
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-void FGModel::BindModelEnabled(const std::string& name)
-{
-  const string path = "simulation/models/" + name + "/enabled";
-  const bool existed = PropertyManager->HasNode(path);
-  ModelEnabled = PropertyManager->GetNode(path, true);
-  // Default to enabled; a freshly created node would otherwise read false.
-  if (!existed && ModelEnabled) ModelEnabled->setBoolValue(true);
-}
-
-//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-
 bool FGModel::Run(bool Holding)
 {
   FGModel::Debug(2);
 
-  // A model whose simulation/models/<name>/enabled property is false is skipped
-  // entirely: its Run() body does not execute, but its last state and property
-  // bindings are preserved so an external system can replace it, for example
-  // external propagation or host-owned ground reactions.
-  if (ModelEnabled && !ModelEnabled->getBoolValue()) return true;
+  if (!enabled) return true;
 
   if (rate == 1) return false; // Fast exit if nothing to do
 
--- a/src/models/FGModel.cpp
+++ b/src/models/FGModel.cpp
@@ -102,14 +102,13 @@ bool FGModel::Run(bool Holding)
 {
   FGModel::Debug(2);
 
-  if (!enabled) return true;
-
-  if (rate == 1) return false; // Fast exit if nothing to do
+  if (rate == 1) return !enabled; // Fast exit if nothing to do
 
   if (exe_ctr >= rate) exe_ctr = 0;
 
-  if (exe_ctr++ == 1) return false;
-  else              return true;
+  if (exe_ctr++ == 1) return !enabled;
+
+  return true;
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/src/models/FGModel.cpp
+++ b/src/models/FGModel.cpp
@@ -56,7 +56,7 @@ GLOBAL DECLARATIONS
 CLASS IMPLEMENTATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
-FGModel::FGModel(FGFDMExec* fdmex)
+FGModel::FGModel(FGFDMExec* fdmex, const std::string& enableName)
 {
   FDMExec     = fdmex;
 
@@ -67,6 +67,8 @@ FGModel::FGModel(FGFDMExec* fdmex)
 
   exe_ctr     = 1;
   rate        = 1;
+
+  if (!enableName.empty()) BindModelEnabled(enableName);
 
   Debug(0);
 }
--- a/src/models/FGModel.cpp
+++ b/src/models/FGModel.cpp
@@ -42,6 +42,7 @@ INCLUDES
 #include "FGFDMExec.h"
 #include "input_output/FGModelLoader.h"
 #include "input_output/FGLog.h"
+#include "input_output/FGPropertyManager.h"
 
 using namespace std;
 
@@ -87,9 +88,26 @@ bool FGModel::InitModel(void)
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+void FGModel::BindModelEnabled(const std::string& name)
+{
+  const string path = "simulation/models/" + name + "/enabled";
+  const bool existed = PropertyManager->HasNode(path);
+  ModelEnabled = PropertyManager->GetNode(path, true);
+  // Default to enabled; a freshly created node would otherwise read false.
+  if (!existed && ModelEnabled) ModelEnabled->setBoolValue(true);
+}
+
+//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
 bool FGModel::Run(bool Holding)
 {
   FGModel::Debug(2);
+
+  // A model whose simulation/models/<name>/enabled property is false is skipped
+  // entirely: its Run() body does not execute, but its last state and property
+  // bindings are preserved so an external system can replace it, for example
+  // external propagation or host-owned ground reactions.
+  if (ModelEnabled && !ModelEnabled->getBoolValue()) return true;
 
   if (rate == 1) return false; // Fast exit if nothing to do
 
--- a/src/models/FGModel.h
+++ b/src/models/FGModel.h
@@ -48,6 +48,8 @@ INCLUDES
 FORWARD DECLARATIONS
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
+class SGPropertyNode;
+
 namespace JSBSim {
 
 class FGFDMExec;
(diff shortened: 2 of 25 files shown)


seanmcleod70:
I took a look at #1390 to refresh my memory of some of the earlier discussions and came across my comment at the time towards the end of the thread, see - https://github.com/JSBSim-Team/jsbsim/issues/1390#issuecomment-3902040361

Which it doesn't look like it was answered at the time, so my questions still stand.

bcoconni:
I agree with you @seanmcleod70 and in addition to that, I am not too happy with the implementation. I am really struggling about why the enable/disable feature should imply modifying the `FGFDMExec` class and not be totally delegated to the classes that inherit from `FGModel` ? 🤔

> [...] some models rename themselves when their config loads. That is the only deviation from the {Name} form in the original suggestion.

This does not look like a reason to me for centralizing the feature in `FGFDMExec`.

Zaretto:
Taking the general question about this and #1482 - I agree that a single MR is better.

> #1390 ... doesn't look like it was answered at the time, so my questions still stand.

@seanmcleod70  the points you raised in #1390 about `Auxiliary` are the reason why my approach changed because you were right that replacing `Auxiliary` with a handful of setters only appeared to be working. 

So this change together with #1482 are what I've ended up with because the better design is actually to run most of the models except the one that absolutely must be replaced which in this case is **Propagate**. 

I did all of the models because that is what I understood that @bcoconni wanted even though all that actually is needed for this case is to be able to disable the **Propagate** model

There is a good argument for disabling ground reactions completely but that isn't necessary for my work because keeping the gear retracted achieves the same, albeit with a small performance hit.

@bcoconni asked a good question about _centralizing the feature_ and original reason was to follow the way that exec is already managing allocating all of the models. In hindsight this isn't really adding much and instead I'd change the FGModel constructor to accept an optional `enableName` parameter that when present would call `BindModelEnabled(enableName);`

bcoconni:
> I did all of the models because that is what I understood that @bcoconni wanted even though all that actually is needed for this case is to be able to disable the Propagate model

My point was that if we want to be able to disable any model then IMHO the best option is to add a property to `FGModel`. Now if your PR only targets `FGPropagate` then maybe modifying the `FGModel` class is not the way to proceed and you should only focus your changes on `FGPropagate`?

I mean we may target a change with a reduced scope for a start and then after having cumulated some experience over the usage of that feature, we may be able to extend the concept to all instances of `FGModel`?

> @seanmcleod70 the points you raised in https://github.com/JSBSim-Team/jsbsim/issues/1390 about Auxiliary are the reason why my approach changed because you were right that replacing Auxiliary with a handful of setters only appeared to be working.

I do not follow your point here. @seanmcleod70 made some good points about the extent of the setters that would have to be added but also about the properties that would no longer be updated. How your PR is addressing that particular issue?

Because `FGPropagate` manages a lot of properties:

https://github.com/JSBSim-Team/jsbsim/blob/d371f384f1920677582d04233b326a45c9033542/src/models/FGPropagate.cpp#L829-L921

Zaretto:
See the update on #1482 - basically I don't need this code change any more.

However being able to disable modules is something that may still have some value; certainly being able to freeze (disable) systems it's something that is a standard feature for the systems on the level D simulators as it permits isolation of issues and there are times when I could see that it might be useful.

One specific example is a project I worked on last year where run time performance was critical so there was a JSBSim patch script as part of the CI to to remove ground reactions, external reactions and some other systems (I can't remember the exact list).

Before I close this and delete the branch I'd like to have the discussion about if there is any need for this code.

src/models/FGModel.cpp
```diff
@@ -55,7 +56,7 @@ GLOBAL DECLARATIONS
 CLASS IMPLEMENTATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
-FGModel::FGModel(FGFDMExec* fdmex)
+FGModel::FGModel(FGFDMExec* fdmex, const std::string& enableName)
```
bcoconni:
```suggestion
FGModel::FGModel(FGFDMExec* fdmex, const std::string& name)
 : Name(name)
```
I would take opportunity of passing a name to the constructor to initialize the member `Name`. That would save setting `Name` in each inherited class.

src/models/FGModel.cpp
```diff
@@ -67,6 +68,8 @@ FGModel::FGModel(FGFDMExec* fdmex)
   exe_ctr     = 1;
   rate        = 1;
 
+  if (!enableName.empty()) BindModelEnabled(enableName);
+
```
bcoconni: Not sure why a separate method is needed for this trivial task? The constructor of `FGModel` is a dozen lines long, the content of `BindModelEnabled` could be copied inside the constructor and its code would still be much shorter than any screen height.

src/models/FGModel.cpp
```diff
@@ -87,10 +90,27 @@ bool FGModel::InitModel(void)
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+void FGModel::BindModelEnabled(const std::string& name)
+{
+  const string path = "simulation/models/" + name + "/enabled";
+  const bool existed = PropertyManager->HasNode(path);
+  ModelEnabled = PropertyManager->GetNode(path, true);
+  // Default to enabled; a freshly created node would otherwise read false.
+  if (!existed && ModelEnabled) ModelEnabled->setBoolValue(true);
```
bcoconni:
```suggestion
  const string path = "simulation/models/" + Name + "/enabled";
  Property->Tie(path, &enabled);
```
Tying the property to a member `bool enabled` makes the code much shorter and saves the trouble of all the *ifs* you have used to guard the code from potential errors.

This code should be moved directly inside the constructor.

src/models/FGModel.cpp
```diff
@@ -87,10 +90,27 @@ bool FGModel::InitModel(void)
...
+
 bool FGModel::Run(bool Holding)
 {
   FGModel::Debug(2);
 
+  // A model whose simulation/models/<name>/enabled property is false is skipped
+  // entirely: its Run() body does not execute, but its last state and property
+  // bindings are preserved so an external system can replace it, for example
+  // external propagation or host-owned ground reactions.
+  if (ModelEnabled && !ModelEnabled->getBoolValue()) return true;
```
bcoconni:
```suggestion
  if (!enabled) return true;
```
Using a tied property makes the test much simpler and more legible.

src/models/FGModel.h
```diff
@@ -48,6 +48,8 @@ INCLUDES
 FORWARD DECLARATIONS
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
+class SGPropertyNode;
+
```
bcoconni:
```suggestion
```
You don't need that when using tied properties.

src/models/FGModel.h
```diff
@@ -70,8 +72,12 @@ class JSBSIM_API FGModel : public FGModelFunctions
...
 public:
 
-  /// Constructor
-  explicit FGModel(FGFDMExec*);
+  /** Constructor.
+      @param enableName optional stable canonical name for this model. When
+                        non-empty it binds simulation/models/<enableName>/enabled
+                        (default true); setting that property false skips this
+                        model's Run(). Empty means the model has no enable flag. */
+  explicit FGModel(FGFDMExec*, const std::string& enableName = "");
```
bcoconni:
```suggestion
  /** Constructor.
      @param name optional stable canonical name for this model. When
                  non-empty it binds simulation/models/<enableName>/enabled
                  (default true); setting that property false skips this
                  model's Run(). Empty means the model has no enable flag. */
  FGModel(FGFDMExec*, const std::string& name);
```
* You no longer need the `explicit` qualifier for a constructor with more than one argument.
* I would rather use `name` for the name argument since the point is to use that value to populate the member `Name` in the process.
* I would not provide a default value to `name`, thus forcing all classes inheriting from `FGModel` to pick a name.

src/models/FGModel.h
```diff
@@ -101,8 +107,18 @@ class JSBSIM_API FGModel : public FGModelFunctions
 protected:
   unsigned int exe_ctr;
   unsigned int rate;
+  SGPropertyNode* ModelEnabled = nullptr;
```
bcoconni:
```suggestion
  bool enable = true;
```
Using a tied property allows to define the default value here and saves using a lot of *ifs* to check for the existence of the pointer and property.

src/models/FGModel.h
```diff
@@ -101,8 +107,18 @@ class JSBSIM_API FGModel : public FGModelFunctions
...
   std::string Name;
 
+  /** Bind this model's execution-enable flag to
+      simulation/models/<name>/enabled (default true), where name is the stable
+      canonical name passed to the constructor. When that property is set false,
+      Run() is skipped while the model's last state and property bindings are
+      preserved, the property-tree mechanism by which an external system can
+      replace a model, for example external propagation owning FGPropagate or a
+      host physics engine owning FGGroundReactions. */
+  void BindModelEnabled(const std::string& name);
```
bcoconni: This method is no longer needed but the docs should be kept and moved in the class documentation (lines 57-63)

src/models/FGOutput.cpp
```diff
@@ -55,7 +55,7 @@ namespace JSBSim {
 CLASS IMPLEMENTATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
-FGOutput::FGOutput(FGFDMExec* fdmex) : FGModel(fdmex)
+FGOutput::FGOutput(FGFDMExec* fdmex) : FGModel(fdmex, "output")
 {
   Name = "FGOutput";
```
bcoconni:
```suggestion
FGOutput::FGOutput(FGFDMExec* fdmex) : FGModel(fdmex, "FGOutput")
{
```
Since the `FGModel` constructor populates `Name` itself, you no longer need to initialize it in the `FGOutput` constructor.

bcoconni (changes requested):
I think the code in the constructor of `FGModel` could be simpler using a tied property instead of managing explicitly an `SGPropertyNode` instance with all the *if's* that that drains.

Also I don't see much point in using 2 different names, one for the `enable` property and another one for the `Name` member? I would merge them into a single one. And the model classes would still be able to rename themselves without affecting the name of the `enabled` property.

Finally, I am not a huge fan of Uncle Bob's clean code principle about using a myriad of small trivial functions. So I'd rather not have the new method `BindModelEnabled`, especially since, when using a tied property, its code is 2 lines long.

src/models/FGOutput.cpp
bcoconni: And the same should be applied to the other model classes `FGAcceleration`, `FGAerodynamics`, etc.

bcoconni:
Just for the record, I have just came across this old feature request on Sourceforge [\#86 Turn off models individually](https://sourceforge.net/p/jsbsim/feature-requests/86) from @jonsberndt and dating back to October 10, 2011.

Your PR will allow to close an almost 15-year-old request @Zaretto 😆

Zaretto:
Thanks @bcoconni for the guidance, and it makes it a lot nicer. All applied and pushed ready for your input.

One thing I'd like your guidance on and maybe @seanmcleod70  was my change to use `FGInertial` instead of the current name "Earth" mainly because "Earth" just seemed wrong whereas `simulation/models/FGInertial/enabled` looked right.

agodemar:
@Zaretto @bcoconni @seanmcleod70 this is a great outcome! 
Can someone remind me how this new feature relates to this [Discussion: Review & revise the state propagation procedure (FGPropagate / FGFDMExec) #1391](https://github.com/JSBSim-Team/jsbsim/discussions/1391#discussion-9477388) ?

bcoconni:
@Zaretto I took the liberty to modify your code to fix the failure of the test case `CheckOutputRate`. The logic in your last proposal for `FGModel::Run()` was always bypassing the update of the counter `exe_ctr` when `enabled` was true. Actually the counter should be updated whatever the value of `enabled`, otherwise the rate at which the model should be executed is ignored, which is a problem for the `<input>` and `<output>` sections.

> One thing I'd like your guidance on and maybe @seanmcleod70 was my change to use `FGInertial` instead of the current name "Earth" mainly because "Earth" just seemed wrong whereas `simulation/models/FGInertial/enabled` looked right.

I do not have strong opinions about that. The thing is that IMHO the class `FGInertial` should better be named `FGPlanet` because it manages the planet gravity, geometry (equatorial and polar radii) and rotation rate. But renaming the class would break the backward compatibility of JSBSim C++ API. Although we have already broken the C++ API in the past (hence version 1.2), I do not feel there is such a need to rename `FGinertial` that would justify breaking the API again. That being said, I have changed `Name` to "Earth" in the commit 5fe0882 because the planet can be renamed via the XML definition files and because its default parameters are those of planet Earth. I guess we can have the best of both worlds by keeping the `Name` setting in the constructor like below:

```c++
FGInertial::FGInertial(FGFDMExec* fgex)
  : FGModel(fgex, "FGInertial")
{
  Name = "Earth";
```

That would allow having the property named `simulation/models/FGInertial/enabled` while still displaying the name "Earth" in the debug logging:
https://github.com/JSBSim-Team/jsbsim/blob/0b688c801c52d800f75d5c41e5434ce5d7618e88/src/models/FGInertial.cpp#L297-L302

@agodemar My understanding is that this PR is reduced in scope with respect to the discussions we had in #1390 and #1391. The PR is now only focusing on individually disabling the `FGmodel` instances but it does not address how data should be transmitted from one model to another when a model is disabled in between. And this PR does not address how the properties from a disabled model could be updated, especially the properties which are read-only.

Zaretto:
@bcoconni thank you for fixing the test case and the dual naming works for me.

@agodemar this PR isn't really related to the discussion about propagation we had in #1390, #1391 and #1482 - in effect disabling the integration in the propagation gives me enough to make it work for my current case. 

However I do think that all of the discussions are pointing us towards the bigger question of how we can extend JSBSim at the class level but that is probably a separate discussion.

bcoconni (approved):
I have pushed a commit for the dual naming of the `FGInertial` instances.

Looks good to me now.

Zaretto:
@bcoconni is there anything else you need from me to merge this ?

bcoconni:
> @bcoconni is there anything else you need from me to merge this ?

No. All is good as far as I am concerned.

I was just waiting to see if there were other comments from the JSBSim team but the summer break made the topic vanish from my radar :wink: Thanks for the reminder.

The PR is now merged.

PR #1482, opened 2026-06-23
Add FGAuxiliary setters for host-owned propagation

Zaretto:
Together with the per-model enable property (#1481), these five FGAuxiliary setters are the main JSBSim-core changes that permit running the engine inside an externally hosted simulation such as DCS World.

Adds to FGAuxiliary:
- Setadot, Setbdot (alpha/beta rates)
- SetNx, SetNy, SetNz (body-axis load factors)

When a host owns the state integration it injects the derived quantities Auxiliary would otherwise compute from the propagation it no longer runs. The host-owned domain is propagation together with ground reactions, tightly coupled through the 6DoF integration and externalised as one unit. These rates and load factors are among the few quantities Auxiliary cannot recompute once that integration is external, so the host supplies them.

Follows the state-propagation review @agodemar raised in #1390 and #1391.

### DIFF
--- a/src/models/FGAuxiliary.h
+++ b/src/models/FGAuxiliary.h
@@ -291,6 +291,16 @@ public:
 
   void SetAeroPQR(const FGColumnVector3& tt) { vAeroPQR = tt; }
 
+  // External-injection setters for a host that owns the state integration
+  // (propagation and ground reactions) and provides the derived quantities
+  // Auxiliary would otherwise compute from the propagation it no longer runs:
+  // the alpha/beta rates and the body-axis load factors.
+  void Setadot(double v) { adot = v; }
+  void Setbdot(double v) { bdot = v; }
+  void SetNx(double v) { Nx = v; }
+  void SetNy(double v) { Ny = v; }
+  void SetNz(double v) { Nz = v; }
+
   struct Inputs {
     double Pressure;
     double Density;


agodemar:
Overall, it looks good to me, thanks @Zaretto for this solution.
@bcoconni and @seanmcleod70 any comments on this?

seanmcleod70:
Will take a look on the weekend at this PR and https://github.com/JSBSim-Team/jsbsim/pull/1481 to refresh my memory of the details.

seanmcleod70:
@Zaretto how come there are so few new `FGAuxiliary` functions?

Looking at my comment - https://github.com/JSBSim-Team/jsbsim/issues/1390#issuecomment-3902040361 in the previous discussion I noted 12 new `FGAuxiliary` functions that you had added to your fork at the time but with this PR you've only added 5.

bcoconni:
There again I agree with @seanmcleod70 and in addition to that, this looks to me like a quick and dirty implementation to solve the particular issues that the DCS interface is facing.

The problem is that in most cases these methods will be no-op's because the content passed to the setters will be overwritten by the method `FGAuxiliary::Run`, unless the `FGAuxiliary` model is disabled using the feature submitted by PR #1481. That does not look like a great API design to me.

My point here is not to reject the PR upfront but to tell that:
1. This PR and PR #1481 are strongly interconnected. And they should not be designed and reviewed separately.
2. We need to put much thoughts into these 2 PRs before proceeding. This implies suggesting and reviewing alternative designs to select the most reasonable one based on a discussion about their relative merits.

Also why are the setters not used to allow modifying their related properties ?
https://github.com/JSBSim-Team/jsbsim/blob/f46a071f5ef5e55bdc4a997461163c7d78b36c3a/src/models/FGAuxiliary.cpp#L438-L440
https://github.com/JSBSim-Team/jsbsim/blob/f46a071f5ef5e55bdc4a997461163c7d78b36c3a/src/models/FGAuxiliary.cpp#L454-L455

Zaretto:
@seanmcleod70 With only propagate and ground reactions disabled, Auxiliary still runs, so the only things that
cannot be calculated are the five variables derived from the propagation the host owns. 

@bcoconni I think making these five properties that are already bound read/write is a much better solution than adding methods, and one that will permit this all to be done from the XML.

This is not designed as a Q&D way to integrate DCS as I've already done that, as discussed on #1390. This is a real attempt to bring extra functionality into JSBSim that will have more widespread use and increase the flexibility of what JSBSim can be used for. I've already seen one real world case where JSBSim couldn't be used without patches 

Merging #1481 into this is my preferred approach but I will take guidance.

seanmcleod70:
> With only propagate and ground reactions disabled, Auxiliary still runs, so the only things that
cannot be calculated are the five variables derived from the propagation the host owns.

Yep, I relooked at the code in that context, i.e. `FGAuxiliary::Run()` still being called, and the 5 variables this PR enables to be set externally, but I'm a bit confused given that `FGAuxiliary::Run()` calculates `adot` and `bdot`.

https://github.com/JSBSim-Team/jsbsim/blob/d371f384f1920677582d04233b326a45c9033542/src/models/FGAuxiliary.cpp#L176-L177

I must be missing something?

bcoconni:
Yes I agree with @seanmcleod70, if `FGPropagate` is disabled then the PR needs to manually update the properties from `FGPropagate` that are no longer updated but are still fed to the other models? The code of `FGAuxiliary` needs not be modified? 🤔

> I think making these five properties that are already bound read/write is a much better solution than adding methods, and one that will permit this all to be done from the XML.

Yes, I like this idea with one caveat: the properties are currently read-only (they only have a getter) so you will need to re-wire them on the fly to make them read/write as soon as the model is disabled.

Zaretto:
Since I first had the idea of using JSBSim inside DCS it's been quite a journey, and the last few months have achieved real progress because now I've got acEFM working with zero changes to JSBSim.  Credit for this goes to @seanmcleod70 and @bcoconni for the thorough review of my pull requests that has consistently guided what I've been doing into the right places in the codebase. 

In the end all that was needed was simply to set the integrators in `FGPropagate` to `eNone` and ensuring the gear stays retracted; which makes #1481 not important.

seanmcleod70:
@Zaretto great to hear it's as simple as disabling the integrators in JSBSim.

Just trying to understand the comment about `in.vUVWdot` being stale?

```c++
    // Alpha/beta rates - Auxiliary::Run() zeros adot/bdot then recomputes
    // from in.vUVWdot (which is stale in DCS mode). We compute from
    // frame-to-frame differencing and apply AFTER Run() in update().
    if (init_body) {
        double factor = 1.0 / dT;
        pending_adot = (alpha_rads - last_alpha_rads) * factor;
        pending_bdot = (beta_rads - last_beta_rads) * factor;
    }
....
....
    // DCS-provided Vt for debug logging (Auxiliary will compute its own from vAeroUVW)
    double Vt_ms = Magnitude(vx - wind_vx, vy - wind_vy, vz - wind_vz);
```

Zaretto:
Apologies - I didn't expect you to review that before I'd finished it. This has now gone and a general tidy up of the interface is completed; if you get a moment to have a quick look I'd appreciate your input.

seanmcleod70:
No problem, I was just quite keen to refresh my memory of the integration of JSBSim as an external flight model for DCS, so went and took a quick look, and given this particular PR with regards to `FGAuxiliary` also looked out for this code.

I've taken a new look after your tidy up, in particular glancing through these 2 files.

https://github.com/Zaretto/acEFM/blob/master/flyt-EFM-dcsJSBSim/JSBSim_interface.cpp
https://github.com/Zaretto/acEFM/blob/master/flyt-EFM-dcsJSBSim/DCS_interface.cpp

Given your comment about simply needing to disable the integrators in JSBSim, i.e.

```c++
    // DCS owns the state, so set all four Propagate integrators to eNone which
    // disables integration and permits injection of the state injected in
    // set_current_state_body_axis
    fgSetDouble("/fdm/jsbsim/simulation/integrator/rate/rotational", 0);
    fgSetDouble("/fdm/jsbsim/simulation/integrator/rate/translational", 0);
    fgSetDouble("/fdm/jsbsim/simulation/integrator/position/rotational", 0);
    fgSetDouble("/fdm/jsbsim/simulation/integrator/position/translational", 0);
```

I guess my initial thoughts were that the breakdown between DCS and JSBSim was along the lines in the following diagram, with JSBSim's responsibilities in the grey block, and with DCS taking care of the blocks I've circled in pale blue.

image

However, looking at the code, it isn't the case that DCS simply takes $\dot{X}$ from JSBSim, integrates it and returns $X$.

Rather, you supply DCS with the total force and total moment (excluding gear) from JSBSim and then DCS uses that, with mass information etc. to run the 6-DOF dynamics, and integrates the result that you then feed back into JSBSim. In other words, the blocks I've highlighted in green are also DCS blocks, including the atmosphere.

```c++
void ed_fm_add_local_force(double &x, double &y, double &z, double &pos_x, double &pos_y, double &pos_z)
{
    FGJSBsim *model = get_model();
    x = model->fgGetDouble("/fdm/jsbsim/forces/fbx-total-lbs") * LBS_TO_N;
    y = -model->fgGetDouble("/fdm/jsbsim/forces/fbz-total-lbs") * LBS_TO_N;
    z = model->fgGetDouble("/fdm/jsbsim/forces/fby-total-lbs") * LBS_TO_N;

    pos_x = center_of_mass.x;
    pos_y = center_of_mass.y;
    pos_z = center_of_mass.z;
}
```

```c++
void ed_fm_add_local_moment(double &x, double &y, double &z)
{
    FGJSBsim *model = get_model();

    // jsb
    // x + back
    // y + right
    // z + up

    // dcs
    // x + forward
    // y + up
    // z + right

    x = model->fgGetDouble("/fdm/jsbsim/moments/l-total-lbsft") * LBSFT_TO_NM;
    y = -model->fgGetDouble("/fdm/jsbsim/moments/n-total-lbsft") * LBSFT_TO_NM;
    z = model->fgGetDouble("/fdm/jsbsim/moments/m-total-lbsft") * LBSFT_TO_NM;
}
```

You then feed $X$ back into JSBSim via these and a couple of other functions. Although DCS's alpha and beta are ignored.

```c++
void FGJSBsim::set_current_state_body_axis(
    double ax, double ay, double az,                      // linear acceleration component in body coordinate system
    double vx, double vy, double vz,                      // linear velocity component in body coordinate system
    double wind_vx, double wind_vy, double wind_vz,       // wind linear velocity component in body coordinate system
    double omegadotx, double omegadoty, double omegadotz, // angular accelearation components in body coordinate system
    double omegax, double omegay, double omegaz,          // angular velocity components in body coordinate system
    double yaw, double pitch, double roll,                // radians
    double alpha_rads,                                    // AoA radians
    double beta_rads,                                      //AoS radians
    double dT,
    double ro_kgm3
)
```

```c++
void FGJSBsim::set_altitude(double h_ft)
{
    Propagate->SetAltitudeASL(h_ft);
}
```

Maybe we should move this discussion to a Github JSBSim discussion regarding general questions and answers about using JSBSim as an external flight model for DCS so that it's easier for others to find as opposed to being buried within the comment section of this PR?

agodemar:
> Maybe we should move this discussion to a Github JSBSim discussion regarding general questions and answers about using JSBSim as an external flight model for DCS so that it's easier for others to find as opposed to being buried within the comment section of this PR?

@seanmcleod70 agreed

seanmcleod70:
Okay I've created a new discussion item for this at - https://github.com/JSBSim-Team/jsbsim/discussions/1488

bcoconni:
So I guess this PR can now be closed?

seanmcleod70:
@Zaretto closing this PR, there is a discussion item at - https://github.com/JSBSim-Team/jsbsim/discussions/1488.

Issue #1483, opened 2026-06-28
Physical unit inconsistency for 'dynamic-fmep' in FGPiston


Tarkiya:
### Description

A potential physical unit/dimension inconsistency was detected in the `FGPiston` propulsion model. The XML configuration element `dynamic-fmep` is parsed and converted strictly as a pressure unit (**Pascal / PA**), but its mathematical utilization throughout the simulation behaves like a speed-dependent friction coefficient with the dimension **$Pa \cdot s / m$** (or equivalent).

This mismatch can lead to incorrect unit conversions. If a user specifies alternative pressure units in the configuration file (e.g., `PSI`), the parser will apply a pure pressure conversion factor, yielding physically incorrect results for a coefficient that must be multiplied by speed.

### Code Evidence

1. Parsing Phase
Location: In `src/models/propulsion/FGPiston.cpp` (Line 194), the configuration is parsed assuming it is a pure pressure value:
Permalink: [https://github.com/JSBSim-Team/jsbsim/blob/f46a071f5ef5e55bdc4a997461163c7d78b36c3a/src/models/propulsion/FGPiston.cpp#L194](https://github.com/JSBSim-Team/jsbsim/blob/f46a071f5ef5e55bdc4a997461163c7d78b36c3a/src/models/propulsion/FGPiston.cpp#L194)

2. Mathematical Usage Phase
In the actual physical calculations within the same file, FMEPDynamic is multiplied by piston speed (converted to meters per second via fttom) to calculate the actual Friction Mean Effective Pressure (FMEP).

Evidence (Line 794):
Permalink: [https://github.com/JSBSim-Team/jsbsim/blob/f46a071f5ef5e55bdc4a997461163c7d78b36c3a/src/models/propulsion/FGPiston.cpp#L794](https://github.com/JSBSim-Team/jsbsim/blob/f46a071f5ef5e55bdc4a997461163c7d78b36c3a/src/models/propulsion/FGPiston.cpp#L794)

### Dimensional Analysis

According to the mathematical model:

$$\text{Pressure (Pa)} = \text{Coefficient} \times \text{Speed (m/s)}$$

Therefore, the correct physical dimension for this coefficient must be:

$$\text{Unit(Coefficient)} = \frac{\text{Pa}}{\text{m/s}} = \text{Pa} \cdot \text{s} / \text{m}$$

[...]

PR #1486, opened 2026-07-11
Clarify actual units for dynamic-fmep

seanmcleod70:
Based on the discussion in - https://github.com/JSBSim-Team/jsbsim/issues/1483


bcoconni:
I guess it's the lesser of two evils.

A topic adding up to [Wrong units on {I|B}SFC](https://sourceforge.net/p/jsbsim/bugs/73/)

PR #1487, opened 2026-07-11
Move console escape codes to `FGLogConsole`

bcoconni:
This PR is a continuation of the logging redesign series (PRs #1094 and others).

Since we now use `LogFormat` in messages, the terminal escape codes can now be moved from `FGJSBBase` to `FGLog`:
https://github.com/JSBSim-Team/jsbsim/blob/d371f384f1920677582d04233b326a45c9033542/src/FGJSBBase.cpp#L49-L59

Indeed, these escape codes are now only used by the class `FGLogConsole`. In the process of this migration, I have discovered that:
* The parameter `--nohighlight` passed to the Python script `python/JSBSim.py` is no longer working since the method `FGJSBBase::disableHighLighting` has no effect on `FGLogConsole`. This has been fixed by creating a new Python class `LogConsole` that inherits `DefaultLogger` and displays the required escape codes on the terminal.

* The method `FGJSBBase::disableHighLighting` is now useless and has been removed. A new method `FGLogConsole::DisableHighLighting` has been added to allow disabling the usage of terminal escape codes when `FGLogConsole` is the current logger.
  * As a result, the executable `JSBSim.cpp` has been rewritten to use `FGLogConsole::disableHighLighting` when the parameter `--nohighlight` is passed.

* Some terminal escape codes are still used here and there (in `JSBSim.cpp`, `FGFDMExec.cpp`, `FGFunction.cpp` and `FGPropulsion.cpp`). Their occurrences are replaced by the equivalent `LogFormat`.
  * The terminal escape codes used in the executable `JSBSim.cpp` have been replaced by their equivalent from `LogFormat`. This was needed so that this formatting would react to the call to `FGLogConsole::DisableHghLighting`.
  * The exception `WrongNumberOfArguments` now inherits from `XMLLogException` instead of `BaseException` to allow using `LogFormat` in replacement of terminal escape codes.

  * For the particular case of `FGPropulsion::GetPropulsionTankReport`, the method is returning a `std::string` so `LogFormat` cannot be used. So I have simply removed the formatting from `FGPropulsion::GetPropulsionTankReport` and updated the formatting of `FGMassBalance::GetMassPropertiesReport` to match that removal (i.e. the first column of the report is no longer displayed in bold characters).
    * The mass and inertia of tank contents are now displayed with a one digit precision to match the precision used for the other items in the mass report.

    * While I was going down the rabbit hole of the mass properties report, I took the opportunity to replace `Weight` and `lbf` by `Mass` and `lb`. I have also added a `Tank Content` label in the report so that it is now more obvious that the last lines in the report are displaying the content of the tanks. And the line for the totals have been modified to match the vertical alignment of the columns.

* The colors `GREEN` and `CYAN` are missing in the method `FGLogConsole::Format`. They have been added. 

I also took this opportunity to clean up the header inclusion in `FGJSBBase.h`. Since the header `<queue>` has been removed from `FGJSBBase.h`, the header `<deque>` had to be added to `FGPropagate.h`.

My initial goal was just to move the terminal escape codes to `FGLogConsole` and while pulling that thread, a lof of other things unraveled 😄

### DIFF
--- a/python/JSBSim.py
+++ b/python/JSBSim.py
@@ -29,6 +29,31 @@ import xml.etree.ElementTree as et
 
 import jsbsim
 
+
+class LogConsole(jsbsim.DefaultLogger):
+    def format(self, log_format:jsbsim.LogFormat):
+        if log_format == jsbsim.LogFormat.RED:
+            print("\x1b[31m\0", end="")
+        elif log_format == jsbsim.LogFormat.BLUE:
+            print("\x1b[34m\0", end="")
+        elif log_format == jsbsim.LogFormat.CYAN:
+            print("\x1b[36m\0", end="")
+        elif log_format == jsbsim.LogFormat.GREEN:
+            print("\x1b[32m\0", end="")
+        elif log_format == jsbsim.LogFormat.DEFAULT:
+            print("\x1b[39m\0", end="")
+        elif log_format == jsbsim.LogFormat.BOLD:
+            print("\x1b[1m\0", end="")
+        elif log_format == jsbsim.LogFormat.NORMAL:
+            print("\x1b[22m\0", end="")
+        elif log_format == jsbsim.LogFormat.UNDERLINE_ON:
+            print("\x1b[4m\0", end="")
+        elif log_format == jsbsim.LogFormat.UNDERLINE_OFF:
+            print("\x1b[24m\0", end="")
+        else:
+            print("\x1b[0m\0", end="")  # Reset
+
+
 def main():
     parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter)
     parser.add_argument("input", nargs='?', help="script file name")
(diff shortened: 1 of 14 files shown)


bcoconni:
> While I was going down the rabbit hole of the mass properties report, I took the opportunity to replace `Weight` and `lbf` by `Mass` and `lb`.

This proposal should be reviewed against the discussion #368 we had several years ago about that very topic. My stance here is that users are actually providing masses not weights, as per the possible conversion from `KG` implies. If opinions against my proposal are too strong I can remove that change.

seanmcleod70:
Looks good. 

Did have to refresh my memory in terms of the non-SI units `slugs`, `lbm` and `lbf` :wink:

It does look like you missed an instance in `FGMassBalance.cpp`: 

https://github.com/JSBSim-Team/jsbsim/blob/285636cc75fa2bdf93820af418b683208d430898/src/models/FGMassBalance.cpp#L512

bcoconni:
> It does look like you missed an instance in `FGMassBalance.cpp`:

This is now fixed.

I have also added a comment in `JSBSim.cpp` about the usage of `FGLogging` rather than `cout` so that the formatting is responsive to the setting of the parameter `--nohighlight`.

Issue #1372, opened 2026-01-03
FGTurbine EGT


Octal450:
**I'm submitting a ...**
  - [x] bug report

**Describe the issue**
EGT is based on throttle. That is very wrong. It should be changed to N2 norm to be a bit more closer to proper.

**What is the motivation / use case for changing the behavior?**
While not exactly right, it at least makes the movement of the EGT gauges of aircraft that don't yet override the EGT with a correct calculation more realistic.

**Fix**
https://github.com/JSBSim-Team/jsbsim/blob/9ea4925a68cdf1ee5a4c0c33e591c744f624cc03/src/models/propulsion/FGTurbine.cpp#L221

Change to
`EGT_degC = in.TAT_c + 363.1 + N2norm * 357.1;`

seanmcleod70:
Does look like a very simple/basic model for EGT, and I would assume too responsive when based on `ThrottlePos`, and using N1norm or N2norm would be better.

> don't yet override the EGT with a correct calculation 

What does a **correct** calculation look like?

Came across this 4K video which shows N1, EGT. Fairly readable if you screen clip and sharpen a bit. TAT also somewhat readable. N2 is fairly difficult to read. 

https://www.youtube.com/watch?v=ZelrHEDOU_M

Example:

<img width="528" height="476" alt="Image" src="https://github.com/user-attachments/assets/6acefabc-2b08-4f5d-a1c0-52a33bbd8b80" />

seanmcleod70:
> No reason to fix the numbers in JSBSim.

Not planning to. But when someone mentions a **correct calculation** I'm interested in seeing the calculation and the data for it. Did you for example come across some table of N1/N2/TAT/EGT for some specific engine(s)? If so, can you share it?

Also to use it as a rough comparison to see how different it is to the current JSBSim model, and to a modified JSBSim model using N1 or N2 norms.

Similiar to how I came across a number of different ground effect models that had been published and compared them.

Octal450:
I'd like to ask for a status update - should be a oneliner fix here. Would be nice to see it fixed even if it doesn't affect me personally.

seanmcleod70:
@Octal450 sorry for the delay. Did some quick plotting to get a feel for the differences.

```python
import numpy as np
import matplotlib.pyplot as plt

x = np.linspace(0, 1, 100)

plt.figure()

egt_throttle = x * 357.1 + 363.1
egt_n2norm = x*x * 357.1 + 363.1

[...]

PR #1490, opened 2026-07-16
Tie turbine EGT to N2norm instead of throttle position

seanmcleod70:
As disccused in issue - https://github.com/JSBSim-Team/jsbsim/issues/1372


bcoconni (approved):
Looks good to me.

PR #1491, opened 2026-07-17
Tie the external reaction and buoyancy properties only once

Zaretto:
FGExternalReactions::Load() and FGBuoyantForces::Load() are public and append to their force and cell lists, so either may be called more than once to add further items.

My specific use case for this that an aircraft declares in <external_reactions> or <buoyant_forces> but then I need to add external forces from the C++ side which causes harmless errors because a second bind would log errors on the tied properties.

### DIFF
--- a/src/models/FGBuoyantForces.cpp
+++ b/src/models/FGBuoyantForces.cpp
@@ -130,8 +130,12 @@ bool FGBuoyantForces::Load(Element *document)
 
   PostLoad(document, FDMExec);
 
-  if (!NoneDefined) {
+  // Load() is public and appends to the cell list, so it may be called more than
+  // once to add further gas cells. The properties must only be tied the first
+  // time: tying them again fails and logs an error for each one.
+  if (!NoneDefined && !isBound) {
     bind();
+    isBound = true;
   }
 
   return true;
--- a/src/models/FGBuoyantForces.cpp
+++ b/src/models/FGBuoyantForces.cpp
@@ -130,12 +130,8 @@ bool FGBuoyantForces::Load(Element *document)
 
   PostLoad(document, FDMExec);
 
-  // Load() is public and appends to the cell list, so it may be called more than
-  // once to add further gas cells. The properties must only be tied the first
-  // time: tying them again fails and logs an error for each one.
-  if (!NoneDefined && !isBound) {
+  if (!NoneDefined) {
     bind();
-    isBound = true;
   }
 
   return true;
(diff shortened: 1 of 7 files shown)


seanmcleod70:
Hmm, although the `Load(Element *document)` methods are public, I'm not sure that in general the model implementations are designed to have their `Load()` method called multiple times. I'd imagine most of them assume they'll be called once while reading in the relevant config files.

So in terms of adding additional external forces etc. later on via your C++ host are you building up XML elements in your C++ host and then passing that in to the `Load()` methods?

So I wonder if it wouldn't be better to rather add additional methods that allow C++ hosts to add additional external forces/moments etc.?

Zaretto:
That's a good catch Sean, and you're right I was approaching this as a model developer rather than reaching for my C++ hat. After some thought I've now pushed what I believe is a more correct implementation of what I've been trying to achieve and just to be clear this is not a final solution I'm hoping for some discussion and validation of what I've done.

Briefly, there is a new `Add()` method that adds a force or a moment, used both by the existing code in `Load()` and by my external code. An API client can add a force to a model that has no `<external_reactions>` element, so I moved the tie into the constructor, which also removes the need for the guard in both `FGExternalReactions` and `FGBuoyantForces`. I've added a few unit tests.

The other thing I've done is make `Add()` take a `std::unique_ptr`, so the caller has to `std::move` the object in which makes it explicit that the model now owns the object and will delete it.

seanmcleod70:
@Zaretto updated version looks good to me. @bcoconni any thoughts?

bcoconni (approved):
Looks **very** good to me:
* `std::unique_ptr<>` are used instead of `new` and `delete`.
* A unit test is added.
* Transfer of ownership is documented in the `Add()` method.

A few things that came to my mind while reviewing this PR:
* I am not comfortable with the code `Add(std::move(force)`. The variable `force` still exists after the call to `Add()` but no longer owns the pointer. This leaves the opportunity to dereference a null pointer if one would use `force` after `Add()` was called. However, the variable `force` only exists in a limited scope (it is destroyed as soon as the `while()` loop is completed) and the other options I could imagine were ugly and not that much safer, if any safer.
* External forces added with the `Add()` method need to bind themselves to the property manager. When I was reading the unit test, I was wondering under what name the `HostForce` instance would appear in the property tree until I realized that this is normally done by the `FGExternalForce::bind()` method which is not called by `HostForce`. I am not sure this is the behavior that a user would expect but calling `FGExternalReactions::Add()` will most likely only be done by advanced users, who generally read and try to understand the code :wink:

Nothing worth pushing back the merge of the PR.

bcoconni:
I took the liberty to use a commit message that is different from the PR title which, I think, no longer reflect the intent of the change.

Issue #1489, opened 2026-07-16
Friction solver may fail to converge within 50 iterations at low speeds under high pitch/yaw moments


jimishol:
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

**Describe the issue**
At zero or ultra-low ground velocity ($v < 1.5\text{ fps}$), the projected Gauss-Seidel friction solver in `FGAccelerations::CalculateFrictionForces()` may occasionally fail to converge within its hard-coded 50-iteration limit. This behavior is highly reproducible on lightweight aircraft (such as the Cessna 172P) when parked in moderate-to-strong tailwinds ($15\text{ to } 22\text{ knots}$) or crosswinds, leading to sudden physical collapses on the ramp.

**What is the current behavior?**
When parked in a strong tailwind, aerodynamic pitching moments shift weight off the nose gear. The resulting micro-oscillations in the gear's mass-spring-damper system (`FGLGear::GetBodyForces()`) cause the normal force ($F_n$) to fluctuate rapidly frame-by-frame. 

Because the static friction bounds (`Min` and `Max`) are calculated dynamically from $F_n$, the solver's constraint limits are moving targets. The loop in `src/models/FGAccelerations.cpp` (lines 292-309) struggles to reach convergence (`norm < 1E-5`) under these shifting limits within the 50-iteration wall. 

When it fails to converge, the remaining constraint forces slip or drop, causing the aircraft's tail to drop to the pavement or the entire fuselage to slide sideways on the spot.

**What is the expected behavior?**
At low speeds, the friction constraints should remain stable despite dynamic fluctuations in the normal force, allowing the aircraft to remain parked or taxied safely without the static constraints collapsing.

**What is the motivation / use case for changing the behavior?**
While I have implemented temporary, parked-only helper moments at the aircraft FDM level as a workaround, I suspect this is a core solver limitation that likely affects many lightweight aircraft across the JSBSim ecosystem during low-speed ground operations in windy conditions. 

**Please tell us about your environment:**
 - OS: All
 - JSBSim version: Core repository (current master)
 - Language if applicable: C++

**Other information**
### Context and Links
I encountered this issue while working on the FlightGear Cessna 172P:
* **C172P Issue Ticket:** https://github.com/c172p-team/c172p/issues/1595
* **C172P Workaround Pull Request:** https://github.com/c172p-team/c172p/pull/1596

[...]

Issue #89, opened 2018-07-05
Pitch oscillations at rest, Brakes "ON", with pure crosswind


dany93:
I've discovered a weird behavior, for which unfortunately I (currently) have no solution to propose. I'm afraid it will not be easy.

Conditions:
- Aircraft at rest on the runway, **brakes (or parking brake) "ON"**,
- **Pure crosswind** (bearing close to + / - 90 deg from the nose, from left or right). 3kt may be enough, 10 - 15 kt worsens the effect. Possibly enough to lift the nose gear up.

Observations: 
`Cmalpha` oscillates, instead of being positive steady for a front wind, or negative but steady for a rear wind. The values of cmalpha do not seem significant, nothing happens if they are steady.
EDIT:
(C172P)
Oscillations also appear if I cancel` Cmalpha` (`<value>0 `in the FDM).  Maybe in a narrower angle range, +/-1 deg around crosswind.
I observed that `Cmadot` took high values (+4000, -1000). Then, I canceled` Cmadot` too, and the oscillations were still here. Even with only 3 kt crosswind.
Of course, I checked that `Cmalpha `and `Cmadot` values were 0 as expected in the `jsbsim/aero/coefficient` Properties for the last test.
/EDIT

Aircraft tested: (engine stopped)
C172P,
SenecaII PA34,
DR400-JSBSim,
Boeing 707 Lake of Constance.

Even with engine stopped, for several aircraft with propeller or a jet aircraft ==> the p-factor is not the cause.

Dany

dany93:
Yes, depending on the wind velocity, it appears to "jump" on the front gear. Or the front gear remains lifted above the ground for a few seconds prior to fall down again. The period may depend on the gear model (stiffness, damping) but, if the front gear remains lifted above the ground for several seconds you can say that an efficient moment (presumably due to wind) is applied. It cannot be solved only by changing the gear model (and every tested gear model cannot be bad).

dany93:
I did an ouput `<moments`> logging.
C172P at rest, brakes ON, engine stopped.
12 kt crosswind from left.
The nose oscillates, the nose wheel lifts up above the ground.

I don't know what it means, but the M moments are very high (which is not surprising, but the hardest is probably to know what it comes from).

L_{Aero} (ft-lbs) | M_{Aero} (ft-lbs)
-- | --
500 to 1600|up to +/- 15000 to 70000


[...]

PR #1493, opened 2026-07-26
Fix alpha dot singularity for models without swept wings.

bcoconni:
While investigating the issue #1489, I have again come across the issue with alpha dot where aerodynamics are diverging when the aircraft is submitted to a pure cross wind (see issue #89). The fix is discussed in https://github.com/JSBSim-Team/jsbsim/issues/89#issuecomment-405082664 and consists in using `qbarUW-psf` instead of `qbar-psf`:
```diff
@@ -1106,7 +1106,8 @@
             <function name="aero/coefficient/Cmadot">
                 <description>Pitch moment due to alpha rate</description>
                 <product>
-                    <property>aero/qbar-area</property>
+                    <property>aero/qbarUW-psf</property>
+                    <property>metrics/Sw-sqft</property>
                     <property>metrics/cbarw-ft</property>
                     <property>aero/ci2vel</property>
                     <property>aero/alphadot-rad_sec</property>
```
It has been shown in #89 that this method is reliable for aircraft without swept wings so this PR implements the fix only for these aircraft.

Whether the same fix can be applied to aircraft with swept wings remains to be investigated but it is not the point of this PR.

**Note:** I have quickly googled each model that uses the property `aero/alphadot-rad_sec` to check if they have swept wings or not. So there is some subjectivity in the list of modified models. If some of them actually have swept wings then normally the sweep angle should be quite small.


seanmcleod70:
I remember issue #89 from 7-8 years ago, and actually I see one of my comments - https://github.com/JSBSim-Team/jsbsim/issues/89#issuecomment-405350923 captures some of what I was thinking about in terms of this PR.

So, the solution then, and now, i.e. to switch to using `qbarUW-psf` instead of `qbar-psf` fixes the reported issue in terms oscillations on the ground with close to pure cross-winds etc. But I'm wondering if it's the best approach.

In my mind the main issue is that we have a single constant coefficient, $c_{m_{\dot{\alpha}}}$ which is then being used far outside it's valid domain. For example it would be similar to having a single value for $C_{L_\alpha}$ which is only valid for the linear portion of the $C_L$ versus $\alpha$ curve but use it across the full $\alpha$ range from $-\pi$ to $\pi$.

Whereas typically we would have a table based lookup for $C_L$ vs $\alpha$ and which would be clipped by the range of $\alpha$ values for which we have real data, i.e. no extrapolation outside the specified table's domain.
 
So my thinking is that the other approach, although not as simple as a 2 character/line change, is to replace it with a table that constrains the domain.

From - "**Flight Stability and Automatic Control, Robert Nelson**"

image

Some constraints would be limiting it to an $\alpha$ range. For example in equation 3.102 which shows a typical way to calculate/estimate $c_{m_{\dot{\alpha}}}$ you can see that the linear portion of the horizontal tail's $C_L$ versus $\alpha$ curve is used. Once the main wing and/or the horizontal tail get close to or exceed their critical angle of attack a single constant coefficient for $c_{m_{\dot{\alpha}}}$ isn't going to be valid.

Plus it's based on the lag in a change in the downwash from the wing reaching the tail. Now think of a 10kt tailwind on the ground, i.e. $\alpha$ = +- $\pi$ with some amount of turbulence to generate $\dot{\alpha}$, and with the airflow now flowing from the tail to the wing. Is that going to generate a pitching moment anywhere close to what the current `Cmadot` functions currently generate for this case? 

Example of $C_{m_q}$ in the F-16 model which is based on and constrained to a range of $\alpha$ values.

https://github.com/JSBSim-Team/jsbsim/blob/9a0b0282a3414409123f783e8b766ff002bb5fc9/aircraft/f16/f16.xml#L1688-L1714

Plus some scale reduction due to $\beta$, somewhat along the lines of what you noticed in terms $C_L$ varying with $\beta$.

In equation 3.102 above the horizontal tail's $C_{L_{\alpha}}$ slope is used, but given a single value, i.e. not being a function of $\beta$ I'm assuming this value is most accurate around $\beta = 0$. So given larger values of $\beta$ the $C_{L_{\alpha}}$ slope would be lowered, therefore reducing the pitch damping.

Plus there would be a second effect of $\beta$ on the tail's pitch damping moment, given $\frac{d \epsilon}{d \alpha}$ in equation 3.102 this will change based on the wing's effect on it, based on the wing's $C_{L_{\alpha}}$ slope changing based on $\beta$.

image

But maybe too overcomplicated for preventing aircraft from bobbing about on the ground with a bit of wind 😉

bcoconni:
> Some constraints would be limiting it to an $\alpha$ range. For example in equation 3.102 which shows a typical way to calculate/estimate $c_{m_{\dot{\alpha}}}$ you can see that the linear portion of the horizontal tail's $C_L$ versus $\alpha$ curve is used. Once the main wing and/or the horizontal tail get close to or exceed their critical angle of attack a single constant coefficient for $c_{m_{\dot{\alpha}}}$ isn't going to be valid.

I was not sure using $\alpha$ as a limiting factor would be sufficient. Below are some curves obtained running [c172_cross_wind.xml](https://github.com/JSBSim-Team/jsbsim/blob/9a0b0282a3414409123f783e8b766ff002bb5fc9/scripts/c172_cross_wind.xml)

image

You can see that $\dot{\alpha}$ can reach high values (i.e. 16.2 rad/s ~ 930 deg/s) for low $\alpha$ values (~0.7 deg).

However, following your suggestion, I have modified the aircraft [c172x](https://github.com/JSBSim-Team/jsbsim/blob/9a0b0282a3414409123f783e8b766ff002bb5fc9/aircraft/c172x/c172x.xml) to cut off the $\dot{\alpha}$ coefficients when $|\alpha|>30\text{ deg}$:
```diff
diff --git a/aircraft/c172x/c172x.xml b/aircraft/c172x/c172x.xml
index 8f121c16e..b3f1ed659 100644
--- a/aircraft/c172x/c172x.xml
+++ b/aircraft/c172x/c172x.xml
@@ -653,7 +653,15 @@
                     <property>aero/qbar-area</property>
                     <property>aero/alphadot-rad_sec</property>
                     <property>aero/ci2vel</property>
-                    <value>1.7</value>
+                    <table>
+                        <independentVar>aero/alpha-deg</independentVar>
+                        <tableData>
+                            -30.0 0.0
+                            -20.0 1.7
+                             20.0 1.7
+                             30.0 0.0
+                        </tableData>
+                    </table>
                 </product>
             </function>
 
@@ -1125,7 +1133,15 @@
                     <property>metrics/cbarw-ft</property>
                     <property>aero/ci2vel</property>
                     <property>aero/alphadot-rad_sec</property>
-                    <value>-5.2</value>
+                    <table>
+                        <independentVar>aero/alpha-deg</independentVar>
+                        <tableData>
+                            -30.0 0.0
+                            -20.0 -5.2
+                             20.0 -5.2
+                             30.0 0.0
+                        </tableData>
+                    </table>
                 </product>
             </function>
```
and it works, all the oscillations have been removed and the aircraft gently reaches a stable attitude:
image
Note that the cut off values I have used (20 and 30 degrees) are totally arbitrary.

seanmcleod70:
Hmm, that worked better than I expected. I thought we might need to add additional clipping/scaling e.g. based on $\beta$ and possibly even limiting $\dot \alpha$.

seanmcleod70:
So I was wondering what the result would be if instead of having the moment `Cmadot` going to 0 when it exceeds the known domain we clip it instead, similar to what typically happens with table lookups for lift vs $\alpha$, i.e. the lift doesn't taper to 0, instead it gets clipped to the value at beginning/end of the table.

So for `Cmadot` and `CLadot` I clipped the moment and force to some maximum $\dot \alpha$ magnitude. For the test I chose 5 rad/s ~= 286 deg/s.

```xml
<function name="aero/coefficient/Cmadot">
    <description>Pitch moment due to alpha rate</description>
    <product>
        <property>aero/qbar-area</property>
        <property>metrics/cbarw-ft</property>
        <property>aero/ci2vel</property>
        <value>-5.2</value>
		<max>
			<value>-5</value>
			<min>
				<property>aero/alphadot-rad_sec</property>
				<value>5</value>
			</min>
		</max>
    </product>
</function>
```

With the following results:

image

PR #1499, opened 2026-09-07
Add https:// to the Doxygen and Graphviz links

goelakash:
In `doc/DevelopersDocs.md` the Doxygen and Graphviz links are written as `(www.doxygen.org)` and `(www.graphviz.org)`. Without a scheme GitHub treats them as relative paths, so both 404.

Checked that both resolve with `https://`.


seanmcleod70:
Thanks for picking that up.

PR #1500, opened 2026-09-07
Fix typos in the documentation

goelakash:
Fixes 5 spelling mistakes in the docs, found with `codespell`.

Prose only. No code identifiers, module paths, package names or proper nouns changed, and anything that was a valid word rather than a misspelling was left alone.


seanmcleod70:
Thanks, looks good.

PR #1502, opened 2026-09-11
Add an optional wheel rotational degree of freedom to BOGEY contacts

Felipegalind0:
### Summary

Adds an opt-in wheel rotational degree of freedom for `BOGEY` contacts.
Declare positive radius and spin inertia to enable it:

```xml
<contact type="BOGEY" name="LEFT_MAIN">
  <!-- existing contact elements -->
  <wheel_radius unit="M"> 0.22 </wheel_radius>
  <wheel_inertia unit="KG*M2"> 0.124 </wheel_inertia>
</contact>
```

The existing `wheel-speed-fps` is a kinematic proxy. The new state lets the
friction solve account for touchdown spin-up and apply braking through the
wheel. For stationary terrain:

- The tread row drives axle ground speed minus radius times internal spin
  toward zero, bounded by static friction times normal load and the ground's
  static friction factor.
- The tread force reaches the airframe through the axle, while its wheel
  torque accelerates the spin state. Applying the full contact-patch moment
  to the airframe as well would count the spin-up moment twice.
- A second row applies equal and opposite wheel/airframe brake torque. Its
  target is zero **relative** wheel spin; its bound is the existing braking
  force coefficient times normal load times radius.

With the DOF enabled, `rolling_friction` therefore describes wheel-to-axle
resisting torque. Without it, the legacy interpretation is tire rolling
resistance/hysteresis. This distinction is documented in `FGLGear.h`.

Ground initialization and trim initialize a contacting wheel without tread
slip. Airborne, an approximation reduces the magnitude of wheel spin relative
to the airframe without reversing it: tread deceleration is
`13 + 100*brake` ft/s². The 13 value comes from the legacy wheel-speed decay;
the additional 100 is an authored, uncalibrated brake approximation, not a
normal-load-derived or measured torque. It has no explicit airborne reaction
torque on the airframe. Damping acts only on positive-time steps outside
trim. Its finite decrement does not enforce a rigid brake lock under
arbitrary body motion.

The opt-in properties are:

| Property | Meaning |
| --- | --- |
| `gear/unit[i]/wheel-spin-rad_sec` | Wheel angular rate relative to the airframe, positive for forward rolling. Reads subtract the body rate projected on the spin axis; writes apply the inverse conversion. |
| `gear/unit[i]/wheel-tread-slip-fps` | Axle ground speed along the roll direction minus radius times the internal wheel rate; zero when airborne. |

The internal spin state stays in the ground reference used by the existing
solve, with the same rate reference as `in.PQR`. It is not an inertial or
“absolute” angular velocity. `wheel-speed-fps` retains its existing meaning.
In contact, the spin axis is ground normal cross projected rolling direction;
airborne it is gear up cross steered forward. The scalar state is not remapped
when those axes differ, so banked or steered contact transitions can produce
a jump in the reported relative rate. `FGLGear.h` documents this limitation.

### Why

This comes from [OSFS](https://0sfs.github.io/), an open-source browser flight
simulator using JSBSim through WebAssembly. Touchdown visuals, tire sound and
haptics need wheel spin, and spin-up/braking should also act on the aircraft.
An application-side state cannot participate in JSBSim's friction solve and
would risk counting existing rolling/brake forces again. The wheel degree
of freedom belongs in that solve.

### Implementation

`LagrangeMultiplier` supplies force, moment and optional scalar wheel
Jacobians. The common solver assembles the body terms for every row, adds a
wheel term only when rows share a wheel, and accumulates its acceleration
alongside the body force/moment. `FGLGear` owns the state, configures the
tread and brake rows, and advances spin using the previous solve's
acceleration in step with the existing integration sequence.

The analysis linked from `ConfigureWheelSpinRows` explains the two
constraints: [wheel dynamics and braking derivation](https://github.com/JSBSim-Team/jsbsim/pull/1502#issuecomment-5654721219).

Without both positive wheel elements there is no additional wheel state or
property. The friction formulation for those aircraft is mathematically
equivalent, but the unified moment-Jacobian arithmetic changes floating-point
evaluation order. **Existing aircraft are not guaranteed byte-identical
trajectories.**

### Validation

`TestWheelSpin.py` covers opt-in/invalid dimensions, touchdown spin-up and
momentum transfer, ground initialization, rolling/braking comparison and
airborne decay. This round adds body-relative property/frame regressions and
a native `FGAccelerationsTest1` for analytical tread/brake constraints,
shared/distinct wheel couplings, limits and zero dt.

Verified candidate `499c38326bb6f39b6360bd82d74c5c00e64faeaf` on macOS 27
arm64, Apple clang 21.0.0, CMake 4.4.3 and Python 3.14.7, Release builds:

- Python enabled: **79/79 CTest targets pass**, including all nine Python
  wheel methods and all eight native constraint methods. Python disabled:
  **21/21 CxxTests pass**, including `FGAccelerationsTest1` in the selection
  used by the coverage workflow. No new hosted coverage percentage is claimed.
- The four new Python frame regressions fail on the pre-frame-change engine.
  A control with the new getter and old damping also fails the locked-wheel
  and near-zero drag checks. Omitting or reversing the native solver's
  shared-wheel off-diagonal term fails the coupled-row case; restoring it
  passes. Existing tolerances were not relaxed.
- Fresh four-binary comparison: all 61 shipped XML inputs, isolated working
  directories, natural data outputs plus per-step traces. Each binary has
  57 successful simulations, the same simplex steady-turn trim failure
  (exit 1), and three output/plot inputs (exit 255). Each emits 102 data files
  and 11,877,246 trace rows, including 13 rows from the trim failure.
- Base `d0d8bc6e` and original PR `24e085bf` data are byte-identical. The
  final candidate matches the isolated common-Jacobian cleanup probe exactly.
  Base/final differ in **60 of 102 files across 31 of 57 successful runs**,
  with no exit, shape or non-finite mismatch. Rebuilt original-head outputs
  and reused base/probe outputs also match their retained earlier records.
  Console banners and timings are excluded from data identity.

Representative maximum absolute base/final differences in the per-step trace:

| Run | u (ft/s) | Pitch (rad) | Altitude (ft) | Gear Fx (lbf) |
| --- | ---: | ---: | ---: | ---: |
| `c172_cross_wind` | 0.054043 | 3.6853e−4 | 0.0023394 | 197.33 |
| `c1723` | 0.0025166 | 4.4618e−5 | 0.0071976 | 3.7558e−6 |
| `ZLT-NT-moored-1` | 9.7489e−5 | 1.5745e−7 | 1.4622e−5 | 67.311 |
| `f16_test` | 71.644 | 0.23604 | 17.819 | 179038 |

Crosswind gear pitching moment differs by up to 1,358.7 lbf·ft; moored-ZLT
gear lateral force by 443.29 lbf and yaw moment by 20,929 lbf·ft. These are
not uniformly small force/moment changes. F-16 Δu stays below 1e−6 ft/s until
127.908 s, when gear Fx is already about −2.11 million lbf; both trajectories
subsequently go below ground. Matching that impact sequence is not a
physical-accuracy criterion, and its divergence is not a general tolerance.

The saved-init XML in `c1723` also changes (nine lines of attitude, local
velocity and body rates). Its exact hashes and text diff were checked; the
CSV harness reports only a text mismatch for that file. Raw 2π angle
differences in crosswind/ZLT heading and F-16 roll
are wrap artifacts; wrapped maxima are 6.258e−6, 2.730e−6 and 0.030706 rad.
All comparisons are specific to this compiler, architecture and input set.

The test suites above ran on this exact candidate and were reused
for the final comparison after source and binary hash checks; they were not
rerun as part of that comparison. `TestInputSocket` historically has a
telnetlib3 timeout on this machine, and `TestActuator` once hit a subprocess
timeout under competing load. Both passed first try in the candidate suite.
No revised WebAssembly verification is claimed.

The C172 wheel dimensions/inertias in the existing fixture are estimates,
not measured assemblies or aircraft calibration. The current scalar axle
model and airborne approximation do not establish full wheel gyroscopic,
camber or airborne reaction-torque dynamics.

### DIFF
--- a/src/math/LagrangeMultiplier.h
+++ b/src/math/LagrangeMultiplier.h
@@ -48,12 +48,29 @@ CLASS DOCUMENTATION
 CLASS DECLARATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
+/** Rotational state of a wheel whose spin is resolved together with the
+    ground friction multipliers (see the <wheel_inertia> element of FGLGear). */
+struct WheelSpinDOF {
+  double InvInertia = 0.0; ///< 1 / spin inertia about the axle [1/(slug*ft^2)]
+  double Rate = 0.0;       ///< absolute spin rate about the axle [rad/s]
+  double Accel = 0.0;      ///< spin acceleration from the last friction solve [rad/s^2]
+};
+
 struct LagrangeMultiplier {
   FGColumnVector3 ForceJacobian;
   FGColumnVector3 LeverArm;
   double Min;
   double Max;
   double value;
+  /// When true, the moment applied to the airframe per unit multiplier is
+  /// MomentJacobian instead of LeverArm * ForceJacobian. Needed for torques
+  /// that act between a wheel and the airframe.
+  bool UseMomentJacobian = false;
+  FGColumnVector3 MomentJacobian;
+  /// Wheel spin degree of freedom coupled to this multiplier (or nullptr) and
+  /// the multiplier's Jacobian entry on that wheel's spin rate.
+  WheelSpinDOF* Wheel = nullptr;
+  double WheelCoeff = 0.0;
 };
 
 } // namespace
--- a/src/models/FGAccelerations.cpp
+++ b/src/models/FGAccelerations.cpp
@@ -241,11 +241,42 @@ void FGAccelerations::CalculateFrictionForces(double dt)
   // If no gears are in contact with the ground then return
   if (!n) return;
 
+  // Multipliers coupled to a wheel spin degree of freedom (or carrying an
+  // explicit moment) use generalized Jacobians [linear, angular, wheel]. When
+  // none are present, the original formulation below is used unchanged.
+  bool generalized = false;
+  for (auto m : multipliers) {
+    if (m->UseMomentJacobian || m->Wheel) {
+      generalized = true;
+      if (m->Wheel) m->Wheel->Accel = 0.0; // Refilled by this solve
+    }
+  }
+  auto moment = [](const LagrangeMultiplier* m) {
+    return m->UseMomentJacobian ? m->MomentJacobian : m->LeverArm * m->ForceJacobian;
+  };
+
   vector<double> a(n*n); // Will contain Jac*M^-1*Jac^T
   vector<double> rhs(n);
 
   // Assemble the linear system of equations
   for (unsigned int i=0; i < n; i++) {
+    if (generalized) {
+      const LagrangeMultiplier* mi = multipliers[i];
+      FGColumnVector3 v1 = mi->ForceJacobian / in.Mass;
+      FGColumnVector3 v2 = in.Jinv * moment(mi);
+
+      for (unsigned int j=0; j < i; j++)
+        a[i*n+j] = a[j*n+i];
+
+      for (unsigned int j=i; j < n; j++) {
+        const LagrangeMultiplier* mj = multipliers[j];
+        a[i*n+j] = DotProduct(mj->ForceJacobian, v1) + DotProduct(moment(mj), v2);
+        if (mi->Wheel && mi->Wheel == mj->Wheel)
+          a[i*n+j] += mi->WheelCoeff * mj->WheelCoeff * mi->Wheel->InvInertia;
+      }
+      continue;
+    }
+
     FGColumnVector3 U = multipliers[i]->ForceJacobian;
     FGColumnVector3 r = multipliers[i]->LeverArm;
     FGColumnVector3 v1 = U / in.Mass;
@@ -282,7 +313,15 @@ void FGAccelerations::CalculateFrictionForces(double dt)
     FGColumnVector3 U = multipliers[i]->ForceJacobian;
     FGColumnVector3 r = multipliers[i]->LeverArm;
 
-    rhs[i] = -DotProduct(U, vdot + wdot*r)/d;
+    if (generalized) {
+      const LagrangeMultiplier* mi = multipliers[i];
+      // The wheel has no other applied torque here; drive its part of the
+      // constraint velocity to zero over the step like the airframe's.
+      double wheelTerm = (mi->Wheel && dt > 0.) ? mi->WheelCoeff * mi->Wheel->Rate / dt : 0.;
+      rhs[i] = -(DotProduct(U, vdot) + DotProduct(moment(mi), wdot) + wheelTerm)/d;
+    }
+    else
+      rhs[i] = -DotProduct(U, vdot + wdot*r)/d;
 
     for (unsigned int j=0; j < n; j++)
       a[i*n+j] /= d;
@@ -317,7 +356,14 @@ void FGAccelerations::CalculateFrictionForces(double dt)
 
     FGColumnVector3 F = lambda * U;
     vFrictionForces += F;
-    vFrictionMoments += r * F;
+    if (generalized) {
+      const LagrangeMultiplier* mi = multipliers[i];
+      vFrictionMoments += lambda * moment(mi);
+      if (mi->Wheel)
+        mi->Wheel->Accel += mi->WheelCoeff * lambda * mi->Wheel->InvInertia;
+    }
+    else
+      vFrictionMoments += r * F;
   }
 
   FGColumnVector3 accel = vFrictionForces / in.Mass;
--- a/src/models/FGLGear.cpp
+++ b/src/models/FGLGear.cpp
@@ -145,6 +145,23 @@ FGLGear::FGLGear(Element* el, FGFDMExec* fdmex, int number, const struct Inputs&
     staticFCoeff = el->FindElementValueAsNumber("static_friction");
   if (el->FindElement("rolling_friction"))
     rollingFCoeff = el->FindElementValueAsNumber("rolling_friction");
+
+  // Optional wheel rotational degree of freedom for BOGEY contacts. Both
+  // elements are required; without them the legacy friction model is used.
+  if (eContactType == ctBOGEY && el->FindElement("wheel_radius")
+      && el->FindElement("wheel_inertia")) {
+    wheelRadius = el->FindElementValueAsNumberConvertTo("wheel_radius", "FT");
+    wheelInertia = el->FindElementValueAsNumberConvertTo("wheel_inertia", "SLUG*FT2");
+    if (wheelRadius > 0.0 && wheelInertia > 0.0) {
+      wheelSpinEnabled = true;
+      wheelSpin.InvInertia = 1.0 / wheelInertia;
+    }
+    else {
+      FGXMLLogging log(el, LogLevel::ERROR);
+      log << "wheel_radius and wheel_inertia must be positive; the wheel spin"
+          << " degree of freedom of contact " << name << " is disabled.\n";
+    }
+  }
   if (el->FindElement("retractable"))
     isRetractable = ((unsigned int)el->FindElementValueAsNumber("retractable"))>0.0?true:false;
 
@@ -269,13 +286,17 @@ void FGLGear::ResetToIC(void)
   WheelSlip = 0.0;
 
   // Initialize Lagrange multipliers
-  for (int i=0; i < 3; i++) {
+  for (int i=0; i < 4; i++) {
     LMultiplier[i].ForceJacobian.InitMatrix();
     LMultiplier[i].LeverArm.InitMatrix();
     LMultiplier[i].Min = 0.0;
     LMultiplier[i].Max = 0.0;
     LMultiplier[i].value = 0.0;
   }
+
+  wheelSpin.Rate = 0.0;
+  wheelSpin.Accel = 0.0;
+  wheelTreadSlip = 0.0;
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -285,6 +306,14 @@ const FGColumnVector3& FGLGear::GetBodyForces(void)
   double gearPos = 1.0;
 
   vFn.InitMatrix();
+
+  if (wheelSpinEnabled) {
+    // Advance the spin with the acceleration from the last friction solve, in
+    // step with FGPropagate's use of the last airframe derivatives.
+    if (!fdmex->GetTrimStatus() && in.TotalDeltaT > 0.0)
+      wheelSpin.Rate += wheelSpin.Accel * in.TotalDeltaT;
+    wheelSpin.Accel = 0.0;
+  }
 
   // Compute AGL
   FGColumnVector3 normal, terrainVel, dummy;
@@ -363,6 +392,16 @@ const FGColumnVector3& FGLGear::GetBodyForces(void)
 
       vGroundWhlVel = mT.Transposed() * vBodyWhlVel;
 
+      if (wheelSpinEnabled) {
+        FGColumnVector3 roll(mT(eX,eX), mT(eY,eX), mT(eZ,eX));
+        FGColumnVector3 vAxleVel = vBodyWhlVel + in.PQR * (wheelRadius * vGroundNormal);
+        // Wheels touching the ground at initialization or during trim roll
+        // without slip; in flight they keep their integrated spin.
+        if (fdmex->GetTrimStatus() || in.TotalDeltaT == 0.0)
+          wheelSpin.Rate = DotProduct(roll, vAxleVel) / wheelRadius;
+        wheelTreadSlip = DotProduct(roll, vAxleVel) - wheelRadius * wheelSpin.Rate;
+      }
+
       if (fdmex->GetTrimStatus() || in.TotalDeltaT == 0.0)
         compressSpeed = 0.0; // Steady state is sought during trimming
       else {
@@ -400,6 +439,7 @@ const FGColumnVector3& FGLGear::GetBodyForces(void)
       LMultiplier[ftRoll].value = 0.0;
       LMultiplier[ftSide].value = 0.0;
       LMultiplier[ftDynamic].value = 0.0;
+      LMultiplier[ftWheelBrake].value = 0.0;
 
       // Return to neutral position between 1.0 and 0.8 gear pos.
       SteerAngle *= max(gearPos-0.8, 0.0)/0.2;
@@ -412,6 +452,15 @@ const FGColumnVector3& FGLGear::GetBodyForces(void)
     // Let wheel spin down slowly
     vWhlVelVec(eX) -= 13.0 * in.TotalDeltaT;
     if (vWhlVelVec(eX) < 0.0) vWhlVelVec(eX) = 0.0;
+
+    if (wheelSpinEnabled) {
+      // No tire torque in the air. Approximate bearing drag with the legacy
+      // 13 ft/s^2 tread deceleration; applied brakes stop the wheel faster.
+      double brake = eBrakeGrp != bgNone ? in.BrakePos[eBrakeGrp] : 0.0;
+      double decrement = (13.0 + 100.0 * brake) / wheelRadius * in.TotalDeltaT;
+      wheelSpin.Rate = sign(wheelSpin.Rate) * max(0.0, fabs(wheelSpin.Rate) - decrement);
+      wheelTreadSlip = 0.0;
+    }
   }
 
   if (!fdmex->GetTrimStatus()) {
@@ -727,7 +776,10 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
 
     switch(eContactType) {
     case ctBOGEY:
-      LMultiplier[ftRoll].Max = fabs(BrakeFCoeff * vFn(eZ));
+      if (wheelSpinEnabled) // Tire grip; brakes act on the wheel instead.
+        LMultiplier[ftRoll].Max = fabs(staticFFactor * staticFCoeff * vFn(eZ));
+      else
+        LMultiplier[ftRoll].Max = fabs(BrakeFCoeff * vFn(eZ));
       LMultiplier[ftSide].Max = fabs(FCoeff * vFn(eZ));
       break;
     case ctSTRUCTURE:
@@ -749,7 +801,46 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
 
     GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftRoll]);
     GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftSide]);
+
+    if (wheelSpinEnabled) {
+      ConfigureWheelSpinRows(vWhlContactVec);
+      GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftWheelBrake]);
+    }
   }
+}
+
+//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+// With a wheel spin DOF, the tread force acts on the tire at the contact patch
+// but reaches the airframe through the axle: the airframe moment arm is the
+// axle, and the wheel receives -radius x force. Applying the force at the
+// contact point *and* spinning the wheel would count its moment twice.
+// Brakes and rolling resistance become a torque between wheel and airframe,
+// bounded by the legacy braking force times the radius, so a braked wheel
+// holds the same force as the legacy anti-skid model while spin-up drag at
+// touchdown now comes out of the aircraft's momentum.
+
+void FGLGear::ConfigureWheelSpinRows(const FGColumnVector3& vWhlContactVec)
+{
+  const FGColumnVector3 roll = LMultiplier[ftRoll].ForceJacobian;
+  const FGColumnVector3 axle = vWhlContactVec + wheelRadius * vGroundNormal;
+  // Positive spin rolls forward: spin axis = ground normal x roll direction.
+  const FGColumnVector3 spinAxis = vGroundNormal * roll;
+
+  LMultiplier[ftRoll].UseMomentJacobian = true;
+  LMultiplier[ftRoll].MomentJacobian = axle * roll;
+  LMultiplier[ftRoll].Wheel = &wheelSpin;
+  LMultiplier[ftRoll].WheelCoeff = -wheelRadius;
+
+  LagrangeMultiplier& brake = LMultiplier[ftWheelBrake];
+  brake.ForceJacobian.InitMatrix();
+  brake.LeverArm.InitMatrix();
+  brake.UseMomentJacobian = true;
+  brake.MomentJacobian = -1.0 * spinAxis;
+  brake.Wheel = &wheelSpin;
+  brake.WheelCoeff = 1.0;
+  brake.Max = fabs(BrakeFCoeff * vFn(eZ)) * wheelRadius;
+  brake.Min = -brake.Max;
+  brake.value = Constrain(brake.Min, brake.value, brake.Max);
 }
 
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -817,6 +908,13 @@ void FGLGear::bind(FGPropertyManager* PropertyManager)
                         &FGLGear::GetstaticFCoeff, &FGLGear::SetstaticFCoeff);
   property_name = base_property_name + "/dynamic_friction_coeff";
   PropertyManager->Tie( property_name.c_str(), &dynamicFCoeff );
+
+  if (wheelSpinEnabled) {
+    property_name = base_property_name + "/wheel-spin-rad_sec";
+    PropertyManager->Tie( property_name.c_str(), &wheelSpin.Rate );
+    property_name = base_property_name + "/wheel-tread-slip-fps";
+    PropertyManager->Tie( property_name.c_str(), &wheelTreadSlip );
+  }
 
   if (eContactType == ctBOGEY) {
     property_name = base_property_name + "/slip-angle-deg";
@@ -953,6 +1051,10 @@ void FGLGear::Debug(int from)
       log << "      Static Friction:  " << staticFCoeff  << "\n";
       if (eContactType == ctBOGEY) {
         log << "      Rolling Friction: " << rollingFCoeff << "\n";
+        if (wheelSpinEnabled) {
+          log << "      Wheel Radius:     " << wheelRadius << " ft\n";
+          log << "      Wheel Inertia:    " << wheelInertia << " slug*ft^2\n";
+        }
         log << "      Steering Type:    " << sSteerType[eSteerType] << "\n";
         log << "      Grouping:         " << sBrakeGroup[eBrakeGrp] << "\n";
         log << "      Max Steer Angle:  " << maxSteerAngle << "\n";
--- a/src/models/FGLGear.h
+++ b/src/models/FGLGear.h
@@ -168,6 +168,8 @@ CLASS DOCUMENTATION
             <max_steer unit="DEG"> {number | 0 | 360} </max_steer>
             <brake_group> {NONE | LEFT | RIGHT | CENTER | NOSE | TAIL} </brake_group>
             <retractable>{0 | 1}</retractable>
+            <wheel_radius unit="{FT | IN | M}"> {number} </wheel_radius>
+            <wheel_inertia unit="{SLUG*FT2 | KG*M2}"> {number} </wheel_inertia>
             <table name="{CORNERING_COEFF}" type="internal">
                 <tableData>
                     {cornering parameters}
@@ -175,6 +177,24 @@ CLASS DOCUMENTATION
             </table>
         </contact>
 @endcode
+
+    <h3>Optional wheel rotational degree of freedom (BOGEY only)</h3>
+
+    When both \<wheel_radius> and \<wheel_inertia> are given, the wheel spin
+    rate becomes a state that is resolved together with the ground friction
+    multipliers. The tire tread force is bounded by static_friction times the
+    normal force and acts on the tread slip (axle ground speed minus radius
+    times spin); it reaches the airframe through the axle while the wheel
+    receives the opposite torque, so spin-up at touchdown is paid for by the
+    aircraft's momentum. Brakes and rolling resistance become a torque between
+    the wheel and the airframe, bounded by the usual braking friction force
+    times the radius: a braked wheel holds the same force as without the spin
+    DOF. Wheels in ground contact at initialization or trim start rolling
+    without slip; in the air they spin down at the same 13 ft/s^2 tread rate as
+    wheel-speed-fps, faster with brakes applied. The extra properties
+    gear/unit[i]/wheel-spin-rad_sec and gear/unit[i]/wheel-tread-slip-fps are
+    created. Without these elements the friction model is unchanged.
+
     @author Jon S. Berndt
     @see Richard E. McFarland, "A Standard Kinematic Model for Flight Simulation at
      NASA-Ames", NASA CR-2497, January 1975
@@ -223,7 +243,7 @@ public:
   /// Damping types
   enum DampType {dtLinear=0, dtSquare};
   /// Friction types
-  enum FrictionType {ftRoll=0, ftSide, ftDynamic};
+  enum FrictionType {ftRoll=0, ftSide, ftDynamic, ftWheelBrake};
   /** Constructor
       @param el a pointer to the XML element that contains the CONTACT info.
       @param Executive a pointer to the parent executive object
@@ -311,6 +331,12 @@ public:
   bool IsBogey(void) const             { return (eContactType == ctBOGEY);}
   double GetGearUnitPos(void) const;
   double GetSteerAngleDeg(void) const { return radtodeg*SteerAngle; }
+  /// True when <wheel_radius> and <wheel_inertia> give this BOGEY a spin DOF.
+  bool HasWheelSpin(void) const { return wheelSpinEnabled; }
+  /// Wheel spin rate about the axle in rad/s (0 without a spin DOF).
+  double GetWheelSpinRate(void) const { return wheelSpin.Rate; }
+  /// Tread slip speed (axle ground speed minus radius x spin) in ft/s.
+  double GetWheelTreadSlip(void) const { return wheelTreadSlip; }
   void SetSteerAngleDeg(double angle) {
     if (eSteerType != stFixed && !Castered)
       SteerAngle = degtorad * angle;
@@ -375,7 +401,14 @@ private:
   DampType    eDampTypeRebound;
   double  maxSteerAngle;
 
-  LagrangeMultiplier LMultiplier[3];
+  LagrangeMultiplier LMultiplier[4];
+
+  // Optional wheel rotational degree of freedom.
+  bool wheelSpinEnabled = false;
+  double wheelRadius = 0.0;  // ft
+  double wheelInertia = 0.0; // slug*ft^2
+  WheelSpinDOF wheelSpin;
+  double wheelTreadSlip = 0.0; // ft/s
 
   // NO std::shared_ptr<FGGroundReactions> here, to avoid circular references
   // since FGGroundReactions owns the instances of FGLGear.
@@ -392,6 +425,7 @@ private:
   void ComputeVerticalStrutForce(void);
   void ComputeGroundFrame(void);
   void ComputeJacobian(const FGColumnVector3& vWhlContactVec);
+  void ConfigureWheelSpinRows(const FGColumnVector3& vWhlContactVec);
   void UpdateForces(void);
   void SetstaticFCoeff(double coeff);
   void CrashDetect(void);
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -41,6 +41,7 @@ set(PYTHON_TESTS ResetOutputFiles
                  TestStdAtmosphere
                  TestMiscellaneous
                  TestGndReactions
+                 TestWheelSpin
                  TestIntegrators
                  TestSwitch
                  TestGain
--- a/tests/TestWheelSpin.py
+++ b/tests/TestWheelSpin.py
@@ -1,0 +1,169 @@
+# TestWheelSpin.py
+#
+# Checks the optional wheel rotational degree of freedom of BOGEY contacts
+# (<wheel_radius> and <wheel_inertia>).
+#
+# Copyright (c) 2026 Felipegalind0
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, see <http://www.gnu.org/licenses/>
+#
+
+import os
+import shutil
+import xml.etree.ElementTree as et
+from JSBSim_utils import JSBSimTestCase, RunTest
+
+# Estimated C172 tire/wheel assemblies (radius in m, spin inertia in kg*m^2).
+WHEELS = {'NOSE': (0.18, 0.052), 'LEFT_MAIN': (0.22, 0.124),
+          'RIGHT_MAIN': (0.22, 0.124)}
+FT_PER_M = 1.0 / 0.3048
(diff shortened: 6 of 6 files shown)


bcoconni:
Thanks @Felipegalind0 for your contribution. This PR is quite complex and it took me some time to understand what it does and how.

For everyone, including my future self, the idea is to make wheels independent bodies from the aircraft. At its very core, JSBSim assumes that the aircraft is a single body and that other bodies are only interacting with the aircraft through the forces and moments that they apply to it.

This is typically the case for propellers where `FGPropeller` is inheriting from `FGForce` via the `FGThruster` class:
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGPropeller.h#L169
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGThruster.h#L77
The class `FGPropeller` manages its own body (the propeller!):
* The propeller rotation rate is updated using its inertia and the angular momentum $I\dot{\theta}=\sum_i M_i$:
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGPropeller.cpp#L294
* Forces and moments are computed and applied to the aircraft:
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGPropeller.cpp#L282
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGPropeller.cpp#L300

The idea of the PR is to apply the same logic to the wheels using `FGLGear`: they become independent bodies with their own degree of freedom (DOF) and they apply a force and a moment to the aircraft.

Free body diagrams (force and moment) of a wheel

Above are free body diagrams of the wheel: the friction force $F_{\text{roll}}$ is the friction that the ground applies to the wheel and that the wheel applies to the aircraft axle (point A). Hence - now that the wheel is an independent body - the friction force is applied on the aircraft at the point A instead of the point C, so the code below is updated by the PR:
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/FGLGear.cpp#L725
Note that in this PR, $F_{\text{roll}}$ is used for 2 different purposes:
1. To apply the friction force to the aircraft body (as is currently the case)
2. To compute the driving torque $R\cdot F_{\text{roll}}$ (in blue on the picture above) that will be applied to the wheel.

### Braking management

Unlike in the current code, the friction coefficient for $F_{\text{roll}}$ is always the static coefficient $\mu_{\text{static}}$ and is independent of the brake position `BrakePos`.

To keep the braking feature, the patch introduces the braking torque $\tau_{brake}$ (in yellow on the picture above) that is currently not modeled by `FGLGear`. Hence the new Lagrange multiplier `ftWheelBrake` whose job is to zero out the relative motion between the wheel and the aircraft.

The braking torque $\tau_{brake}$ obviously depends on the brake position `BrakePos` and is using `BrakeFCoeff`:
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/FGLGear.cpp#L590-L593

It is following that the wheel applies permanently a resisting torque $\mu_{\text{rolling}}F_{\text{roll}}R$ to the axle, whatever the braking position and this has 2 consequences:
1. The rolling friction coefficient is now characterizing the friction between the wheel and its axle. Legacy behavior is to assume that the rolling friction coefficient characterizes the hysteresis in the compression/expansion of the tire while rolling. There is nothing problematic in these two different interpretations of the rolling friction coefficient but model designers need to know that.
2. When the wheel is spinning $\tau_{\text{brake}}$ will **always** be equal to the maximum torque allowed by the Coulomb law i.e. $\tau_{\text{brake}}=\mu_{\text{brake}}F_{\text{roll}}R$ because the `ftWheelBrake` multiplier is targetting a *no wheel spin* condition.

So, how is braking working, now? When applying some brake, $\tau_{brake}$ increases, $\omega$ decreases so $F_{\text{roll}}$ increases to try to restore the no slip condition $R\omega=v_A$.

Note that $F_{\text{roll}}$ can reach $\mu_{\text{static}}F_n$ with the wheel still spinning.

### Wheel dynamics

The driving torque and the braking torque are then used to compute the spin rate $\omega$ using the classic angular momentum equation 

$$I\dot{\omega}=R\cdot F_{\text{roll}}-\tau_{\text{brake}}$$

Note that there is a subtlety here: $\omega$ is the wheel spin rate **measured in the NED frame** (so-called *absolute spin rate* in the PR). This is deviating from what most people would expect i.e. a spin rate measured in the body frame:
* This allows to define the no slip condition as $R\omega=V_{\text{axle}}$ instead of $R\omega=V_C$ where $V_C$ is the velocity at the contact point with respect to the ground (i.e. $V_C=V_G+\Omega_{\text{aircraft}}\times GC$). It simplifies the code a bit but;
* The condition that the `ftWheelBrake` multiplier has to resolve is not $\omega=0$ but 

$$\omega=\vec{\Omega}_{\text{aircraft}}\cdot\vec{u}_{\text{axle}}$$

### No slip condition vs *no wheel spin* condition

With respect to that point, the code in the PR is not so easy to understand because it is far from obvious that, for the `ftWheelBrake` multiplier, the following code actually simplifies to the equation above:
```c++
    double wheelTerm = (mi->Wheel && dt > 0.) ? mi->WheelCoeff * mi->Wheel->Rate / dt : 0.;
    rhs[i] = -(DotProduct(U, vdot) + DotProduct(moment(mi), wdot) + wheelTerm)/d;
```
As a reminder, according to the following code:
https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/FGAccelerations.cpp#L265-L273
we have (the terrain movement is ignored to simplify the analysis)

$$\text{vdot}=\dfrac{d\vec{v}_G}{dt}+\dfrac{\vec{v}_G}{\Delta t}$$

$$\text{wdot}=\dfrac{d\vec{\Omega}_{\text{aircraft}}}{dt}+\dfrac{\vec{\Omega}_{\text{aircraft}}}{\Delta t}$$

So, for `ftWheelBrake`, we have $U=0$, `moment(mi)` is equal to $-u_{\text{axle}}$, `mi->WheelCoeff == 1.0`, and `mi->Wheel->Rate` is obviously $\omega$ so we have:

$$\text{rhs}=\left(-\dfrac{d\vec{\Omega}_{\text{aircraft}}}{dt}\cdot\vec{u}_{\text{axle}}+\dfrac{\omega-\vec{\Omega}_{\text{aircraft}}\cdot\vec{u}_{\text{axle}}}{\Delta t}\right)\dfrac{1}{d}$$

where one can recognize the *no wheel spin* condition $\omega-\Omega_{\text{aircraft}}\cdot u_{\text{axle}}$

*Note: You can ignore the time derivation and the division by `d` because they are technical details unrelated to the point I am trying to make.*

On the other hand, for the `ftRoll` multiplier, the same C++ code leads to the no slip condition because $U=u_{\text{roll}}$, `moment(mi)` is equal to $GA\times u_{\text{roll}}$ and `mi->Wheel->Rate` is equal to $-R$ so we have

$$\text{rhs}=\left(\vec{u}_{\text{roll}}\cdot\left(\dfrac{d\vec{v}_G}{dt}+\dfrac{\vec{v}_G}{\Delta t}\right)+(\vec{GA}\times\vec{u}_{\text{roll}})\cdot\left(\dfrac{d\vec{\Omega}_{\text{aircraft}}}{dt}+\dfrac{\vec{\Omega}_{\text{aircraft}}}{\Delta t}\right)+\dfrac{-R\omega}{\Delta t}\right)\dfrac{1}{d}$$

which simplifies to (using the identity $(A\times B)\cdot C=(C\times A)\cdot B$)

$$\text{rhs}=\left(\vec{u}_{\text{roll}}\cdot\left(\dfrac{d\vec{v}_G}{dt}+\dfrac{d\vec{\Omega}_{\text{aircraft}}}{dt}\times\vec{GA}\right)-\dfrac{R\omega-\vec{u}_{\text{roll}}\cdot(\vec{v}_G+\vec{\Omega}_{\text{aircraft}}\times\vec{GA})}{\Delta t}\right)\dfrac{1}{d}$$

and since the velocity at the axle is $v_A=v_G+{\Omega}_{\text{aircraft}}\times GA$, we finally have:

$$\text{rhs}=\left(\vec{u}_{\text{roll}}\cdot\dfrac{d\vec{v}_A}{dt}-\dfrac{R\omega-\vec{u}_{\text{roll}}\cdot\vec{v}_A}{\Delta t}\right)\dfrac{1}{d}$$

where we recognize the no slip condition $R\omega-u_{\text{roll}}\cdot v_A$.

So, great! the code works. However it has been a pain to understand that this code covers two entirely different conditions using some tricks: setting $U$ to 0 and the not-very-well-named `WheelCoeff` to either 1 or $-R$!

### Conclusion (for now)

As claimed, the PR brings a better and more realistic handling of braking, wheel spin up energy, wheel spin up when touching ground, etc.

However, the code deserves some clean up and there are a couple of errors that need to be addressed.

To be continued... (sorry I'm running out of time for now but questions/comments on my analysis above are welcome).

seanmcleod70:
Merci beaucoup @bcoconni for the addition of your writeup. In an ideal world every non-trivial bit of physics code in JSBSim would include a URL in the class/method pointing to writeups like this, i.e. markdown/jupyter style with snippets of explanatory text, equations, diagrams, snippets of code.

src/math/LagrangeMultiplier.h
```diff
@@ -48,12 +48,29 @@ CLASS DOCUMENTATION
 CLASS DECLARATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
+/** Rotational state of a wheel whose spin is resolved together with the
+    ground friction multipliers (see the <wheel_inertia> element of FGLGear). */
+struct WheelSpinDOF {
+  double InvInertia = 0.0; ///< 1 / spin inertia about the axle [1/(slug*ft^2)]
```
bcoconni, line 54: Use `Jinv` for the inverse of the inertia for consistency with the rest of the code:
```suggestion
  double Jinv = 0.0; ///< 1 / spin inertia about the axle [1/(slug*ft^2)]
```

src/math/LagrangeMultiplier.h
```diff
@@ -48,12 +48,29 @@ CLASS DOCUMENTATION
...
+  double Rate = 0.0;       ///< absolute spin rate about the axle [rad/s]
+  double Accel = 0.0;      ///< spin acceleration from the last friction solve [rad/s^2]
+};
+
 struct LagrangeMultiplier {
   FGColumnVector3 ForceJacobian;
   FGColumnVector3 LeverArm;
```
bcoconni, line 61: You should remove `LeverArm` now that you have introduced the moment jacobian, it would simplify the code using it everywhere.

src/models/FGAccelerations.cpp
```diff
@@ -241,11 +241,42 @@ void FGAccelerations::CalculateFrictionForces(double dt)
...
+  bool generalized = false;
+  for (auto m : multipliers) {
+    if (m->UseMomentJacobian || m->Wheel) {
+      generalized = true;
+      if (m->Wheel) m->Wheel->Accel = 0.0; // Refilled by this solve
+    }
+  }
+  auto moment = [](const LagrangeMultiplier* m) {
+    return m->UseMomentJacobian ? m->MomentJacobian : m->LeverArm * m->ForceJacobian;
+  };
```
bcoconni, line 256: Once having removed `LeverArm`, you will no longer need the lambda function `moment`.

src/models/FGAccelerations.cpp
```diff
@@ -241,11 +241,42 @@ void FGAccelerations::CalculateFrictionForces(double dt)
...
+        a[i*n+j] = a[j*n+i];
+
+      for (unsigned int j=i; j < n; j++) {
+        const LagrangeMultiplier* mj = multipliers[j];
+        a[i*n+j] = DotProduct(mj->ForceJacobian, v1) + DotProduct(moment(mj), v2);
+        if (mi->Wheel && mi->Wheel == mj->Wheel)
+          a[i*n+j] += mi->WheelCoeff * mj->WheelCoeff * mi->Wheel->InvInertia;
+      }
+      continue;
+    }
```
bcoconni, line 278: This loop is basically a duplication of the code just below with only the specificity of handling the wheel:
```c++
        if (mi->Wheel && mi->Wheel == mj->Wheel)
          a[i*n+j] += mi->WheelCoeff * mj->WheelCoeff * mi->Wheel->InvInertia;
```
Since this code is guarded by the `if (mi->Wheel)` it can be integrated in the legacy code rather than duplicating the loops.

src/models/FGLGear.cpp
```diff
@@ -727,7 +776,10 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
 
     switch(eContactType) {
     case ctBOGEY:
-      LMultiplier[ftRoll].Max = fabs(BrakeFCoeff * vFn(eZ));
+      if (wheelSpinEnabled) // Tire grip; brakes act on the wheel instead.
+        LMultiplier[ftRoll].Max = fabs(staticFFactor * staticFCoeff * vFn(eZ));
```
bcoconni, line 780: This code should be located in the new method `ConfigureWheelSpinRows`. With the current code, it is far from obvious that when a wheel is accounted for, then the friction coefficient is `staticFFactor * staticFCoeff`. The reason for that is that with the current code in the PR, I was expecting the friction coefficient of `ftRoll` to be updated in `ConfigureWheelSpinRows`.

src/models/FGLGear.cpp
```diff
@@ -749,9 +801,48 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
 
     GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftRoll]);
     GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftSide]);
+
+    if (wheelSpinEnabled) {
+      ConfigureWheelSpinRows(vWhlContactVec);
+      GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftWheelBrake]);
+    }
```
bcoconni, line 808: This code should be moved before the calls to `Constrain`:
```c++
    LMultiplier[ftRoll].value = Constrain(LMultiplier[ftRoll].Min, LMultiplier[ftRoll].value, LMultiplier[ftRoll].Max);
    LMultiplier[ftSide].value = Constrain(LMultiplier[ftSide].Min, LMultiplier[ftSide].value, LMultiplier[ftSide].Max);
```
The reason being that once the update of `LMultiplier[ftRoll].Max` will be moved to `ConfigureWheelSpinRows`, `Constrain` shall be called **after** `LMultiplier[ftRoll].Max` has been updated.

src/models/FGLGear.cpp
```diff
@@ -749,9 +801,48 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
...
+// contact point *and* spinning the wheel would count its moment twice.
+// Brakes and rolling resistance become a torque between wheel and airframe,
+// bounded by the legacy braking force times the radius, so a braked wheel
+// holds the same force as the legacy anti-skid model while spin-up drag at
+// touchdown now comes out of the aircraft's momentum.
+
+void FGLGear::ConfigureWheelSpinRows(const FGColumnVector3& vWhlContactVec)
+{
+  const FGColumnVector3 roll = LMultiplier[ftRoll].ForceJacobian;
+  const FGColumnVector3 axle = vWhlContactVec + wheelRadius * vGroundNormal;
```
bcoconni, line 825: These variable names are misleading because `roll` is actually the direction of the roll friction force (unit vector) and `axle` is not the axle direction but the axle **position** with respect to the CG. And since the code is historically using *lever arm* to designate this position, then do so here as well.

src/models/FGLGear.cpp
```diff
@@ -818,6 +909,13 @@ void FGLGear::bind(FGPropertyManager* PropertyManager)
   property_name = base_property_name + "/dynamic_friction_coeff";
   PropertyManager->Tie( property_name.c_str(), &dynamicFCoeff );
 
+  if (wheelSpinEnabled) {
+    property_name = base_property_name + "/wheel-spin-rad_sec";
+    PropertyManager->Tie( property_name.c_str(), &wheelSpin.Rate );
```
bcoconni, line 914: Do not export the so-called *absolute spin rate*. Instead export the wheel spin rate measured with respect to the body frame. The latter is the spin rate that the Flight Control System and the Anti Blocking System will have access to (using a phonic wheel) in the real world.

Also it is the spin rate relative to the aircraft that 3D modelers need to display the wheels since models are generally organized through a hierarchy of bodies, each one being positioned relative to its parent in the hierarchy.

src/math/LagrangeMultiplier.h
```diff
@@ -48,12 +48,29 @@ CLASS DOCUMENTATION
 CLASS DECLARATION
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*/
 
+/** Rotational state of a wheel whose spin is resolved together with the
+    ground friction multipliers (see the <wheel_inertia> element of FGLGear). */
+struct WheelSpinDOF {
+  double InvInertia = 0.0; ///< 1 / spin inertia about the axle [1/(slug*ft^2)]
+  double Rate = 0.0;       ///< absolute spin rate about the axle [rad/s]
```
bcoconni, line 55: There is no "absolute" in JSBSim. Instead tell that it is the spin rate measured with respect to the NED frame.
```suggestion
  double Rate = 0.0;       ///< spin rate about the axle measured in the NED frame [rad/s]
```

bcoconni (changes requested):
A few comments that should hopefully simplify and clarify the code.

After these have been implemented, the code should no longer need the lambda function `moment`, neither the members `LeverArm` or `UseMomentJacobian`.

After these have been discussed/implemented/rejected, I will have some other comments.

src/models/FGAccelerations.cpp
bcoconni: And the `continue` keyword to avoid an `else` is debatable...

src/math/LagrangeMultiplier.h
Felipegalind0: Renamed `InvInertia` to `Jinv` in “Use explicit moment Jacobians for friction.”

src/math/LagrangeMultiplier.h
Felipegalind0: Removed `LeverArm` and `UseMomentJacobian` in “Use explicit moment Jacobians
for friction.” All rows now supply their moment Jacobian.

I rechecked all 61 shipped XML inputs on the final candidate. Base and
original PR data match byte-for-byte; the cleanup changes 60 of 102 files
across 31 of 57 successful simulations. Crosswind reaches Δu = 0.0540 ft/s
and Δgear Fx = 197 lbf; the moored ZLT reaches Δgear yaw moment = 20,929
lbf·ft. F-16 Δu reaches 71.6 ft/s after severe ground impacts. These are
not uniformly rounding-sized output differences. The final candidate matches
the isolated cleanup probe exactly, and the PR body now reports these limits
instead of claiming existing aircraft are unaffected.

src/models/FGAccelerations.cpp
Felipegalind0: Removed the lambda in “Use explicit moment Jacobians for friction.” Assembly,
RHS and final moment accumulation now read `MomentJacobian` directly.

src/math/LagrangeMultiplier.h
Felipegalind0: Removed “absolute” in “Report wheel spin relative to the airframe.” I used
“ground reference frame,” matching the rate supplied to the solve.
`FGPropagate::GetPQR()` documents that rate as ECEF-relative; a transported
NED frame would require another correction. The state and its reference are
unchanged; the public property is now relative to the airframe.

src/models/FGAccelerations.cpp
Felipegalind0: Merged the matrix assembly, RHS and accumulation paths in “Use explicit
moment Jacobians for friction.” The wheel term is conditional on the two
rows sharing a wheel; the `generalized` branch and its `continue` are gone.
I also renamed `WheelCoeff` to `WheelJacobian`.

src/models/FGLGear.cpp
Felipegalind0: Moved the wheel roll row's `Max` and `Min` into `ConfigureWheelSpinRows` in
“Configure wheel friction bounds before clamping.” The static tire-grip
bound is now alongside the wheel row, separate from the brake-torque bound.

src/models/FGLGear.cpp
Felipegalind0: “Configure wheel friction bounds before clamping” calls the configuration
before the roll and side warm-start clamps. Registration still runs in
roll, side, brake order.

src/models/FGLGear.cpp
Felipegalind0: Renamed them `rollDirection` and `axleLeverArm` in “Configure wheel friction
bounds before clamping.” “Document wheel friction and airborne spin-down”
also clarifies that the lever arm runs from the CG to the axle.

src/models/FGLGear.cpp
Felipegalind0: “Report wheel spin relative to the airframe” keeps `wheel-spin-rad_sec` and
exports `Rate - dot(in.PQR, spinAxis)`. Writes apply the inverse conversion,
so the property remains writable. The constraint solve retains its internal
state. The same commit fixes airborne drag to reduce this relative rate and
adds nonzero body-rate and steering regressions.

Felipegalind0:
For stationary terrain, your reading matches the two constraints: the roll
row drives axle tread slip toward zero, while the brake row drives wheel
spin relative to the airframe toward zero, within their bounds. Both remain
in the shared solve, now with explicit moment Jacobians and `WheelJacobian`
naming. The internal reference is unchanged; the exported rate is
airframe-relative. I used “ground reference frame” in the comments because
the supplied `in.PQR` comes from the ECEF-relative `FGPropagate::GetPQR()`.

“Document wheel friction and airborne spin-down” explains the enabled
`rolling_friction` interpretation as wheel-to-axle resistance in the
`FGLGear.h` class documentation. The comment above `ConfigureWheelSpinRows`
links your analysis and states the two rows, addressing Sean's link request.

I also changed two things outside the inline requests: airborne drag now
reduces airframe-relative spin, and the 13/100 ft/s² spin-down constants are
named with their provenance. The 100 ft/s² extra brake term is an authored,
uncalibrated approximation; I retained its value rather than deriving a
torque from zero airborne normal load. A CxxTest now checks the constraint
signs, shared-wheel coupling, clamping and zero-dt case in the Python-disabled
test configuration.

The final candidate `499c3832` passes 79/79 CTest targets with Python enabled
and 21/21 CxxTests with Python disabled. The new native test also detects
omitted or reversed shared-wheel coupling in mutation checks.

I ran all 61 shipped XML inputs with per-step logging on the base, original
PR, isolated cleanup probe and final candidate. Original/base data match
byte-for-byte, as do final/probe data. Base/final differ in 60 of 102 files
across 31 of 57 successful simulations. Crosswind and moored-ZLT contact
forces/moments change materially, and the F-16 trajectory diverges after
severe impacts. The body gives the measurements and removes the “unaffected”
claim. One common trim failure and three non-runscript inputs are accounted
for separately. These results are specific to this macOS arm64 build.

src/math/LagrangeMultiplier.h
bcoconni: You are entirely correct that `in.vPQR` is the rotation rate vector of the body relative to the ECEF frame, not the NED frame as I wrongly implied. My point was not to request a transformation to a frame but to be specific about which frame the rate is measured in.

The *ground reference frame* does not exist in JSBSim, please use the proper name which in this case is the ECEF frame. In addition to that it is not the spin rate *about the axle*, because the same rate cannot be measured relative to the ground and relative to its axle. It's either one or the other.

So please change the comment for the member `Rate` to `wheel spin rate in the ECEF frame`.

src/models/FGAccelerations.cpp
```diff
@@ -311,13 +319,13 @@ void FGAccelerations::CalculateFrictionForces(double dt)
...
+    const LagrangeMultiplier* mi = multipliers[i];
+    double lambda = mi->value;
 
-    FGColumnVector3 F = lambda * U;
-    vFrictionForces += F;
-    vFrictionMoments += r * F;
+    vFrictionForces += lambda * mi->ForceJacobian;
+    vFrictionMoments += lambda * mi->MomentJacobian;
+    if (mi->Wheel)
+      mi->Wheel->Accel += mi->WheelJacobian * lambda * mi->Wheel->Jinv;
```
bcoconni, line 328: There is no reason to delegate the computation of the wheel spin acceleration to `FGAccelerations`. This should be internal to `FGLGear` just as the update of the propeller rate is internal to `FGPropeller`. All the data needed to compute the wheel spin acceleration are available to `FGLGear`.
```suggestion
    vFrictionForces += lambda * mi->ForceJacobian;
    vFrictionMoments += lambda * mi->MomentJacobian;
```

src/models/FGAccelerations.cpp
```diff
@@ -241,23 +241,29 @@ void FGAccelerations::CalculateFrictionForces(double dt)
   // If no gears are in contact with the ground then return
   if (!n) return;
 
+  // Wheel spin accelerations are refilled by this solve.
+  for (auto m : multipliers)
+    if (m->Wheel) m->Wheel->Accel = 0.0;
+
```
bcoconni, line 247: Same as previously mentioned: this should be removed from `FGAccelerations` and the computation of the spin rate acceleration should be made in `FGLGear`.
```suggestion
```

src/math/LagrangeMultiplier.h
```diff
@@ -48,12 +48,30 @@ CLASS DOCUMENTATION
...
 
+/** Rotational state of a wheel whose spin is resolved together with the
+    ground friction multipliers (see the <wheel_inertia> element of FGLGear).
+    Rate uses the same reference as the airframe rates FGLGear receives in
+    in.PQR (FGPropagate::GetPQR(), relative to the ECEF frame), so Rate minus
+    the airframe rate about the axle is the spin relative to the airframe. */
+struct WheelSpinDOF {
+  double Jinv = 0.0;       ///< 1 / spin inertia about the axle [1/(slug*ft^2)]
+  double Rate = 0.0;       ///< spin rate about the axle in the ground reference frame [rad/s]
+  double Accel = 0.0;      ///< spin acceleration from the last friction solve [rad/s^2]
```
bcoconni, line 59: Once the computation of the wheel spin acceleration is moved to `FGLGear`, you no longer need the member `Accel`.
```suggestion
 double Jinv = 0.0;       ///< 1 / spin inertia about the axle [1/(slug*ft^2)]
 double Rate = 0.0;       ///< spin rate about the axle in the ground reference frame [rad/s]
```

src/math/LagrangeMultiplier.h
```diff
@@ -48,12 +48,30 @@ CLASS DOCUMENTATION
...
   double Max;
   double value;
+  /// Moment applied to the airframe per unit multiplier: the lever arm cross
+  /// ForceJacobian for a force, or a torque acting between wheel and airframe.
+  FGColumnVector3 MomentJacobian;
+  /// Wheel spin degree of freedom coupled to this multiplier (or nullptr) and
+  /// the multiplier's Jacobian entry on that wheel's spin rate, i.e. the torque
+  /// on the wheel per unit multiplier [ft for a force, 1 for a torque].
+  WheelSpinDOF* Wheel = nullptr;
+  double WheelJacobian = 0.0;
```
bcoconni, line 74: Not sure why `WheelJacobian` is not part of the above structure `WheelSpinDOF`?

src/models/FGAccelerations.cpp
```diff
@@ -279,10 +285,12 @@ void FGAccelerations::CalculateFrictionForces(double dt)
...
     double d = a[i*n+i];
-    FGColumnVector3 U = multipliers[i]->ForceJacobian;
-    FGColumnVector3 r = multipliers[i]->LeverArm;
+    const LagrangeMultiplier* mi = multipliers[i];
 
-    rhs[i] = -DotProduct(U, vdot + wdot*r)/d;
+    // The wheel has no other applied torque here; drive its part of the
+    // constraint velocity to zero over the step like the airframe's.
+    double wheelTerm = (mi->Wheel && dt > 0.) ? mi->WheelJacobian * mi->Wheel->Rate / dt : 0.;
+    rhs[i] = -(DotProduct(mi->ForceJacobian, vdot) + DotProduct(mi->MomentJacobian, wdot) + wheelTerm)/d;
```
bcoconni, line 293: This is not very legible. What about
```suggestion
    rhs[i] = DotProduct(mi->ForceJacobian, vdot) + DotProduct(mi->MomentJacobian, wdot);
    if (mi->Wheel && dt > 0.)
      rhs[i] += mi->WheelJacobian * mi->Wheel->Rate / dt;
    rhs[i] /= -d;
```

src/models/FGLGear.cpp
```diff
@@ -286,6 +307,14 @@ const FGColumnVector3& FGLGear::GetBodyForces(void)
 
   vFn.InitMatrix();
 
+  if (wheelSpinEnabled) {
+    // Advance the spin with the acceleration from the last friction solve, in
+    // step with FGPropagate's use of the last airframe derivatives.
+    if (!fdmex->GetTrimStatus() && in.TotalDeltaT > 0.0)
+      wheelSpin.Rate += wheelSpin.Accel * in.TotalDeltaT;
+    wheelSpin.Accel = 0.0;
```
bcoconni, line 315: All the data are available here to compute the wheel spin rate. You no longer need the `Accel` member and therefore no longer need to reset it to 0.0.
```suggestion
    if (!fdmex->GetTrimStatus() && in.TotalDeltaT > 0.0) {
      const LagrangeMultiplier& lm_roll = LMultiplier[ftRoll];
      const LagrangeMultiplier& lm_brake = LMultiplier[ftBrake];
      wheelSpin.Rate += (lm_roll.WheelJacobian * lm_roll.value + lm_brake.WheelJacobian * lm_brake.value) * in.TotalDeltaT * wheelSpin.Jinv;
    }
```

src/models/FGLGear.cpp
```diff
@@ -269,13 +286,17 @@ void FGLGear::ResetToIC(void)
...
     LMultiplier[i].ForceJacobian.InitMatrix();
-    LMultiplier[i].LeverArm.InitMatrix();
+    LMultiplier[i].MomentJacobian.InitMatrix();
     LMultiplier[i].Min = 0.0;
     LMultiplier[i].Max = 0.0;
     LMultiplier[i].value = 0.0;
   }
+
+  wheelSpin.Rate = 0.0;
+  wheelSpin.Accel = 0.0;
```
bcoconni, line 298: No longer needed.
```suggestion
```

src/models/FGLGear.cpp
```diff
@@ -818,6 +973,14 @@ void FGLGear::bind(FGPropertyManager* PropertyManager)
...
   PropertyManager->Tie( property_name.c_str(), &dynamicFCoeff );
 
+  if (wheelSpinEnabled) {
+    property_name = base_property_name + "/wheel-spin-rad_sec";
+    PropertyManager->Tie( property_name.c_str(), (FGLGear*)this,
+                          &FGLGear::GetWheelSpinRate, &FGLGear::SetWheelSpinRate);
+    property_name = base_property_name + "/wheel-tread-slip-fps";
+    PropertyManager->Tie( property_name.c_str(), &wheelTreadSlip );
+  }
+
```
bcoconni, line 983: The condition `wheelSpinEnabled` can only be true if `eContactType == ctBOGEY` so this code should be inside the `if (eContactType == ctBOGEY) {` below, shouldn't it?

Also is there still some point in exporting the property `gear/unit[*]/wheel-speed-fps` when wheels are modeled? I personally don't think so.

src/models/FGLGear.cpp
```diff
@@ -722,8 +792,8 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
 
     LMultiplier[ftRoll].ForceJacobian = mT * FGColumnVector3(1.,0.,0.);
     LMultiplier[ftSide].ForceJacobian = mT * FGColumnVector3(0.,1.,0.);
-    LMultiplier[ftRoll].LeverArm = vWhlContactVec;
-    LMultiplier[ftSide].LeverArm = vWhlContactVec;
+    LMultiplier[ftRoll].MomentJacobian = vWhlContactVec * LMultiplier[ftRoll].ForceJacobian;
+    LMultiplier[ftSide].MomentJacobian = vWhlContactVec * LMultiplier[ftSide].ForceJacobian;
```
bcoconni, line 796: This code would be easier to read using references as you did in `ConfigureWheelSpinRows`:
```c++
    LagrangeMultiplier& lm_roll = LMultiplier[ftRoll];
    LagrangeMultiplier& lm_side = LMultiplier[ftSide];

    lm_roll.ForceJacobian = mT * FGColumnVector3(1.,0.,0.);
    lm_side.ForceJacobian = mT * FGColumnVector3(0.,1.,0.);
    lm_roll.MomentJacobian = vWhlContactVec * lm_roll.ForceJacobian;
    lm_side.MomentJacobian = vWhlContactVec * lm_side.ForceJacobian;
```

src/models/FGLGear.cpp
```diff
@@ -739,6 +809,9 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
     LMultiplier[ftRoll].Min = -LMultiplier[ftRoll].Max;
     LMultiplier[ftSide].Min = -LMultiplier[ftSide].Max;
 
+    if (wheelSpinEnabled)
+      ConfigureWheelSpinRows(vWhlContactVec);
+
```
bcoconni, line 814: You should call `ConfigureWheelSpinRows` before `LMultiplier[ftRoll].Min` is set so that you don't need to set it in `ConfigureWheelSpinRows`:
```suggestion
    if (wheelSpinEnabled) {
      ConfigureWheelSpinRows(vWhlContactVec);
      GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftWheelBrake]);
    }

    lm_roll.Min = -lm_roll.Max;
    lm_side.Min = -lm_side.Max;
```
Also replacing `LMultiplier[ftRoll]` and `LMultiplier[ftSide]` by their references `lm_roll` and `lm_side`.

And `GroundReactions->RegisterLagrangeMultiplier` can be called here because it the Gauss-Seidel algorithm is insensitive to the order of the matrix rows.

src/models/FGLGear.cpp
```diff
@@ -749,9 +822,91 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
 
     GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftRoll]);
     GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftSide]);
+
+    if (wheelSpinEnabled)
+      GroundReactions->RegisterLagrangeMultiplier(&LMultiplier[ftWheelBrake]);
```
bcoconni, line 827: Not sure why you have split that from the call to `ConfigureWheelSpinRows` above ? Registering the wheel brake Lagrange multiplier first will only change the order of the matrix rows for the projected Gauss-Seidel algorithm which is harmless.

src/models/FGLGear.cpp
```diff
@@ -749,9 +822,91 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
...
+  const FGColumnVector3 axleLeverArm = vWhlContactVec + wheelRadius * vGroundNormal;
+  // Positive spin rolls forward: spin axis = ground normal x roll direction.
+  const FGColumnVector3 spinAxis = vGroundNormal * rollDirection;
+
+  // Tire grip; brakes act on the wheel instead.
+  LMultiplier[ftRoll].Max = fabs(staticFFactor * staticFCoeff * vFn(eZ));
+  LMultiplier[ftRoll].Min = -LMultiplier[ftRoll].Max;
+  LMultiplier[ftRoll].MomentJacobian = axleLeverArm * rollDirection;
+  LMultiplier[ftRoll].Wheel = &wheelSpin;
+  LMultiplier[ftRoll].WheelJacobian = -wheelRadius;
```
bcoconni, line 867: For improved legibility, you should use a reference to `LMultiplier[ftRoll]` just as you did for `LMultiplier[ftWheelBrake]`:
```suggestion
  LagrangeMultiplier& lm_roll = LMultiplier[ftRoll];
  const FGColumnVector3 rollDirection = lm_roll.ForceJacobian;
  const FGColumnVector3 axleLeverArm = vWhlContactVec + wheelRadius * vGroundNormal;
  // Positive spin rolls forward: spin axis = ground normal x roll direction.
  const FGColumnVector3 spinAxis = vGroundNormal * rollDirection;

  // Tire grip; brakes act on the wheel instead.
  lm_roll.Max = fabs(staticFFactor * staticFCoeff * vFn(eZ));
  lm_roll.MomentJacobian = axleLeverArm * rollDirection;
  lm_roll.Wheel = &wheelSpin;
  lm_roll.WheelJacobian = -wheelRadius;
```
Also you don't need to set `LMultiplier[ftRoll].Min` since the call to `ConfigureWheelSpinRows` will be made before `LMultiplier[ftRoll].Min` is updated.

src/models/FGLGear.cpp
```diff
@@ -749,9 +822,91 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
...
+
+  LagrangeMultiplier& brake = LMultiplier[ftWheelBrake];
+  brake.ForceJacobian.InitMatrix();
```
bcoconni, line 870: This is useless. `brake.ForceJacobian` is never modified and it has already been set to zero at line 290.
```suggestion
```

src/models/FGLGear.cpp
```diff
@@ -749,9 +822,91 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
...
+// In ground contact this is the axis of the brake row, ground normal (pointing
+// away from the ground) cross the rolling direction projected on the ground.
+// In the air it follows the current gear orientation and steering angle.
+
+FGColumnVector3 FGLGear::GetWheelSpinAxis(void) const
+{
+  if (WOW)
+    return vGroundNormal * FGColumnVector3(mT(eX,eX), mT(eY,eX), mT(eZ,eX));
+
+  return mTGear * FGColumnVector3(sin(SteerAngle), -cos(SteerAngle), 0.0);
```
bcoconni, line 890: The spin axis is the same axis as the *side axis* computed in `ComputeGroundFrame`:
https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/models/FGLGear.cpp#L438-L450
So you do not need to compute it again:
```suggestion
  if (WOW)
    return FGColumnVector3(mT(eX,eY), mT(eY,eY), mT(eZ,eY));

  return mTGear * FGColumnVector3(sin(SteerAngle), -cos(SteerAngle), 0.0);
```

src/models/FGLGear.cpp
```diff
@@ -701,7 +771,7 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
     LMultiplier[ftDynamic].ForceJacobian = mT * velocityDirection;
     LMultiplier[ftDynamic].Max = 0.;
     LMultiplier[ftDynamic].Min = -fabs(staticFFactor * dynamicFCoeff * vFn(eZ));
-    LMultiplier[ftDynamic].LeverArm = vWhlContactVec;
+    LMultiplier[ftDynamic].MomentJacobian = vWhlContactVec * LMultiplier[ftDynamic].ForceJacobian;
```
bcoconni, line 774: This code would be easier to read using references as you did in `ConfigureWheelSpinRows`:
```c++
    LagrangeMultiplier& lm_dynamic = LMultiplier[ftDynamic];

    lm_dynamic.ForceJacobian = mT * velocityDirection;
    lm_dynamic.Max = 0.;
    lm_dynamic.Min = -fabs(staticFFactor * dynamicFCoeff * vFn(eZ));
    lm_dynamic.MomentJacobian = vWhlContactVec * lm_dynamic.ForceJacobian;
```

src/models/FGLGear.cpp
```diff
@@ -739,6 +809,9 @@ void FGLGear::ComputeJacobian(const FGColumnVector3& vWhlContactVec)
     LMultiplier[ftRoll].Min = -LMultiplier[ftRoll].Max;
     LMultiplier[ftSide].Min = -LMultiplier[ftSide].Max;
```
bcoconni, line 810: For the code above use the references `lm_roll` and `lm_side`.

bcoconni (changes requested):
Thank you for taking into account my previous comments @Felipegalind0.

Some more change requests but the goal now is to move the wheel spin acceleration to `FGLGear`. There is no reason why the wheel acceleration should be computed in `FGAccelerations::CalculateFrictionForces`. Instead the computation of the wheel spin acceleration should be owned by `FGLGear` which has all the data needed to update the wheel spin rate (since it is actually the goal of computing the acceleration). This modification should simplify the code a bit and mimics the `FGPropeller` class which owns the acceleration computation of the propellers. 

I have also made some suggestions to hopefully improve the code legibility, mainly by extending the usage of references to Lagrange multiplier entries such as `LagrangeMultiplier& lm_roll = LMultiplier[ftRoll]` just as you do in `ConfigureWheelSpinRows`.

PR #1504, opened 2026-09-12
Fix Emscripten portability

Felipegalind0:
## Summary
- include the POSIX socket declarations required when compiling `FGfdmSocket` with Emscripten
- use Emscripten's POSIX `strerror_r` ABI even when `_GNU_SOURCE` is defined

## Testing
- configured and built the `jsbsim_wasm` target with Emscripten 6.0.9 against this branch to completion

This ports the socket and `strerror_r` fixes from the compatibility patch in my downstream `jsbsim-wasm` fork. It does not change terminal-color behavior or add a WASM build target to this repository.

### DIFF
--- a/src/input_output/FGfdmSocket.cpp
+++ b/src/input_output/FGfdmSocket.cpp
@@ -40,10 +40,13 @@ INCLUDES
 
 #if defined(_MSC_VER) || defined(__MINGW32__)
 #include <ws2tcpip.h>
-#elif defined(__OpenBSD__)
+#elif defined(__OpenBSD__) || defined(__EMSCRIPTEN__)
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <netinet/in.h>
 #include <netdb.h>
+#include <sys/select.h>
+#include <sys/time.h>
 #include <unistd.h>
 #include <fcntl.h>
 #else
--- a/src/simgear/misc/strutils.cxx
+++ b/src/simgear/misc/strutils.cxx
@@ -685,9 +685,9 @@ std::string error_string(int errnum)
   errno_t retcode;
   // Always makes the string in 'buf' null-terminated
   retcode = strerror_s(buf, sizeof(buf), errnum);
-#elif defined(_GNU_SOURCE)
+#elif defined(_GNU_SOURCE) && !defined(__EMSCRIPTEN__)
   return std::string(strerror_r(errnum, buf, sizeof(buf)));
-#elif (_POSIX_C_SOURCE >= 200112L) || defined(SG_MAC) || defined(__FreeBSD__)
+#elif (_POSIX_C_SOURCE >= 200112L) || defined(SG_MAC) || defined(__FreeBSD__) || defined(__EMSCRIPTEN__)
   int retcode;
   // POSIX.1-2001 and POSIX.1-2008
   retcode = strerror_r(errnum, buf, sizeof(buf));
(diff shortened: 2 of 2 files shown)


bcoconni:
Thanks for the PR @Felipegalind0.

> * omit the obsolete terminal-color portion of the downstream patch because current `master` has already removed those statics

Could you clarify what you mean by that? I could not find any reference to terminal colors in the diff.

seanmcleod70:
@Felipegalind0, also for some context. Is this PR in preparation for submitting a PR to include a wasm/Emscripten build option in the JSBSim repo? Or is it just to make your Emscripten builds more convenient.

Part of the reason I ask is that I pinged @0x62 about his https://github.com/0x62/jsbsim-wasm repo a fair while ago in terms of potentially integrating it into the JSBSim repo. I see you have a fork of his repo.

Felipegalind0:
@bcoconni You're right: there are no terminal-color changes in this diff. My description mixed background about a downstream patch with the changes actually included here. Sorry for the confusion.

The [older compatibility patch in my SDK fork](https://github.com/Felipegalind0/jsbsim-wasm/blob/27b353e679e3948a9daa78edac4acd917f3bb862/patches/jsbsim-emscripten-compat.patch#L9-L29) made the `FGJSBBase` ANSI escape-code arrays empty when compiling with Emscripten. Upstream [0b688c8 / #1487](https://github.com/JSBSim-Team/jsbsim/commit/0b688c801c52d800f75d5c41e5434ce5d7618e88) moved those codes into `FGLogConsole`, so that old hunk no longer applies as written.

Calling the terminal-color portion "obsolete" was imprecise: the old implementation disappeared, but terminal coloring did not. Preserving the downstream plain-text default would require adapting it to the new logger; that change is not part of this PR.

I've corrected the description. This PR contains only the POSIX socket-header and `strerror_r` ABI fixes.

Felipegalind0:
@seanmcleod70 Yes, I'd like to work towards WASM support in JSBSim. This PR handles two fixes our Emscripten build needs.

My plan is to follow the Python package's setup: keep the bindings and JavaScript/TypeScript package here, build them from the same C++ source, and publish a package people can install. I'd build on 0x62's work and keep his attribution and license notices.

I'm going to bring the SDK into my JSBSim fork, preserve the fixes we've already made, and have it build against the engine in the same checkout. Then I'll check the native and WASM builds, package it, and switch my [simulator](https://github.com/0SFS/0SFS.github.io) over to that package. Once it's working there and the tests pass, I'll follow up here with the results and a plan for upstreaming it in smaller PRs.

Felipegalind0:
I have this working in my fork now: [engine and WASM SDK together](https://github.com/Felipegalind0/jsbsim/tree/f7a80a6f064d22efe0591c20c8637b9c38179ba9/wasm).

The SDK lives under `wasm/` and builds against the engine in the same checkout. My simulator now uses a package built from that revision, and I've moved the old SDK checkout into a backup. The import preserves 0x62's history and notices, along with our existing fixes.

The [new CI workflow](https://github.com/Felipegalind0/jsbsim/actions/runs/34791576179) passes the native build with WASM disabled, the WASM build, SDK tests and packaging. Locally, the browser checks and matched native/WASM scenarios also pass.

I'll use this setup to work through the existing PR feedback and prepare the WASM integration as separate changes for upstream review.

bcoconni (approved):
> Preserving the downstream plain-text default would require adapting it to the new logger; that change is not part of this PR.

OK assuming this topic will be addressed in another PR, the current PR looks good to me.

bcoconni:
> I'll use this setup to work through the existing PR feedback and prepare the WASM integration as separate changes for upstream review.

For the record, the discussion about this topic is held in #984 and supported by the PR #1507.

0x62:
Looks great, and thanks @Felipegalind0 for pushing this over the line to upstream

PR #1505, opened 2026-09-14
Keep turbine spool state consistent during zero-time trim

Felipegalind0:
Calling `RunIC()` at a new turbine throttle setting updates steady thrust but leaves the exposed N1/N2 values at the previous operating point. The next time step then repairs the stale spool state. `FGTurbine::Trim()` calculates a local N2 that shadows the member and never updates N1.

This updates both spool members to the same dry-throttle operating point used by trim. The existing thrust calculation and time advancement are unchanged.

The regression uses the repository's F-16 fixture, checks idle/intermediate/full-power/idle transitions without advancing time, and checks continuity on the next integrated idle sample. It fails against current master and passes with this change. All 11 focused turbine, trim, initialization, clock and model-loading CTest targets passed on this branch.

Tested against master `14c19022943f5850daf2c6b90554050b3139b853`. This is independent of the WASM integration and requires no aircraft-specific changes.


seanmcleod70:
Hmm, what was a bit confusing when I took an initial look is why `FGTurbine` was starting off in a trim phase given that I was using the 737 cruise script, and there is no trim command in `737\cruise_init.xml`

https://github.com/JSBSim-Team/jsbsim/blob/master/aircraft/737/cruise_init.xml

Setting a breakpoint, I could see that `FGIC->trim_requested == 6 (TrimMode::tNone)` but `FGTurbine::Trim()` was being called as a result of `FGFDMExec::RunIC()`.

Turns out that in `FGTurbine::Calculate()` there is code to set the `phase = tpTrim` when `in.TotalDeltaT == 0`, i.e. for the first simulation frame, irrespective of/independently of a trim command via `FGInitialCondition`.

https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGTurbine.cpp#L151

Felipegalind0:
Thanks for tracing that @seanmcleod70

Yes, the turbine’s internal tpTrim phase is independent of the aircraft trim request: RunIC() evaluates it with zero elapsed time, even with tNone. That existing path is where this PR updates the exposed N1/N2 values.
Separately, our testing found that those new assignments also affect shut-off engines. They need a running-engine guard and regression coverage.

seanmcleod70:
> when in.TotalDeltaT == 0, i.e. for the first simulation frame...

Hmm, I blame it on being late at night and having a quick look and being influenced by the `zero-time` in the subject line for equating `in.TotalDeltaT == 0` to the very first simulation frame.

And missing the `Delta` in it's name.

 https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/FGFDMExec.cpp#L508

`Propulsion->GetRate()` always returns 1, so `in.TotlaDeltaT == 0` whenever `FGFDMExec::SuspendIntegration()` is called.

When trim is invoked `SuspendIntegration()` is called, but it's not the only time it's called, so although the turbine code uses it to set it's own phase to `tpTrim` there are other cases where this happens.

https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGTurbine.cpp#L151

https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/models/propulsion/FGTurbine.cpp#L609-L611

https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/FGFDMExec.cpp#L636-L638

bcoconni:
https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/FGFDMExec.cpp#L636-L638
Hmm, that means that turbines are trimmed each time `RunIC()` is called then. Not sure that it was the intended purpose? Shouldn't it rather be:
```diff
diff --git a/src/models/propulsion/FGTurbine.cpp b/src/models/propulsion/FGTurbine.cpp
index 685892579..987730e92 100644
--- a/src/models/propulsion/FGTurbine.cpp
+++ b/src/models/propulsion/FGTurbine.cpp
@@ -148,7 +148,7 @@ void FGTurbine::Calculate(void)
   }
 
   if (Cutoff && (phase != tpSpinUp)) phase = tpOff;
-  if (in.TotalDeltaT == 0) phase = tpTrim;
+  if (FDMExec->GetTrimStatus()) phase = tpTrim;
```

seanmcleod70:
No, I think the term `trim` is being overused here. What is effectively happening in `FGTurbine::Trim()` is that no lag filters are being applied which means that the thrust calculated is the steady-state thrust value for the particular throttle position. Which is what the actual trim routine requires, but there are other cases where this is also required.

For example if you want some Initial Condition with the thrust set at a steady state value. Now you can't actually specify the throttle position as an IC, but you can set the throttle position, e.g. to 1 and then call `RunIC()` which would then start things off with full thrust for the given conditions (altitude, airspeed etc.).

So no seeking/spool up/lag when calculating thrust based on throttle position.

https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/models/propulsion/FGTurbine.cpp#L350-L357

Versus seek/spool up/lag in `FGTurbine::Run()`.

https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/models/propulsion/FGTurbine.cpp#L215-L220

In the IC case there is no actual trim being performed, except if a trim command is one of the IC elements.

It did initially confuse me when I saw `FGTurbine::Trim()` being called even though I knew there was no trim operation being undertaken.

So renaming it to something like `CalculateThrustWithNoLag()` (not suggesting this is the ideal name) which is then called when a trim operation is being performed, plus in other cases like from `RunIC()` may prevent confusion for someone looking at the code, stepping through a debugger etc.

seanmcleod70:
As I mentioned in the very similar issue - https://github.com/JSBSim-Team/jsbsim/pull/1508 to this one.

> @Felipegalind0 coincidentally I had just started looking at the turbine trim code last night while looking into https://github.com/JSBSim-Team/jsbsim/discussions/1501, and in particular was looking at the difference between the `run()` method and the `trim()` method to see what calculations weren't performed in the trim case.

So in general the `trim()` routine calculates steady state thrust without lag but also omits the calculation of certain states, e.g. fuel flow, EGT, oil temp, oil pressure, N1 etc. Some of these states are updated during the next simulation frame after the integration suspension has been lifted, e.g. N1 and oil temp, but some like fuel flow will take some time to reach steady state via the `Seek()` method.

This PR and #1508 are designed to keep a subset of the states, i.e. N1 and fuel flow in sync, but don't address the other states.

Overall it seems a bit messy in terms of which states are in complete sync, which get into sync 1 simulation frame later and which take a couple of frames to get into sync based on their acceleration/deceleration values passed to `Seek()`.

So I'm wondering if the better overall solution is to not have a separate `trim()` method, but rather to simply handle the case of suspended integration within `run()`?

Either with conditionals around each call to `Seek()`, or modifying `Seek()` to return the target value if `in.TotalDeltaT == 0`?

```diff
double FGTurbine::Seek(double *var, double target, double accel, double decel) {
+ if (in.TotalDeltaT == 0)
+   return target;
  double v = *var;
  if (v > target) {
    v -= in.TotalDeltaT * decel;
    if (v < target) v = target;
  } else if (v < target) {
    v += in.TotalDeltaT * accel;
    if (v > target) v = target;
  }
  return v;
}
```

Issue #1248, opened 2025-03-23
Playing well with REPLs


bcoconni:
JSBSim has traditionally been employed within monolithic applications (such as FlightGear), where any JSBSim error generally resulted in the termination of the entire program. This assumption has simplified error handling within JSBSim itself: errors were considered unrecoverable since the calling application was expected to terminate when they occurred.

These days are gone now that we are providing a Python module and a Matlab S-function. This has opened up new possibilities for using JSBSim in interactive [Read-Eval-Print-Loop](https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop) (REPL) environments. This shift necessitates a reassessment and upgrade of our error handling and object state management strategies.

Specifically, this issue aims to discuss the following key requirements for robust JSBSim usage within REPLs (including, but not limited to, Python and Matlab):
1. **REPL Stability:** Under no circumstances should JSBSim cause the REPL to crash or terminate unexpectedly. Sudden REPL terminations are extremely frustrating for users, as it can lead to the loss of unsaved work and create a perception of poor software quality.
2. **Robust Object State Management:** Users working in a REPL environment often engage in exploratory, trial-and-error workflows. JSBSim objects may therefore exist in various intermediate or inconsistent states due to user errors. JSBSim should handle these states gracefully, allowing users to recover and continue their work without requiring, for example, the complete destruction and recreation of their instance of `FGFDMExec`. This implies the need for mechanisms to:
    * Detect and report errors clearly and informatively.
    * Provide methods for users to query the state of an object?
    * Offer options to reset or correct invalid object configurations, such as the ability to call `FGFDMExec::Load` again after a previous failure.
    * Prevent operations that would lead to crashes or undefined behavior when an object is in an invalid state. Typically prevent executing `FGFDMExec::Run` if `FGFDMExec::Load` has failed.

To address these requirements, the following work areas should be addressed:
1. **Memory Management:** Implement more rigorous memory management practices to prevent leaks, especially when the execution is leaving the *happy path* (i.e. the ideal execution flow of a program, where all conditions are met, and no errors occur).
    * For example, pull requests #1230 and #1245 addressed memory leaks that occurred during error conditions, which was previously not a concern since the operating system would automatically reclaim all reserved memory upon program termination.
    * The straightforward way to address this requirement is to replace all the calls to `new` and `delete` by smart pointers (i.e. `unique_ptr` and `shared_ptr`).
[...]

PR #1506, opened 2026-09-14
Unbind model properties before replacing their owners

Felipegalind0:
Loading another aircraft into an existing `FGFDMExec` can leave property nodes tied to objects that have been destroyed. Untying those nodes later reads through the stale pointers. A failed XML load also leaves partially populated models available to the next load attempt.

This unties properties while their owners are still alive, allocates fresh models after every prior load attempt, and restores the executive's own property bindings. It also clears the old property catalog before rebuilding it.

The regression covers repeated successful loads, missing aircraft, partial XML failure, recovery and destruction. It exercises the executive and model properties after replacement and checks for duplicate catalog entries.

On master `14c19022943f5850daf2c6b90554050b3139b853`, the new regression crashes with a segmentation fault. AddressSanitizer identifies a heap-use-after-free in `SGPropertyNode::untie`. With this change, all six reload cases and all 11 focused loading, atmosphere, clock, initialization and trim CTest targets pass. Six separate native AddressSanitizer sequences also pass without memory errors.

This is a native lifetime fix. It is also needed for repeated model loading in the planned WASM package, but does not depend on that package.


src/FGFDMExec.cpp
```diff
@@ -395,8 +403,12 @@ void FGFDMExec::InitializeModels(void)
...
 {
-
+  // Untying reads the last value from each bound object. Do this while all
+  // models and initial conditions are still alive, before replacing them.
+  Unbind();
   Models.clear();
+  PropertyCatalog.clear();
+  modelLoadAttempted = false;
   modelLoaded = false;
   return modelLoaded;
```
bcoconni, line 413: This is now the only location where the member `modelLoaded` is used and `FGFDMExec::Deallocate()` actually always returns `false` so you can safely remove the member `FGFDMExec::modelLoaded`.

src/FGFDMExec.cpp
```diff
@@ -395,8 +403,12 @@ void FGFDMExec::InitializeModels(void)
 
 bool FGFDMExec::DeAllocate(void)
 {
-
+  // Untying reads the last value from each bound object. Do this while all
+  // models and initial conditions are still alive, before replacing them.
+  Unbind();
   Models.clear();
+  PropertyCatalog.clear();
```
bcoconni, line 410: Actually most `FGFDMExec` members should be reset to their default values since they may have been modified by the previous calls to `LoadModel()` or `LoadScript()`:
https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/FGFDMExec.h#L636-L645
https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/FGFDMExec.h#L678-L684
https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/FGFDMExec.h#L689-L691
https://github.com/JSBSim-Team/jsbsim/blob/a6f86ae93256eb00c2c990b711faa5bbb72ecb20/src/FGFDMExec.h#L694-L701
```suggestion
  PropertyCatalog.clear();
  ChildFDMList.clear();
  TemplateFunctions.clear();
  Frame = 0;
  Terminate = false;
  dT = 1.0/120.0;
  sim_time = 0.0;
  holding = false;
  IncrementThenHolding = false;
  TimeStepsUntilHold = -1;
  trim_status = false;
  trim_completed = 0;
  IC.reset();
  Script.reset();
  Trim.reset();
  HoldDown = false;
  SRand(0);
  if (!IsChild) *FDMctr = 1;
```

src/FGFDMExec.h
```diff
@@ -644,6 +644,7 @@ class JSBSIM_API FGFDMExec : public FGJSBBase
   bool IncrementThenHolding;
   int TimeStepsUntilHold;
   bool Constructing;
+  bool modelLoadAttempted;
   bool modelLoaded;
```
bcoconni, line 648: This member is no longer needed.
```suggestion
```

tests/TestModelReload.py
```diff
@@ -0,0 +1,91 @@
...
+import xml.etree.ElementTree as et
+
+from JSBSim_utils import JSBSimTestCase, RunTest
+
+
+class TestModelReload(JSBSimTestCase):
+    def check_live_bindings(self, fdm):
+        # Exercise setters and getters owned by both the executive and newly
+        # allocated models; comparing only load return codes missed stale ties.
+        fdm.set_dt(0.01)
```
bcoconni, line 14: Your test does not check that `fdm["simulation/dt"]` should have been reset to 1/120 before calling `fdm.set_dt()`.

tests/TestModelReload.py
```diff
@@ -0,0 +1,91 @@
...
+        self.assertEqual(fdm["simulation/dt"], 0.01)
+        fdm["simulation/randomseed"] = 123
```
bcoconni, line 16: Ditto for `fdm["simulation/randomseed"]` which should have been reset to 0 before calling `fdm["simulation/randomseed"] = 123`.

tests/TestModelReload.py
```diff
@@ -0,0 +1,91 @@
...
+        self.assertEqual(fdm["simulation/randomseed"], 123)
+        fdm["simulation/reset"] = 2
+        self.assertEqual(fdm.get_sim_time(), 0.0)
+        fdm["ic/h-sl-ft"] = 3000
+        fdm["ic/vc-kts"] = 90
+        self.assertTrue(fdm.run_ic())
+        self.assertAlmostEqual(fdm["position/h-sl-ft"], 3000, places=6)
+        self.assertAlmostEqual(fdm["velocities/vc-kts"], 90, places=6)
+        fdm["simulation/pause"] = 1
```
bcoconni, line 25: Same as above.
Etc.

bcoconni (changes requested):
Thanks for your contribution. This PR is a step forward with regard to the goals of #1248 (playing well with REPLs).

However, in its current version, your PR does not restore the state of the `FGFDMExec` instance to its default state. For example, if a model changes the time step to something else than 120Hz and then you try to load a model that is using the default time step, then the time step will still have the value set by the first loaded model.

And your test case should check that default values are restored.

PR #1508, opened 2026-09-14
Establish steady turbine fuel flow during zero-time trim

Felipegalind0:
`FGTurbine::Trim()` computes steady thrust and spool speeds without advancing time, but never assigns `FuelFlow_pph`. Every zero-time evaluation therefore reports the fuel flow of whatever operating point preceded it, including the evaluation `propulsion/set-running` performs through `InitRunning()`. A caller that trims several throttle settings in turn reads each earlier setting's value.

On the repository's F-16 fixture the trimmed value is wrong in both directions:

| Throttle | Reported by trim | Steady value | Ratio |
| --- | --- | --- | --- |
| 0.35 | 1933.75 gph | 1064.16 gph | 1.82 |
| 1.00 | 1064.16 gph | 8548.54 gph | 0.12 |

This assigns the same steady products `Run()` seeks: the pre-bleed dry thrust times the corrected TSFC, floored at the idle flow, and the augmented thrust times `ATSFC` in whichever augmentation branch applies. `N2norm` is updated before TSFC is evaluated so that a TSFC parameter sees the operating point being trimmed rather than the previous one. Thrust calculation and time advancement are unchanged.

`TestTurbineTrimFuelFlow` covers three properties: a trimmed value cannot be sought away by the next integrated frame, repeated trims are independent of the order they are requested in, and the reported TSFC follows its own operating point. All three methods fail before this change. Thirteen focused turbine, turboprop, trim, hold-down, initialization, clock, model-loading and ground-reaction CTest targets pass on this branch.

Trim still leaves EGT, oil pressure, nozzle position and EPR at their previous values. That is out of scope here.

Depends on #1505, which makes the spool members authoritative during trim. Without it `Trim()` keeps a local `N2norm` that shadows the member, and updating the member for TSFC would leave the state only partly refreshed. This branch contains that prerequisite commit followed by this change. It is independent of the WASM integration and requires no aircraft-specific changes.


seanmcleod70:
@Felipegalind0 coincidentally I had just started looking at the turbine trim code last night while looking into https://github.com/JSBSim-Team/jsbsim/discussions/1501, and in particular was looking at the difference between the `run()` method and the `trim()` method to see what calculations weren't performed in the trim case.

In terms of updating the fuel flow during the trim solution iterations my first thought was whether this was going to result in a change in mass of the aircraft during the trim solution iterations. But double-checking the code it looks like it won't.

https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/FGPropulsion.cpp#L167-L170

In terms of updating the fuel flow instead of doing it on each iteration of the trim solution we could potentially do it within this section of `FGTurbine::Calculate()`?

https://github.com/JSBSim-Team/jsbsim/blob/14c19022943f5850daf2c6b90554050b3139b853/src/models/propulsion/FGTurbine.cpp#L124-L134

PR #1511, opened 2026-09-16
Publish the fuel tank temperature as a property

Felipegalind0:
`FGTank` computes a fuel temperature for every tank that declares an initial `<temperature>`, but `FGTank::bind` ties none of the accessors that expose it. `GetTemperature_degC()`, `GetTemperature()` and `SetTemperature()` all exist and are all unreachable from the property tree. A model author can set an initial fuel temperature from XML and can then never read it back, write it, log it or drive a system from it. Any model with a fuel temperature limit, a fuel temperature gauge, a fuel icing or anti-ice system or a cold-soak warning needs that value and cannot get it.

This ties `propulsion/tank[i]/temperature-degC` read/write, and does so only for the tanks whose thermal model is active. A tank that declares no `<temperature>` holds the `-9999.0` "no temperature set" flag rather than a temperature, so an unconditional tie would publish that flag under a name that promises a temperature, and every consumer that does not know the sentinel would log or plot it as one. Tying only what a tank actually computes follows what `FGTurbine::bindmodel` already does with `n1`, `n2` and `seized`. The change is strictly additive: a tank without an initial temperature keeps exactly the properties it has today.

Celsius is the unit the model works in, since `FGTank::Calculate` runs its heat balance against the total air temperature in Celsius. `GetTemperature()` would need a sentinel guard of its own before it could be published, because it converts unconditionally and would report -17966.2 °F for an inactive tank. A `temperature-degF` property remains compatible to add later if it is wanted; it is not in this change.

A note on that `-9999.0`, since this change is partly about not letting it escape. It is pre-existing and lives only in `FGTank`, where it is compared in three places, all guarded. What this change does is keep it from reaching anyone: with a conditional tie there is no property to read, which is the property tree's own way of saying "no value" and is stronger than publishing any in-band number, `-9999` or NaN alike. The same commit corrects `FGTank::Debug(0)`, which was the only place the flag was user-visible, on both counts: the constructor converts `Temperature` to Celsius before `Debug(0)` runs, so the startup line reported a Celsius number labelled Fahrenheit, and for a tank with no initial temperature it printed `-9999` as though it were one. It now names the unit and says when no temperature is modeled. After this the sentinel is not visible anywhere outside `FGTank` itself. Replacing it with `std::optional<double>` or an explicit flag would be better still, but that changes the public contract of three `FGTank` accessors, so I have left it out of this change rather than decide it here.

One limitation worth stating: absence protects the read path, not the write path. Writing `propulsion/tank[i]/temperature-degC` for a tank that models no temperature creates an untied node and is silently ignored, as it would be for any property name that does not exist. That is property tree behaviour rather than something this change introduces.

`TestFuelTankTemperature` uses the stock `c172x`, whose two fuel tanks both lack a `<temperature>`, so one tank can be given an initial temperature and the other left without. It checks that the property appears for the first tank and not the second and that the catalog holds exactly one entry; that the property is writable and that the written value is the one the tank then integrates; that the temperature falls towards the total air temperature at every step without passing it, with the endpoint bracketed against the closed-form relaxation `TAT + (T0 - TAT)*exp(-t/tau)`; and that the startup line names the unit it prints and never echoes the flag.

On master `29d2d6b8031569655560e260850c42314ab83ef0` the new regression fails four ways: the property does not exist, a write lands on an untied node the tank never reads so the temperature never moves, the startup line reads `Initial temperature: 15 Fahrenheit`, and the tank without a temperature reports `Initial temperature: -9999 Celsius`. With this change all four cases pass, as does the full 79-target CTest suite, built Release with Apple clang 21 and the Python module enabled. Concorde is the only stock aircraft that really declares tank temperatures: 17 tanks, 4 of them with a `<temperature>`, and it publishes exactly 4 of these properties.

This is an independent change. It does not depend on any of my other open PRs.


seanmcleod70:
Looks good to me.

tests/TestFuelTankTemperature.py
```diff
@@ -0,0 +1,218 @@
...
+        # the written value rather than from the configured one.
+        temperature0 = fdm['propulsion/tat-c'] + 90.0
+        fdm[TANK0_TEMPERATURE] = temperature0
+        self.assertAlmostEqual(fdm[TANK0_TEMPERATURE], temperature0,
+                               msg="The temperature property is not writable")
+
+        t0 = fdm.get_sim_time()
+        temperature = temperature0
+        tat_min = tat_max = fdm['propulsion/tat-c']
+        contents_min = contents_max = fdm['propulsion/tank[0]/contents-lbs']
```
bcoconni: Not sure why you are tracking the min/max of the ambient temperature and of the tank contents ? Both will have fixed values during the test, won't they?

tests/TestFuelTankTemperature.py
```diff
@@ -0,0 +1,218 @@
...
+
+        fdm = self.create_fdm()
+        fdm.set_aircraft_path('aircraft')
+        fdm.load_script(self.script_path)
+        fdm.run_ic()
+
+        # The first tank models a temperature, so it publishes one.
+        self.assertAlmostEqual(fdm[TANK0_TEMPERATURE],
+                               INITIAL_TEMPERATURE_DEGC,
+                               msg="The tank temperature (%f degC) should be the configured %f degF i.e. %f degC" % (fdm[TANK0_TEMPERATURE], INITIAL_TEMPERATURE_DEGF, INITIAL_TEMPERATURE_DEGC))
```
bcoconni, line 92: Use f-strings everywhere.

Historically, JSBSim have used c-format strings (i.e the `%` operator) because we have used Python since before the introduction of f-strings in Python 3.6. However, new Python code should use f-strings.

tests/TestFuelTankTemperature.py
```diff
@@ -0,0 +1,218 @@
...
+        aircraft_name, capacity = self.PrepareAircraft(INITIAL_TEMPERATURE_DEGF)
+
+        fdm = self.create_fdm()
+        fdm.set_aircraft_path('aircraft')
+        fdm.load_model(aircraft_name)
+        fdm.load_ic('reset01', True)
+        # Hold the aircraft down and leave its engine stopped so that neither
+        # the air the tank exchanges heat with nor the mass of fuel it holds
+        # moves much while the fuel temperature evolves.
+        fdm['forces/hold-down'] = 1.0
```
bcoconni, line 131: This comment is misleading since the engine is running (per the initial conditions in `aircraft/c172x/reset01.xml`:
https://github.com/JSBSim-Team/jsbsim/blob/29d2d6b8031569655560e260850c42314ab83ef0/aircraft/c172x/reset01.xml#L10
However the tank contents diminishes during the simulation otherwise we would have `contents_min == contents_max` and `tat_min == tat_max` (since the velocity would not change) and the tests below about the value of `temperature` being between `coldest` and `warmest` would fail.

bcoconni (changes requested):
The test `test_temperature_relaxes_towards_the_air_temperature` does not do what the comments tell (see my comments on lines 128-130).

If the comment on lines 128-130 would be true, then `coldest` would be equal to `warmest` (because `contents_min` would be equal to `contents_max` and `tat_min` would be equal to `tat_max`). So the tests `assertGreaterEqual(temperature, coldest, ...)` and `assertLessEqual(temperature, warmest, ...)` would reduce to `assertEqual(temperature, coldest, ...)`. But that latter condition would mean that the Euler scheme in `FGTank::Calculate` would match the exact solution in $T_{\text{amb}}+(T_0-T_{\text{amb}})\exp(-\frac{t}{\tau})$ which is notoriously false.
https://github.com/JSBSim-Team/jsbsim/blob/29d2d6b8031569655560e260850c42314ab83ef0/src/models/propulsion/FGTank.cpp#L354-L357
So the test succeeds **because** the engine is running and consuming some fuel from the tanks unlike what the comment claims.

